PR #2, opened 2015-10-22
Introduce analyzer and mapping

safchain:
This patch introduces:

- mapping, first one is based on Openstack Neutron, retrieving TenantID
  and VNI.
- analayzer, which is basically forwarding flow form agent to storage
  using mapping to complete flow structure.
- config package, used currently in neutron mapping
- logging package, which is basically a helper

The neutron mapping is using a local in-memory cache that is updated in
an async way.

This patch fixes the concept of flow computing a unique flow id based on
headers.

### DIFF
--- a/agents/sflow.go
+++ b/agents/sflow.go
@@ -23,15 +23,14 @@
 package agents
 
 import (
-	"fmt"
 	"net"
 
 	"github.com/google/gopacket"
 	"github.com/google/gopacket/layers"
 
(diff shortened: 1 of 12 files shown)


nplanel:
LGFM ; Just the flow timestamp granuality need to be changed

safchain:
Agreed, I'll reduce the precision of the timestamp

PR #21, opened 2016-01-13
[SFlow] Implement a more discret SFlow parser

nplanel:
Aim to get more informations from SFlow metadata, by enhancing via SkydivePacket

SkydrivePacket (gopacket.Packet inheritance)
SFlowReader (gopacket.PacketSource inheritance):
  SFlowReader read packet from packetsource via a Skydive PacketChannel and inject it in related agents
  Packetsource inject Packets in gopacket parser, with collected/correlated metadate from SFlow library (github.com/Preetam/sflow)

### DIFF
--- a/flow/probes/sflow.go
+++ b/flow/probes/sflow.go
@@ -23,6 +23,7 @@
 package probes
 
 import (
+	"io"
 	"net"
 	"strconv"
 	"strings"
@@ -32,6 +33,8 @@ import (
 	"github.com/google/gopacket/layers"
 
 	"github.com/pmylund/go-cache"
+
+	"github.com/Preetam/sflow"
 
 	"github.com/redhat-cip/skydive/analyzer"
 	"github.com/redhat-cip/skydive/config"
@@ -110,8 +113,6 @@ func (probe *SFlowProbe) getProbePath(index uint32) *string {
 }
 
 func (probe *SFlowProbe) Start() error {
-	var buf [maxDgramSize]byte
-
 	addr := net.UDPAddr{
 		Port: probe.Port,
 		IP:   net.ParseIP(probe.Addr),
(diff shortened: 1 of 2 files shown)


flow/probes/sflow.go
```diff
@@ -190,3 +183,136 @@ func NewSFlowProbe(a string, p int, g *graph.Graph) (*SFlowProbe, error) {
...
+	data := raw.Header
+	r.PacketChannel <- SkydivePacket{
+		gopacket.NewPacket(data, r.LinkType(), gopacket.DecodeOptions{}),
+		ci,
+		raw,
+		sample}
+}
+
+/* PacketSource */
+func (r *SFlowReader) NextPacket() (SkydivePacket, error) {
```
safchain: It seems that this function is always returning nil as error so I think this should be removed

flow/probes/sflow.go
```diff
@@ -190,3 +183,136 @@ func NewSFlowProbe(a string, p int, g *graph.Graph) (*SFlowProbe, error) {
...
+	packet := <-r.PacketChannel
+	m := packet.Metadata()
+	m.CaptureInfo = packet.ci
+	m.Truncated = m.Truncated || packet.ci.CaptureLength < packet.ci.Length
+
+	return packet, nil
+}
+
+func (p *SFlowReader) packetsToChannel() {
+	defer close(p.PacketChannel)
```
safchain: Never called since the loop will never end

flow/probes/sflow.go
```diff
@@ -190,3 +183,136 @@ func NewSFlowProbe(a string, p int, g *graph.Graph) (*SFlowProbe, error) {
...
+	for {
+		packet, err := p.NextPacket()
+		if err == io.EOF {
+			return
+		} else if err == nil {
+			p.PacketChannel <- packet
```
safchain: Something I don't understand we are reading the PacketChannel thanks to the NextPacket function a then we re-inject to the same channel what we just read ? we will have a infinite loop, iterating over the same packet ?

flow/probes/sflow.go
```diff
@@ -190,3 +183,136 @@ func NewSFlowProbe(a string, p int, g *graph.Graph) (*SFlowProbe, error) {
...
+		}
+	}
+}
+
+func (p *SFlowReader) Packets() chan SkydivePacket {
+	if p.PacketChannel == nil {
+		p.PacketChannel = make(chan SkydivePacket, 1000)
+		go p.packetsToChannel()
```
safchain: I do not understand the purpose of this go-routine since we are already in a go-routine started by the Start function. If I understand correctly we have a publisher of packets which is the Parse function and a consumer function which is asyncStart ? so from my point of view having only one go-routine the "Start" one should be enough ?

flow/probes/sflow.go
```diff
@@ -190,3 +183,136 @@ func NewSFlowProbe(a string, p int, g *graph.Graph) (*SFlowProbe, error) {
...
+}
+
+func (r *SFlowReader) Parse(dgram *sflow.Datagram) {
+	for _, sample := range dgram.Samples {
+		switch sample.SampleType() {
+		case sflow.TypeCounterSample:
+			for _, record := range sample.GetRecords() {
+				switch record.(type) {
+				case sflow.HostDiskCounters:
+					rec := record.(sflow.HostDiskCounters)
```
safchain: Since this is useless I think you could remove this

flow/probes/sflow.go
```diff
@@ -190,3 +183,136 @@ func NewSFlowProbe(a string, p int, g *graph.Graph) (*SFlowProbe, error) {
...
+		case sflow.TypeFlowSample:
+			samp := sample.(*sflow.FlowSample)
+			for _, record := range sample.GetRecords() {
+				switch record.(type) {
+				case sflow.RawPacketFlow:
+					rec := record.(sflow.RawPacketFlow)
+					r.RawPacketCounter++
+					r.CreatePacket(samp, &rec)
+				case sflow.ExtendedSwitchFlow:
+					rec := record.(sflow.ExtendedSwitchFlow)
```
safchain: Same here, could be removed

flow/probes/sflow.go
```diff
@@ -190,3 +183,136 @@ func NewSFlowProbe(a string, p int, g *graph.Graph) (*SFlowProbe, error) {
 
 	return probe, nil
 }
+
+type SkydivePacket struct {
+	gopacket.Packet
+	ci     gopacket.CaptureInfo
+	raw    *sflow.RawPacketFlow
```
safchain: Why not just using the pointer to the sflow struct ?

PR #22, opened 2016-01-13
[storage] elasticsearch config

nplanel:
"elasticsearch" key under "storage" section describe the connection to elasticsearch server.

[storage]
elasticsearch = 127.0.0.1:9200

### DIFF
--- a/storage/elasticsearch/elasticsearch.go
+++ b/storage/elasticsearch/elasticsearch.go
@@ -25,8 +25,6 @@ package elasticseach
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
-	"os"
 	"strconv"
 
 	elastigo "github.com/mattbaird/elastigo/lib"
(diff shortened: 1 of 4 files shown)


storage/elasticsearch/elasticsearch.go
```diff
@@ -141,10 +144,17 @@ func (c *ElasticSearchStorage) initialize() error {
...
 
-func New(addr string, port int) (*ElasticSearchStorage, error) {
+func New() (*ElasticSearchStorage, error) {
 	c := elastigo.NewConn()
-	c.Domain = addr
-	c.Port = strconv.FormatInt(int64(port), 10)
+
+	elasticonfig := config.GetConfig().Section("storage").Key("elasticsearch").Strings(":")
+	if len(elasticonfig) != 2 {
+		fmt.Fprintf(os.Stderr, "Config file is misconfigured : check elasticsearch key format\n")
```
safchain: Since this is a library and not a binary I think you should return an error instead of exit, IHMO

nplanel: Fixed

safchain: Thanks

PR #28, opened 2016-01-26
Remove static listening addr/port and add config parameters

safchain:
This patch add an helper to the config package in order to parse
host/port parameter. It removes static listening parameters and use
config parameters instead. It removes also dead code in alerting.


nplanel:
LGFM

PR #31, opened 2016-01-29
Flow Statistics

nplanel:
FlowTable
   * Implemented on Agent and Analyzer
   * Conversation JSON Flows dump 
   * FlowTable conversation output /statics/conversation.html /data/conversation.json

PCAP probe agent
   * read a PCAP from skydive_agent command line argument (-trace)
   * or if set with a empty string (-trace "") read the filename from the config file ([agent].pcaptrace)
   * 5 PCAP packets are included in each fake (generated) SFlow packet


analyzer/client.go
```diff
@@ -61,6 +58,21 @@ func (c *Client) SendFlows(flows []*flow.Flow) {
 	}
 }
 
+func (c *Client) AsyncFlowsUpdate(ft *flow.FlowTable, every time.Duration) {
+	ticker := time.NewTicker(every)
+	defer func() {
```
safchain: nit: defer ticker.Stop()

analyzer/client.go
```diff
@@ -61,6 +58,21 @@ func (c *Client) SendFlows(flows []*flow.Flow) {
...
+		ticker.Stop()
+	}()
+	for {
+		select {
```
safchain: select not needed here

analyzer/server.go
```diff
@@ -128,6 +138,50 @@ func (s *Server) FlowSearch(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *Server) serveIndex(w http.ResponseWriter, r *http.Request) {
+	if r.URL.Path == "/data/conversation.json" {
```
safchain: I would prefer to have two handlers. One for the html page and one for the json

analyzer/server.go
```diff
@@ -179,14 +233,19 @@ func NewServer(port int, router *mux.Router) (*Server, error) {
...
 		Port:                port,
 		Router:              router,
 		TopoServer:          tserver,
 		GraphServer:         gserver,
 		FlowMappingPipeline: pipeline,
+		FlowTable:           flowtable,
 	}
+	server.RegisterStaticEndpoints()
 	server.RegisterRpcEndpoints()
+	go flowtable.AsyncExpire(server.flowExpire, 10*time.Minute)
```
safchain: I do not the purpose of having a callback function (flowExpire) which is only using the flowtable remove function. It seems to me this is an useless indirection ?

cmd/skydive_agent/skydive_agent.go
```diff
@@ -42,12 +42,14 @@ import (
...
 
 func usage() {
-	fmt.Printf("Usage: %s -conf <config.ini> [-h]\n", os.Args[0])
+	fmt.Printf("Usage: %s -conf <config.ini> -trace <trace.pcap> [-h]\n", os.Args[0])
 }
 
 func main() {
 	filename := flag.String("conf", "/etc/skydive/skydive.ini",
 		"Config file with all the skydive parameter.")
+	pcaptrace := flag.String("trace", "",
```
safchain: I think this functionality should be moved to another binary, like a tool to read a pcap file sending sflow packet to the agent. BTW I think this feature is really useful for functional testing for instance.

etc/skydive.ini.default
```diff
@@ -20,7 +20,7 @@ listen = 8082
 [agent]
 listen = 8081
 analyzers = 127.0.0.1:8082
-
+pcaptrace = "testing-trace.pcap"
```
safchain: should be removed

flow/flow.go
```diff
@@ -150,10 +138,21 @@ func FLowsFromSFlowSample(sample *layers.SFlowFlowSample, probePath *string) []*
...
 		}
 
-		flow := New(&record.Header, probePath)
+		packet := &record.Header
+		key := (FlowKey{}).fillFromGoPacket(packet)
+		flow, new := ft.GetFlow(key.String(), packet)
+		if new {
+			flow.ProbeGraphPath = *probePath
+
+			flow.DebugTimestamp = time.Now().Unix()
```
safchain: What is the purpose of these fields ? If this is only for debugging purpose this should be removed, if this is for something else maybe we could change the name.

flow/probes/pcap.go
```diff
@@ -0,0 +1,399 @@
...
+
+	return nil
+}
+
+func (probe *PcapProbe) flowExpire(f *flow.Flow) {
+	/* send a special event to the analyzer */
+}
+
+var nbpackets int = 0
+var nbsflowmsg int = 0
```
safchain: seems to be useless.

flow/probes/pcap.go
```diff
@@ -0,0 +1,399 @@
...
+		//				CounterSamples  []SFlowCounterSample
+	}
+	sflowLayer.FlowSamples = append(sflowLayer.FlowSamples, sf)
+	sflowLayer.SampleCount = uint32(len(sflowLayer.FlowSamples))
+
+	rawBytes := SFlowDatagramSerialize(sflowLayer, packets)
+	return rawBytes
+}
+
+func genEthIPUdp(payload []byte) []byte {
```
safchain: seems to be not used and not exported

flow/probes/pcap.go
```diff
@@ -0,0 +1,399 @@
...
+		ethernetLayer,
+		ipLayer,
+		udpLayer,
+		gopacket.Payload(payload),
+	)
+
+	return buffer.Bytes()
+}
+
+func writePcap(packet []byte) {
```
safchain: same here

flow/probes/pcap.go
```diff
@@ -0,0 +1,399 @@
...
+	for {
+		select {
+		case <-ticker.C:
+			logging.GetLogger().Debug("%d packets replayed | %s", nbpackets, probe.flowtable.String())
+		}
+	}
+}
+
+func (probe *PcapProbe) Start() error {
+	f, err := os.Open(probe.Filename)
```
safchain: This probe should live capture from an interface, maybe specified from a conf parameter for a first step. As specified before I would prefer to have a other tool in order to read pcap file a send sflow packet to the agent.

analyzer/client.go
nplanel: copy/paste behavior
I will remove them

analyzer/client.go
nplanel: agree

analyzer/server.go
nplanel: ok

flow/probes/sflow.go
```diff
@@ -126,6 +130,9 @@ func (probe *SFlowProbe) Start() error {
 	// start index/mac cache updater
 	go probe.cacheUpdater()
 
+	flowtable := flow.NewFlowTable()
+	go flowtable.AsyncExpire(probe.flowExpire, 5*time.Minute)
```
safchain: does it mean that the flows are not going to expire at the agent side ?

flow/table.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/redhat-cip/skydive/logging"
+)
+
+type FlowTable struct {
+	lock  sync.RWMutex
+	table map[string]*Flow
+}
+
+type FlowTableAsyncNotificaionUpdate interface {
```
safchain: typo: FlowTableAsyncNotificaionUpdate -> FlowTableAsyncNotificationUpdate

analyzer/server.go
nplanel: Good catch, the Remove should not be called on the callback, as AsyncExpire delete it just after calling the callback.

The callback is here to do something the flow just before it will be removed.
It's could be used to send it to the database for example.

flow/table.go
```diff
@@ -0,0 +1,157 @@
...
+		ticker.Stop()
+	}()
+	for {
+		select {
+		case now := <-ticker.C:
+			flowTableSzBefore := len(ft.table)
+			expire := now.Unix() - int64((every).Seconds())
+
+			ft.lock.Lock()
+			defer ft.lock.Unlock()
```
safchain: Dead lock here, defer will be applied when leaving the function, but since we are in a infinite loop...

cmd/skydive_agent/skydive_agent.go
nplanel: Agree
I will do a pcap2sflow-injector external tool
That make more sense and will test the whole chain

flow/probes/pcap.go
nplanel: I used it to dump to wireshark ... for debugging purpose ... yeah I will remove it :)

analyzer/server.go
```diff
@@ -47,9 +51,15 @@ type Server struct {
 	GraphServer         *graph.Server
 	FlowMappingPipeline *mappings.FlowMappingPipeline
 	Storage             storage.Storage
+	FlowTable           *flow.FlowTable
+}
+
+func (s *Server) flowExpire(f *flow.Flow) {
+	s.FlowTable.Remove(f)
```
safchain: I think there is an issue here, I'll try to explain what I understood :)

1. flowExpire is a callback that is called by AsyncExpire
2. flowExpire is calling Remove which is going to delete the flow
3. then AsyncExpire is going to delete the flow as well ?

flow/probes/sflow.go
nplanel: They will expire, even with an empty callback
See my comment (on the server side) above

statics/conversation.html
```diff
@@ -0,0 +1,202 @@
+
+<!DOCTYPE html>
+<html class="ocks-org do-not-copy">
+<meta charset="utf-8">
+<title>Les Misérables Co-occurrence</title>
```
safchain: Should be cleanup if we want to merge it, maybe adding it to the main UI

flow/table.go
nplanel: Good catch

flow/probes/sflow.go
safchain: Yeah, thx, I saw that after I commented :)

analyzer/server.go
nplanel: Yes this line should not be here, see my comment : https://github.com/redhat-cip/skydive/pull/31/files#r51262801

flow/table.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+type ExpireFunc func(f *Flow)
+
+func (ft *FlowTable) AsyncExpire(fn ExpireFunc, every time.Duration) {
+	ticker := time.NewTicker(every)
+	defer ticker.Stop()
+	for {
+		now := <-ticker.C
+		flowTableSzBefore := len(ft.table)
```
safchain: There is a race here, you should move up the lock. If you build your project with -race we will be notified of this

analyzer/client.go
```diff
@@ -61,6 +58,19 @@ func (c *Client) SendFlows(flows []*flow.Flow) {
 	}
 }
 
+func (c *Client) AsyncFlowsUpdate(ft *flow.FlowTable, every time.Duration) {
+	ticker := time.NewTicker(every)
+	defer ticker.Stop()
+	for {
+		select {
```
safchain: nit: useless select

cmd/pcap2sflow-replay/pcap2sflow-replay.go
```diff
@@ -0,0 +1,329 @@
+/*
```
safchain: great to have this new tool

flow/flow.go
```diff
@@ -150,10 +138,21 @@ func FLowsFromSFlowSample(sample *layers.SFlowFlowSample, probePath *string) []*
...
 			continue
 		}
 
-		flow := New(&record.Header, probePath)
+		packet := &record.Header
+		key := (FlowKey{}).fillFromGoPacket(packet)
+		flow, new := ft.GetFlow(key.String(), packet)
+		if new {
+			if probePath == nil {
+				flow.ProbeGraphPath = "probe-topology-path-not-found"
```
safchain: I suggest to let it empty IMHO

safchain:
LGTM, just a race condition remaining

storage/elasticsearch/elasticsearch.go
```diff
@@ -47,7 +47,7 @@ func (c *ElasticSearchStorage) StoreFlows(flows []*flow.Flow) error {
 			logging.GetLogger().Debug("Indexing: %s", string(j))
 		}
 
-		_, err = c.connection.Index("skydive", "flow", flow.GetUUID(), nil, *flow)
+		_, err = c.connection.Index("skydive", "flow", flow.UUID, nil, *flow)
```
lebauce: This fails with :
`Error while indexing: json: unsupported type: map[int32]*flow.FlowEndpointsStatistic`

safchain: Since the Flow schema changed, maybe we could add something to convert new flow format to the old one in order to keep the index working while thinking to another way of storing. So that we will not have feature regression.

safchain:
I got the following traceback while playing with the topology script

panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x8 pc=0x490b90]

goroutine 52 [running]:
github.com/redhat-cip/skydive/storage/elasticsearch.flow2OldFlow(0xc820017b60, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/safchain/mycode/gocode/src/github.com/redhat-cip/skydive/storage/elasticsearch/elasticsearch.go:82 +0x200
github.com/redhat-cip/skydive/storage/elasticsearch.(*ElasticSearchStorage).StoreFlows(0xc8200321a0, 0xc8212da618, 0x1, 0x1, 0x0, 0x0)
	/home/safchain/mycode/gocode/src/github.com/redhat-cip/skydive/storage/elasticsearch/elasticsearch.go:119 +0x24b
github.com/redhat-cip/skydive/analyzer.(*Server).AnalyzeFlows(0xc820619200, 0xc8212da618, 0x1, 0x1)
	/home/safchain/mycode/gocode/src/github.com/redhat-cip/skydive/analyzer/server.go:66 +0xe4
github.com/redhat-cip/skydive/analyzer.(*Server).handleUDPFlowPacket(0xc820619200, 0xc8212f4008)
	/home/safchain/mycode/gocode/src/github.com/redhat-cip/skydive/analyzer/server.go:87 +0x3dc
github.com/redhat-cip/skydive/analyzer.(*Server).ListenAndServe.func3(0xc821118160, 0xc820619200)
	/home/safchain/mycode/gocode/src/github.com/redhat-cip/skydive/analyzer/server.go:115 +0x194
created by github.com/redhat-cip/skydive/analyzer.(*Server).ListenAndServe
	/home/safchain/mycode/gocode/src/github.com/redhat-cip/skydive/analyzer/server.go:116 +0xc7

PR #35, opened 2016-02-02
Fix AddEdge in the memory backend, edge was added even with missing node

safchain:
This patch initiate UTs for the memory backend


nplanel:
LGTM

PR #36, opened 2016-02-02
Split out the agent main and the agent code aiming to instanciate agent

safchain:
Splitted out an agent can be instanciate and tested.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -32,7 +32,7 @@ import (
 
 var cfg *ini.File
 
-func InitConfig(filename string) error {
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -32,7 +32,11 @@ import (
 
 var cfg *ini.File
 
-func InitConfig(filename string) error {
+func InitConfig() {
```
nplanel: Why keeping an empty config ? ... who calling it ?

safchain: I Introduced it in order to be able to write functional tests which don't have config files

PR #37, opened 2016-02-03
Add config parameter in order to set the WebSocket pong timeout

safchain:
Default pong timeout is set to 5 seconds


nplanel:
LGTM

PR #40, opened 2016-02-03
Fix potential race condition as ListenAndServe can be exec in goroutine

safchain:


topology/graph/server.go
```diff
@@ -329,16 +329,11 @@ func (s *Server) serveMessages(w http.ResponseWriter, r *http.Request) {
 func (s *Server) ListenAndServe() {
 	s.Graph.AddEventListener(s)
```
nplanel: Why not registered the graph listener in NewServer as well ?

safchain: If I move this line in the New function, we will have events coming from the graph and then writing to the broadcast chan while the server will not be able (yet) to consume them. That's why I wanted to keep it close to the start of the messages consuming. Moreover there no race condition with this function which is already protected by a lock.

PR #42, opened 2016-02-05
Add functional tests for topology

safchain:
This set of commit add functional tests and fix coverage reports

The PR https://github.com/redhat-cip/skydive/pull/41 needs to be merged before this could pass travis

### DIFF
--- a/tests/topology_test.go
+++ b/tests/topology_test.go
@@ -1,0 +1,309 @@
+/*
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


tests/topology_test.go
```diff
@@ -0,0 +1,579 @@
...
+cleanup = 30
+
+[sflow]
+listen = 5000
+
+[ovs]
+ovsdb = 6400
+
+[logging]
+default=DEBUG
```
nplanel: we should not print debug message during test ..... please remove the logging section or set it to INFO or even better ERROR

tests/topology_test.go
```diff
@@ -0,0 +1,579 @@
...
+
+	setupCmds := []string{
+		"ovs-vsctl add-br br-test1",
+	}
+
+	tearDownCmds := []string{
+		"ovs-vsctl del-br br-test1",
+	}
+
+	testPassed := false
```
nplanel: why not propagate the t *testing.T in the callback ?
It will be helpful to spot the error directly at the right place

tests/topology_test.go
safchain: Yes I will it was to debug the CI issue due to the PR 41 which was not merged

tests/topology_test.go
safchain: Since these are functional tests we rely on external events like netlink, ovsdb callback and we don't know when and how the topology will be fully updated. The only thing we can say is that the topology is not the one expected after a delay. Here the maximum delay to have something corresponding to our expectation is set to 1m. So in the callback there is no real error but just incomplete informations.

PR #44, opened 2016-02-05
Flow tests

nplanel:
Introduce Tests for
 o flow/table_test.go
 o pcap2sflow-replay (functional testing)


flow/table_test.go
```diff
@@ -0,0 +1,300 @@
...
+	return ft
+}
+
+func TestNewFlowTable(t *testing.T) {
+	ft := NewFlowTable()
+	if ft == nil {
+		t.Error("new FlowTable return null")
+	}
+}
+func TestFlowTable_String(t *testing.T) {
```
safchain: Should be TestFlowTableString, please remove the "_"

flow/table_test.go
```diff
@@ -0,0 +1,300 @@
...
+	ft := NewFlowTable()
+	if "0 flows" != ft.String() {
+		t.Error("FlowTable too big")
+	}
+	ft = NewFlowTableSimple(t)
+	if "2 flows" != ft.String() {
+		t.Error("FlowTable too big")
+	}
+}
+func TestFlowTable_Update(t *testing.T) {
```
safchain: Same here, please remove "_"

flow/table_test.go
```diff
@@ -0,0 +1,300 @@
...
+	ft := NewFlowTableSimple(t)
+	/* simulate a collision */
+	f := &Flow{}
+	ft.table["789"] = f
+	ft.table["789"].UUID = "78910"
+	f = &Flow{}
+	f.UUID = "789"
+	ft.Update([]*Flow{f})
+	NewFlowTableComplex(t)
```
safchain: Should be removed no ? why a new complex table here

flow/table_test.go
```diff
@@ -0,0 +1,300 @@
...
+	ft = NewFlowTableSimple(t)
+	if "2 flows" != ft.String() {
+		t.Error("FlowTable too big")
+	}
+}
+func TestFlowTable_Update(t *testing.T) {
+	ft := NewFlowTableSimple(t)
+	/* simulate a collision */
+	f := &Flow{}
+	ft.table["789"] = f
```
safchain: Why not using ft.Update here ?

flow/table_test.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func isJSON(s string) bool {
+	var js map[string]interface{}
+	return json.Unmarshal([]byte(s), &js) == nil
+}
+
+func TestFlowTable_JSONFlowConversationEthernetPath(t *testing.T) {
+	ft := NewFlowTableComplex(t)
+	statStr := ft.JSONFlowConversationEthernetPath()
```
safchain: I would prefer to unmarshal the JSON so that the isJSON will be done first and then we can test the object received like this len(obj.nodes) == 0 ...
Because there is not guarantee that the order will be kept

flow/table_test.go
```diff
@@ -0,0 +1,300 @@
...
+	if statStr == `{"nodes":[],"links":[]}` {
+		t.Error("stat should not be empty")
+	}
+	if !isJSON(statStr) {
+		t.Error("stat should be JSON format")
+	}
+
+	decoded := new(interface{})
+	if err := json.Unmarshal([]byte(statStr), decoded); err != nil {
```
safchain: didn't see before, but you decode the object here, you did the job twice

flow/table_test.go
nplanel: I used a tool (go-tests) to generated the template
As String is the method for FlowTable, I think the _ make sense

flow/table_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	ft2 := ft.NewFlowTableFromFlows(flows)
+	if len(ft.table) != len(ft2.table) {
+		t.Error("NewFlowTable(copy) are not the same size")
+	}
+	flows = flows[:0]
+	for _, f := range ft.table {
+		flows = append(flows, f)
+	}
+	ft3 := ft.NewFlowTableFromFlows(flows)
```
safchain: Maybe I missed something, I don't understand the purpose of this test. It seems to me that your testing the same thing than L271 except that flow are not copied before ? so the NewFlowTableFromFlows that you are testing process always a pointer ?

flow/table_test.go
nplanel: NewFlowTableComplex is called to test it, as it's not the same generation of FlowTable
Basically Update have 2 tests (with Simple and Complex)
 Simple is use to do coverage on the collision case

flow/table_test.go
nplanel: all the trick is here, to simulate the collision and trig the checker (for coverage)

flow/table_test.go
nplanel: Yeah I did it in 2 steps, I can remove isJSON()

safchain:
CI tests are failing since 3 patches it seems

flow/table_test.go
safchain: K, indeed

flow/table_test.go
safchain: k, got it

flow/table_test.go
safchain: mmm, but you are not doing anything with this Complex table

nplanel: yeah but inside Compex it's been checked : checkFlowTable(t, ft)

flow/table_test.go
nplanel: The point here is to test if you pass a copy of flow or a list of referenced flow, they will be counted and inserted in the new flowtable the same way.
It's more likely to have strong test for futur code development

flow/table_test.go
safchain: Hehe, Maybe I'm not clear or didn't understand anything :)

TestFlowTable_Update let me think that this method is going to check the Update method, but:

- you don't check that the Update method worked, checking the status of the flow table after.
- NewFlowTableComplex(t) doesn't do anything with the local variable. If the goal is to call it in order to trigger a test, this function is already called by some other test that does thing with the FT returned.
- The Update method in the case of the complex call doesn't seem to be called.

nplanel:
@safchain : Travis seems still failed with "Connection timeout reached" on topology_test
I tested locally on master and the flow-tests branch, and have the same result every time.

Previous failed was due to a slow fork/exec on Travis platform, as he can't inject the packets within 3 secs .... I change it to 10 second, and now it's pass.

flow/table_test.go
nplanel: Thanks you catch the mistake : Simple do the Update test internally as well ... In my head I was assume the Complex did too.

nplanel:
@safchain : After a couplesss of test, the only way to fix it is to stop the global agent, as the instance is keep between tests (topology_test and agent_sflow_test)
So we will merge this PR after #45

PR #45, opened 2016-02-08
Make agent and analyzer handle stop gracefully

lebauce:


analyzer/server.go
```diff
@@ -108,18 +109,24 @@ func (s *Server) ListenAndServe() {
...
-		conn, err := net.ListenUDP("udp", addr)
+		s.Conn, err = net.ListenUDP("udp", addr)
 		if err != nil {
 			panic(err)
 		}
-		defer conn.Close()
+		defer s.Conn.Close()
 
-		s.handleUDPFlowPacket(conn)
+		s.handleUDPFlowPacket(s.Conn)
```
nplanel: I think the connection should not be passed as argument as you change it to a class member. So no argument must be applied.

cmd/skydive_agent/skydive_agent.go
```diff
@@ -49,5 +51,13 @@ func main() {
...
-	agent.NewAgent().Start()
+	agent := agent.NewAgent()
+	agent.Start()
+
+	ch := make(chan os.Signal)
+	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
+	<-ch
+
+	agent.Stop()
+	fmt.Println("Skydive Agent stopped")
```
nplanel: logging.GetLogger().Notice() will be more appropriated

cmd/skydive_analyzer/skydive_analyzer.go
```diff
@@ -67,5 +69,11 @@ func main() {
...
 
 	logging.GetLogger().Notice("Skydive Analyzer started !")
-	server.ListenAndServe()
+	go server.ListenAndServe()
+
+	ch := make(chan os.Signal)
+	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
+	<-ch
+
+	server.Stop()
```
nplanel: logging.GetLogger().Notice("Skydive Analyzer stopped")

topology/graph/client.go
```diff
@@ -178,6 +187,8 @@ func NewAsyncClient(addr string, port int, path string) *AsyncClient {
 		Port:      port,
 		Path:      path,
 		messages:  make(chan string, 500),
+		quit:      make(chan struct{}, 1),
```
nplanel: why not using bool ?   It's more readable like : c.quit <- true

topology/graph/server.go
```diff
@@ -68,6 +69,7 @@ type WSServer struct {
 	Alert      *Alert
 	clients    map[*WSClient]bool
 	broadcast  chan string
+	quit       chan struct{}
```
nplanel: same here

topology/probes/netlink.go
```diff
@@ -373,11 +373,14 @@ Loop:
 	for {
 		n, err := syscall.EpollWait(epfd, events[:], 1000)
 		if err != nil {
-			logging.GetLogger().Error("Failed to receive from netlink messages: %s", err.Error())
-			continue
+			errno, ok := err.(syscall.Errno)
+			if ok && errno != syscall.EINTR {
```
nplanel: you should call continue (recall syscall) in case of syscall.EINTR

topology/probes/netlink.go
```diff
@@ -373,11 +373,14 @@ Loop:
...
+				logging.GetLogger().Error("Failed to receive from netlink messages: %s", err.Error())
+				continue
+			}
 		}
 
-		if n == 0 {
+		if n <= 0 {
```
nplanel: n **>=** 0 will be correct

n==0 mean timeout
> 0 mean you received n descriptors events

topology/graph/client.go
lebauce: That's what I did at first, but then I thought it would be strange to exit if we do `c.quit <- false`. Maybe it was just too much thinking :-)
Fixed

topology/graph/server.go
lebauce: This one is not mine, I just made `quit` a class member.
Done

topology/probes/netlink.go
lebauce: The idea is to keep the original behavior, except that with a `syscall.EINTR` (Ctrl+C for instance) we don't restart the loop. In this case, `n == -1`, that why I changed `n == 0` to `n < 0`.

topology/probes/netlink.go
lebauce: See comment above

cmd/skydive_analyzer/skydive_analyzer.go
lebauce: Done

analyzer/server.go
lebauce: Indeed. Fixed

topology/probes/netlink.go
nplanel: you should not do like this, 
syscall.EINTR **!=** Ctrl+C (SIGINT)
A Syscall can return EINTR for many other reasons, like after attach gdb to the process and continue, basically it's a cancellation point.

BTW, the chan is used as a asynchrous flag and break Loop didn't break the for loop.
I will annotate the code we a solution

topology/probes/netlink.go
```diff
@@ -373,11 +373,14 @@ Loop:
 	for {
```
nplanel: The code should be :    (package "sync/atomic")

	for ;atomic.LoadInt32(&u.running) > 0 ; {
		n, err := syscall.EpollWait(epfd, events[:], 1000)
		if err != nil {
			errno, ok := err.(syscall.Errno)
			if ok && errno != syscall.EINTR {
				logging.GetLogger().Error("Failed to receive from netlink messages: %s", err.Error())
			}
			continue
		}
if n == 0 {
      continue
}
		msgs, err := s.Receive()

topology/graph/client.go
```diff
@@ -47,6 +47,8 @@ type AsyncClient struct {
 	Port      int
 	Path      string
 	messages  chan string
+	quit      chan bool
+	wg        *sync.WaitGroup
```
nplanel: Don't set a pointer here, use it directly as it not coming from outside the object
It will be auto instanciated with object

topology/graph/client.go
```diff
@@ -178,6 +187,8 @@ func NewAsyncClient(addr string, port int, path string) *AsyncClient {
 		Port:      port,
 		Path:      path,
 		messages:  make(chan string, 500),
+		quit:      make(chan bool, 1),
+		wg:        &sync.WaitGroup{},
```
nplanel: This line can be removed

topology/graph/server.go
```diff
@@ -41,11 +41,12 @@ const (
...
-	Alert    *Alert
-	Router   *mux.Router
-	wsServer *WSServer
-	Host     string
+	Graph     *Graph
+	Alert     *Alert
+	Router    *mux.Router
+	wsServer  *WSServer
+	Host      string
+	waitGroup *sync.WaitGroup
```
nplanel: be consistant in name : wg vs waitGroup
Same comment here declare it like : wg sync.WaitGroup

topology/graph/server.go
```diff
@@ -327,20 +331,30 @@ func (s *Server) serveMessages(w http.ResponseWriter, r *http.Request) {
...
+
 func NewServer(g *Graph, a *Alert, router *mux.Router, pongWait time.Duration) *Server {
 	s := &Server{
-		Graph:  g,
-		Alert:  a,
-		Router: router,
+		Graph:     g,
+		Alert:     a,
+		Router:    router,
+		waitGroup: &sync.WaitGroup{},
```
nplanel: same here

topology/server.go
```diff
@@ -44,6 +47,8 @@ type Server struct {
 	Router *mux.Router
 	Addr   string
 	Port   int
+	sl     *stoppableListener.StoppableListener
+	wg     *sync.WaitGroup
```
nplanel: same here

topology/server.go
```diff
@@ -119,6 +144,7 @@ func NewServer(g *graph.Graph, a string, p int, router *mux.Router) *Server {
 		Router: router,
 		Addr:   a,
 		Port:   p,
+		wg:     &sync.WaitGroup{},
```
nplanel: this line can be removed

topology/server.go
```diff
@@ -110,7 +115,27 @@ func (s *Server) RegisterRpcEndpoints() {
 }
 
 func (s *Server) ListenAndServe() {
-	http.ListenAndServe(s.Addr+":"+strconv.FormatInt(int64(s.Port), 10), s.Router)
+	defer s.wg.Done()
+	s.wg.Add(1)
+
+	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", s.Addr, s.Port))
+	if err != nil {
+		panic(err)
```
nplanel: don't use panic, use logging.GetLogger().Fatal() with a simple and different message

topology/probes/netlink.go
safchain: mmm, I'm ok to change the way to stop, but break Loop, break the  loop :)

nplanel: I tested it after, yeah the go syntax is a little bit wierd like :+1: 
Loop:
   for {
    if x {
        break Loop  // looks like a goto, but it just to nominate a loop in go in case of multiple nested loop.
    } 
}

nplanel:
LGTM

PR #46, opened 2016-02-08
Change log from Error to Warning when netlink doesn't find intf

safchain:
We can be notified that a interface is created while it has just be
removed, then it is impossible to find it.


lebauce:
LGTM

PR #48, opened 2016-02-08
Add Docker probe

lebauce:
This Docker probes leverages the Netns probe when a new container is created, so the topology shows the veth of the containers. It keeps a cache of the MAC addresses of all the containers to enhance the flow with the ID of the container. Note that this last part is disabled at the moment as the format of the flows has changed.


topology/probes/netns.go
```diff
@@ -123,7 +123,7 @@ func NewNetNsNetLinkTopoUpdater(g *graph.Graph, n *graph.Node) *NetNsNetLinkTopo
 	}
 }
 
-func (u *NetNSProbe) onNetNsCreated(path string) {
+func (u *NetNSProbe) Probe(path string) {
```
nplanel: why not use Register/Unregister words ?

topology/probes/netns.go
```diff
@@ -167,6 +167,14 @@ func (u *NetNSProbe) onNetNsDeleted(path string) {
 	delete(u.nsnlProbes, name)
 }
 
+func (u *NetNSProbe) onNetNsCreated(path string) {
```
nplanel: Why keeping the old names ?

topology/probes/netns.go
lebauce: Register/Unregister is indeed better. Done

topology/probes/netns.go
lebauce: Done

agent/agent.go
```diff
@@ -97,19 +97,27 @@ func (a *Agent) Start() {
...
+
+	if _, err := config.GetConfig().GetSection("docker"); err == nil {
+		dockerEnhancer, err := mappings.NewDockerMapper(ns)
+		if err != nil {
+			panic(err)
+		}
+		flowEnhancers = append(flowEnhancers, dockerEnhancer)
+	}
+
+	pipeline := mappings.NewFlowMappingPipeline(flowEnhancers)
```
nplanel: why not change NewFlowMappingPipeline API by a mappings.FlowEnhancer...
so the call will be, mappings.NewFlowMappingPipeline(gfe, dockerEnhancer)

agent/agent.go
```diff
@@ -97,19 +97,27 @@ func (a *Agent) Start() {
...
 	}
 
-	pipeline := mappings.NewFlowMappingPipeline([]mappings.FlowEnhancer{gfe})
+	ns := tprobes.NewNetNSProbe(a.Graph, root)
+	nl := tprobes.NewNetLinkProbe(a.Graph, root)
+	ovs := tprobes.NewOvsdbProbe(a.Graph, root, ovsmon)
+	flowEnhancers := []mappings.FlowEnhancer{gfe}
+
+	if _, err := config.GetConfig().GetSection("docker"); err == nil {
+		dockerEnhancer, err := mappings.NewDockerMapper(ns)
```
nplanel: a call like mappings.NewDockerMapperFromConfig() will be more consistant in our code

agent/agent.go
lebauce: Unfortunately, as the docker flow enhancer is optional, it doesn't make the code shorter :-/

agent/agent.go
lebauce: Indeed. Done

nplanel:
LGTM

flow/mappings/docker.go
```diff
@@ -0,0 +1,155 @@
...
+	// TODO: find a place to put the container ID
+	// ethFlow := f.GetStatistics().Endpoints[flow.FlowEndpointType_ETHERNET.Value()]
+	// mapper.getContainerFromMAC(ethFlow.AB.Value)
+	// mapper.getContainerFromMAC(ethFlow.BA.Value)
+}
+
+func (mapper *DockerMapper) handleDockerEvent(event *dockerclient.Event) {
+	info, err := mapper.client.InspectContainer(event.ID)
+	if err != nil {
+		return
```
safchain: Question: does it make sense to add a log here ?

flow/mappings/docker.go
```diff
@@ -0,0 +1,155 @@
+/*
```
safchain: I think we should move this file to topology probe as it basically retrieves topology informations. We could add metadata information about docker on the veth node ? We already have the graph mapping in order to associate flow and intf attributes.

safchain: Maybe the neutron mapping should be moved as well

PR #54, opened 2016-02-10
[flowtable] Introduce ft.SelectLayer()

nplanel:
SelectLayer(type, list) return the flows (unique) that match type layer
(Ethernet, IP, ...) with one of element in the list argurment.

### DIFF
--- a/flow/table.go
+++ b/flow/table.go
@@ -170,3 +170,31 @@ func (ft *FlowTable) FilterLast(last time.Duration) []*Flow {
 	ft.lock.RUnlock()
 	return flows
 }
+
+func (ft *FlowTable) SelectLayer(endpointType FlowEndpointType, list []string) []*Flow {
+	meth := make(map[string][]*Flow)
+	ft.lock.RLock()
+	for _, f := range ft.table {
+		layerFlow := f.GetStatistics().Endpoints[endpointType.Value()]
+		if layerFlow.AB.Value == "ff:ff:ff:ff:ff:ff" || layerFlow.BA.Value == "ff:ff:ff:ff:ff:ff" {
+			continue
+		}
+		meth[layerFlow.AB.Value] = append(meth[layerFlow.AB.Value], f)
+		meth[layerFlow.BA.Value] = append(meth[layerFlow.BA.Value], f)
+	}
+	ft.lock.RUnlock()
+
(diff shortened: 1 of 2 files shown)


flow/table.go
```diff
@@ -170,3 +170,31 @@ func (ft *FlowTable) FilterLast(last time.Duration) []*Flow {
 	ft.lock.RUnlock()
 	return flows
 }
+
+func (ft *FlowTable) SelectLayer(endpointType FlowEndpointType, list []string) []*Flow {
+	meth := make(map[string][]*Flow)
```
safchain: do we not need to lock the ft ?

nplanel: Yeah, I forgot :( I updated the PR, we need a golang static analysis better than go-lint

safchain: Yeah, that's why we review code :)

nplanel: :+1:

flow/table.go
```diff
@@ -170,3 +170,33 @@ func (ft *FlowTable) FilterLast(last time.Duration) []*Flow {
...
+		layerFlow := f.GetStatistics().Endpoints[endpointType.Value()]
+		if layerFlow.AB.Value == "ff:ff:ff:ff:ff:ff" || layerFlow.BA.Value == "ff:ff:ff:ff:ff:ff" {
+			continue
+		}
+		meth[layerFlow.AB.Value] = append(meth[layerFlow.AB.Value], f)
+		meth[layerFlow.BA.Value] = append(meth[layerFlow.BA.Value], f)
+	}
+	ft.lock.RUnlock()
+
+	mflows := make(map[*Flow]struct{})
```
safchain: What about 

	var flows []*Flow
	mflows := make(map[*Flow]struct{})
	for _, eth := range list {
		if flist, ok := meth[eth]; ok {
			for _, f := range flist {
				if _, ok := mflows[f]; !ok {
					flows = append(flows, f)
					mflows[f] = struct{}{}
				}
			}
		}
	}
        return flows

This will remove the 2d loop

nplanel: Yeah I known we can merge the uniq() with the flow list append, but it's less readable
I don't known I still divided

nplanel: done

PR #55, opened 2016-02-10
Fix termination of non connected graph client

lebauce:


topology/graph/client.go
```diff
@@ -168,9 +168,11 @@ func (c *AsyncClient) AddListener(l EventListener) {
 }
 
 func (c *AsyncClient) Disconnect() {
+	if c.IsConnected() {
+		c.quit <- true
+		c.wg.Wait()
+	}
 	c.running.Store(false)
```
nplanel: c.running.Store(false) need to be set first
the real fix is to call close(c.quit) to broadcast to all thread

c.running.Store(false)
close(c.quit)
c.wg.Wait()

topology/graph/client.go
```diff
@@ -124,7 +124,9 @@ func (c *AsyncClient) connect() {
 	go func() {
 		for {
```
nplanel: the for need to check c.running.Load()
for c.running.Load() {
...
}

topology/graph/client.go
```diff
@@ -124,7 +124,9 @@ func (c *AsyncClient) connect() {
...
 			if _, _, err := wsConn.NextReader(); err != nil {
-				c.quit <- true
+				for c.running.Load() == true {
```
nplanel: this one need to be deleted

topology/graph/client.go
```diff
@@ -122,18 +122,18 @@ func (c *AsyncClient) connect() {
 	}
 
 	go func() {
-		for {
-			if _, _, err := wsConn.NextReader(); err != nil {
-				c.quit <- true
-				return
+		for c.running.Load() == true {
```
safchain: for {, should be enough as you close the connection ?

topology/graph/client.go
```diff
@@ -169,8 +173,12 @@ func (c *AsyncClient) AddListener(l EventListener) {
 
 func (c *AsyncClient) Disconnect() {
 	c.running.Store(false)
-	c.quit <- true
+	if c.connected.Load() == true {
+		c.wsConn.Close()
+		c.connected.Store(false)
```
safchain: can be removed as Connection function will do it

topology/graph/client.go
lebauce: Yep. Fixed

safchain:
Thx for fixing this painful issue :)

topology/graph/client.go
nplanel: But if the connection is not established, you should not call wsconn.Close()

....
The logic keep changing patch after patch ....

There are 2 states : connected and running :
**connected** can turned false if the sendWSMessage() failed ... and close the connection (wsConn)
**running** should only set false by Disconnect() and all loop need to terminate smoothly (without reporting error)

the chan c.quit is used as an event to break the main waiting loop (that sendMsg) and call OnDisconnected()

What is still wrong :
   o close(c.quit) need to be called just after c.running.Store(false)  
   o the if running line 146 make no sense anymore, as close(c.quit) will send a broadcast and the loop return directly
   o line 149 (before for) a c.wg.Add(1) and defer c.wd.Done() would be great but not mandatory

safchain: > But if the connection is not established, you should not call wsconn.Close()
If the connection is not established "connected" is not set to true so no Close

> the if running line 146 make no sense anymore, as close(c.quit) will send a broadcast and the loop return directly
the running line 146 is needed as we can have a close from the line 177 and we need to know if already closed or not. Close + running false meaning do not try to connect anymore.

> o close(c.quit) need to be called just after c.running.Store(false)
no as if we don't wait that the goroutine end to close the chan the goroutine will write on a closed chan

I tested it a didn't any issue. Do we have an issue anymore ?

PR #57, opened 2016-02-11
[storage][elasticsearch] New FlatFlow format

nplanel:
o Latest FlatFlow format that can be injected in ElasticSearch.
o Inject Flow in Bulk to ES server

### DIFF
--- a/storage/elasticsearch/elasticsearch.go
+++ b/storage/elasticsearch/elasticsearch.go
@@ -35,81 +35,64 @@ import (
 	"github.com/redhat-cip/skydive/storage"
 )
 
-type OldFlow struct {
-	UUID       string
-	LayersPath string
-	EtherSrc   string
-	EtherDst   string
-	Ipv4Src    string
-	Ipv4Dst    string
-	PortSrc    uint32
-	PortDst    uint32
-	ID         uint64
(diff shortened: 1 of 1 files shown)


storage/elasticsearch/elasticsearch.go
```diff
@@ -35,101 +35,81 @@ import (
...
-		}
-
-		_, err = c.connection.Index("skydive", "flow", flow.UUID, nil, flow2OldFlow(flow))
+		err := c.indexer.Index("skydive", "flow", flow.UUID, "", "", nil, flow2FlatFlow(flow))
 		if err != nil {
 			logging.GetLogger().Error("Error while indexing: %s", err.Error())
 			continue
 		}
 	}
+	c.indexer.Stop()
```
safchain: you could move this line close to the Start with a defer

nplanel:
I understand we need to increase the timeout, because travis is sometime overloaded, but
I'm not agreed with the 5 minute timeout, as : 
 o We consume (unit second) 40 user 2 system 50 real
 o In case of Fail the CI is locked for 5min

So let's setup a relative conservative value to 90 sec or 2 min.
This will be more productive for all of us.

PR #59, opened 2016-02-11
[flow][table] Introduce Layer Conversation

nplanel:
The conversation matrix can be selected/display by Layer (Ethernet,IPv4,TCP ...)


analyzer/server.go
```diff
@@ -207,6 +221,12 @@ func (s *Server) RegisterRpcEndpoints() {
 			"/rpc/flows",
 			s.FlowSearch,
 		},
+		{
+			"LayerConversation",
+			"GET",
+			"/data/conversation.json/{layer}",
```
safchain: should be /rpc/flow/{layer}/conversation.json or something like that, meaning having the extension at the end

nplanel: Could gorilla parse that in the middle ?
Is it REST compliant ?
It's not rpc either ...

safchain: yeah gorilla supports it, take a look a the REAME. I'm ok it's not RPC but it's quite usual to put the extension at the end in order to say the output will be in such format or such format. Since we are going to probably return always json, I suggest to remove the extension!

nplanel: ok, so you're ok for :
/rpc/flow/conversation/{layer}

safchain: yep sounds good

nplanel:
done

PR #61, opened 2016-02-11
Adapt log of SFlow packet parser, log as critical only non SFlow packets

safchain:
Before this patch even the SFLow packet that was not of type Raw were
logged as critical. But this kind of packet are legitimate we just
don't support them for the moment, log them in debug.

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -135,11 +135,18 @@ func FLowsFromSFlowSample(ft *FlowTable, sample *layers.SFlowFlowSample, probePa
 	for _, rec := range sample.Records {
 
 		/* FIX(safchain): just keeping the raw packet for now */
-		record, ok := rec.(layers.SFlowRawPacketFlowRecord)
-		if !ok {
-			logging.GetLogger().Critical("1st layer is not SFlowRawPacketFlowRecord type")
+		switch rec.(type) {
+		case layers.SFlowRawPacketFlowRecord:
+			/* We only support RawPacket from SFlow probe */
+		case layers.SFlowExtendedSwitchFlowRecord:
+			logging.GetLogger().Debug("1st layer is not SFlowRawPacketFlowRecord type")
+			continue
+		default:
+			logging.GetLogger().Critical("1st layer is not a SFlow supported type")
 			continue
 		}
+
+		record := rec.(layers.SFlowRawPacketFlowRecord)
 
 		packet := &record.Header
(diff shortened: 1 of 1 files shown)


flow/flow.go
```diff
@@ -135,12 +135,18 @@ func FLowsFromSFlowSample(ft *FlowTable, sample *layers.SFlowFlowSample, probePa
 	for _, rec := range sample.Records {
 
 		/* FIX(safchain): just keeping the raw packet for now */
-		record, ok := rec.(layers.SFlowRawPacketFlowRecord)
-		if !ok {
-			logging.GetLogger().Critical("1st layer is not SFlowRawPacketFlowRecord type")
+		switch rec.(type) {
+		case layers.SFlowRawPacketFlowRecord:
+		case layers.SFlowExtendedSwitchFlowRecord:
```
nplanel: may add an explicit break for visibility for layers.SFlowRawPacketFlowRecord

flow/flow.go
```diff
@@ -135,12 +135,18 @@ func FLowsFromSFlowSample(ft *FlowTable, sample *layers.SFlowFlowSample, probePa
 	for _, rec := range sample.Records {
 
 		/* FIX(safchain): just keeping the raw packet for now */
-		record, ok := rec.(layers.SFlowRawPacketFlowRecord)
-		if !ok {
-			logging.GetLogger().Critical("1st layer is not SFlowRawPacketFlowRecord type")
+		switch rec.(type) {
+		case layers.SFlowRawPacketFlowRecord:
```
nplanel: may add an explicit break for visibility

safchain: using my C eyes I could be ok, but I'm using my Go eyes :)

nplanel: hmm in C you print a log and continue
in Go you breaking out the switch

At least put the general case ( case layers.SFlowRawPacketFlowRecord: ) at the end of the switch

safchain: Well, adding a break will add dead code. It's quite usual to have the default statement at the end, not so readable too (IHMO). If you think moving at the end "case layers.SFlowRawPacketFlowRecord" is mandatory I can do it but I'm not conviced :)

nplanel: Ok let's say add a comment like : 

case layers.SFlowRawPacketFlowRecord:
 /* We only support RawPacket from SFlow probe */

safchain: Done

PR #63, opened 2016-02-11
Fix probePath mapping, refactor test helpers add functional test

safchain:


flow/probes/sflow.go
```diff
@@ -68,41 +65,58 @@ func (probe *SFlowProbe) GetTarget() string {
...
+}
+
 func (probe *SFlowProbe) cacheUpdater() {
 	probe.wg.Add(1)
 	defer probe.wg.Done()
 
 	logging.GetLogger().Debug("Start SFlowProbe cache updater")
 
-	var index uint32
+	var index int64
```
nplanel: why not using uint64 as it seems that index is always >0

flow/probes/sflow.go
```diff
@@ -215,7 +230,7 @@ func NewSFlowProbe(a string, p int, g *graph.Graph, expire int, cleanup int) (*S
 	}
 
 	probe.cache = cache.New(time.Duration(expire)*time.Second, time.Duration(cleanup)*time.Second)
-	probe.cacheUpdaterChan = make(chan uint32, 200)
+	probe.cacheUpdaterChan = make(chan int64, 200)
```
nplanel: obviously related here too

tests/sflow_test.go
```diff
@@ -94,10 +92,12 @@ func NewTestStorage() *TestStorage {
 
 func (s *TestStorage) StoreFlows(flows []*flow.Flow) error {
 	s.lock.Lock()
+	defer s.lock.Unlock()
```
nplanel: you love defer :) I love it only when there are multiple exit path

tests/sflow_test.go
```diff
@@ -105,7 +105,19 @@ func (s *TestStorage) SearchFlows(filters storage.Filters) ([]*flow.Flow, error)
 	return nil, nil
 }
 
-func (s *TestStorage) CheckFlow(t *testing.T, f *flow.Flow, trace *flowsTraceInfo) bool {
+func (s *TestStorage) GetFlows() []*flow.Flow {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	flows := make([]*flow.Flow, 0, len(s.flows))
```
nplanel: As we known the size of the final list, why not initialized to the right value instead of 0

tests/sflow_test.go
```diff
@@ -161,6 +160,47 @@ func TestSFlowAgent(t *testing.T) {
...
+	tearDownCmds := []helper.Cmd{
+		{"ovs-vsctl del-br br-sflow", true},
+	}
+
+	helper.ExecCmds(t, setupCmds...)
+	defer helper.ExecCmds(t, tearDownCmds...)
+
+	ok := false
+	for _, f := range ts.GetFlows() {
+		if f.ProbeGraphPath != hostname+"/br-sflow" {
```
nplanel: may check if one of the flow f.LayerPath match with eth/ip/icmp

tests/sflow_test.go
```diff
@@ -161,6 +160,47 @@ func TestSFlowAgent(t *testing.T) {
...
+
+	agent, analyzer := helper.StartAgentAndAnalyzerWithConfig(t, confAgentAnalyzer, ts)
+	defer agent.Stop()
+	defer analyzer.Stop()
+
+	time.Sleep(1 * time.Second)
+	setupCmds := []helper.Cmd{
+		{"ovs-vsctl add-br br-sflow", true},
+		{"ovs-vsctl add-port br-sflow sflow-intf1 -- set interface sflow-intf1 type=internal", true},
+		{"ip address add 10.0.0.1/24 dev sflow-intf1", true},
```
nplanel: May change the subnet to a much random one, like 10.254.178.1/24   or something else...
Too many setup have already a subnet set to 10.0.0.1/xx

tests/sflow_test.go
safchain: you're right I love it !

tests/sflow_test.go
safchain: Yeah will be better indeed

tests/sflow_test.go
safchain: This was not the purpose of the test but I can add something like this

tests/sflow_test.go
safchain: If I initialized to the len and then I use append I will have at this end twice the number of element if the array.

flow/probes/sflow.go
safchain: because all the metadata are in int64

tests/sflow_test.go
safchain: Done

tests/sflow_test.go
nplanel: Yeah, I was imply this too :   (cost less than append calls)
```
i:=0
for _, f := range s.flows {
   flows[i] = f
    i++
}
```

tests/sflow_test.go
```diff
@@ -161,6 +160,47 @@ func TestSFlowAgent(t *testing.T) {
...
+	defer agent.Stop()
+	defer analyzer.Stop()
+
+	time.Sleep(2 * time.Second)
+	setupCmds := []helper.Cmd{
+		{"ovs-vsctl add-br br-sflow", true},
+		{"ovs-vsctl add-port br-sflow sflow-intf1 -- set interface sflow-intf1 type=internal", true},
+		{"ip address add 169.254.33.33/24 dev sflow-intf1", true},
+		{"ip link set sflow-intf1 up", true},
+		{"ping -c 10 -I sflow-intf1 169.254.33.34", false},
```
nplanel: may add a fake mac for .34 via arp -s, before the ping, this will generate icmp that we can test via f.LayersPath

tests/sflow_test.go
safchain: Done

tests/sflow_test.go
safchain: Well as I said before the purpose of this test is not to cover the layerpath which is cover by tests you wrote. This test is to cover the commit which fixes the issue about "probeGraph not filled". Just checking that the probeGraph is filled with the correct value whatever the layerpath value is should be enough.

nplanel: Ok, it's was more likely too make 2 tests in once and validate the all path.
Meanwhile your test could be use as the use case, gateway not configured (flood arp req, no response)

safchain: I'm ok to add more test for sure, but as this patchset fixes a bug, and it seems your pretty agree this patchset, just arguing about the fact we need to add more test, I would prefer to have this one merged and then adding more test for the layerpath, what do you think ?

PR #65, opened 2016-02-11
Increase the test timeout as we add more tests

safchain:


lebauce:
LGTM

nplanel:
I understand we need to increase the timeout, because travis is sometime overloaded, but
I'm not agreed with the 5 minute timeout, as : 
 o We consume (unit second) 40 user 2 system 50 real
 o In case of Fail the CI is locked for 5min

So let's setup a relative conservative value to 90 sec or 2 min.
This will be more productive for all of us.

safchain:
We experienced working tests over the 5 minutes. Restarting any CI job will consume more than that (I'm really writing this :).

PR #67, opened 2016-02-12
Cleanup lint etc

nplanel:


safchain:
Maybe you could squash some of the patches, like simplify expression, and unrelated comment, and simplify expression, there is no added value to have them separate in the historic

PR #70, opened 2016-02-15
Fix ES storage

safchain:
Fix elasticsearch storage which was broken, due to 

- bad usage of bulk indexer
- mapping non adapted after changing the flow format

### DIFF
--- a/storage/elasticsearch/elasticsearch.go
+++ b/storage/elasticsearch/elasticsearch.go
@@ -75,9 +75,8 @@ func flow2FlatFlow(f *flow.Flow) FlatFlow {
 	ff.IfDstGraphPath = f.IfDstGraphPath
 
 	fs := f.GetStatistics()
-	ffs := FlatFlowStatistics{}
-	ffs.Start = fs.Start
-	ffs.Last = fs.Last
+	ff.Statistics.Start = fs.Start
+	ff.Statistics.Last = fs.Last
 	for _, endp := range fs.Endpoints {
 		ffes := FlatFlowEndpointsStatistics{}
 		ffes.Type = endp.Type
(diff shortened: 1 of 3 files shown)


storage/elasticsearch/elasticsearch.go
```diff
@@ -101,8 +99,7 @@ type ElasticSearchStorage struct {
 }
 
 func (c *ElasticSearchStorage) StoreFlows(flows []*flow.Flow) error {
-	c.indexer.Start()
-	defer c.indexer.Stop()
+	defer c.indexer.Flush()
```
nplanel: It's a bad idea to call flush on all flows() .... resulting is the same as synchronous call
This code should me under indexer.Start()/Stop() like before

indexer.Flush() should be called on the ElasticSearchStorage.Close()

storage/elasticsearch/elasticsearch.go
```diff
@@ -197,13 +201,16 @@ func (c *ElasticSearchStorage) initialize() error {
...
 
-	c.indexer = c.connection.NewBulkIndexerErrors(10, 60)
-
 	return nil
 }
 
 var ErrBadConfig = errors.New("elasticseach : Config file is misconfigured, check elasticsearch key format")
 
+func (c *ElasticSearchStorage) Close() {
+	c.indexer.Stop()
```
nplanel: c.indexer.Flush() before Stop()

storage/elasticsearch/elasticsearch.go
safchain: mmm, Flush is called when Stop, so no need to call it. I hesitated to remove the Flush but I thought you wanted to have something synchronous as you used Stop! finally bulk doesn't mean async :). I will remove it anyway

storage/elasticsearch/elasticsearch.go
safchain: see my previous comment

storage/elasticsearch/elasticsearch.go
nplanel: We need a flush call, but it could be implemented in this story :

     * [F.PROBE] implement Agent.Flush() and Analyzer.Flush() ... 
Storage.Flush() for testing purpose

PR #72, opened 2016-02-15
Make use of cobra/viper

lebauce:
This pull requests introduces 2 major changes.
The first change is to use only one binary for both the agent and analyzer. Both are now commands and can be therefore be launched with 'skydive agent' and 'skydive analyzer'. This work leverages the cobra framework.

The second change changes the configuration file format from INI to YAML, using the viper framework. YAML files makes it easier to use parameters that are lists or hashes, for instance for the list of the probes that are enabled on the agent. The viper framework also allows to bind configuration to the command line so it is possible to override configuration items from the command line. It also supports a wide list of features, such as reloading configuration files when they are modified or storing configuration in etcd.


nplanel:
Just the validateConfig() topic
except that LGTM :)

safchain:
Thx for this ! 

Got error with tests. It seems that while ws-pong-timeout is not initialized

Just another comment but something that could be addressed later.

We need to define what will be in Flags or not.

lebauce:
@nplanel I think I converted the tests static config file. They failed because of "go fmt" issues... :-)

nplanel:
@lebauce rebase it and you can merge it

safchain:
LGTM, thx

PR #265, opened 2017-06-20
Full TLS support

nplanel:
Remove the old StopableListener as listener have a Close() method now and
StopableListener is not compatible with TLSListener (not exported).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -238,3 +238,12 @@ func GetEtcdServerAddrs() []string {
 	}
 	return []string{"http://localhost:2379"}
 }
+
+func IsTLSenabled() bool {
+	certPEM := GetConfig().GetString("analyzer.X509_cert")
+	keyPEM := GetConfig().GetString("analyzer.X509_key")
+	if len(certPEM) > 0 && len(keyPEM) > 0 {
+		return true
+	}
+	return false
+}
--- a/http/client.go
+++ b/http/client.go
@@ -32,6 +32,7 @@ import (
 	"net/http"
 
 	"github.com/skydive-project/skydive/common"
+	"github.com/skydive-project/skydive/config"
 )
 
 type RestClient struct {
@@ -52,8 +53,24 @@ func readBody(resp *http.Response) string {
 	return string(data)
 }
 
-func NewRestClient(addr string, port int, authOptions *AuthenticationOpts) *RestClient {
+func getHttpClient() *http.Client {
 	client := &http.Client{}
+	if config.IsTLSenabled() == true {
+		certPEM := config.GetConfig().GetString("agent.X509_cert")
+		keyPEM := config.GetConfig().GetString("agent.X509_key")
+		analyzerCertPEM := config.GetConfig().GetString("analyzer.X509_cert")
+		tlsConfig := common.SetupTLSClientConfig(certPEM, keyPEM)
+		tlsConfig.RootCAs = common.SetupTLSLoadCertificate(analyzerCertPEM)
+		checkTLSConfig(tlsConfig)
+
+		tr := &http.Transport{TLSClientConfig: tlsConfig}
+		client = &http.Client{Transport: tr}
+	}
+	return client
+}
+
+func NewRestClient(addr string, port int, authOptions *AuthenticationOpts) *RestClient {
+	client := getHttpClient()
 	authClient := NewAuthenticationClient(addr, port, authOptions)
 	return &RestClient{
 		client:     client,
--- a/http/server.go
+++ b/http/server.go
@@ -23,6 +23,7 @@
 package http
 
 import (
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"mime"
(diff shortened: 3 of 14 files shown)


statics/js/websocket.js
```diff
@@ -38,7 +38,11 @@ WSHandler.prototype = {
     });
     this.connecting = true;
 
-    this.conn = new WebSocket("ws://" + this.host + "/ws");
+    this.protocol = "ws://";
+    if (location.protocol = "https:") {
```
safchain: s/=/==/ at least not sure this works anyway with "https:"

http/server.go
```diff
@@ -52,18 +52,36 @@ type Route struct {
...
 	Port        int
 	Auth        AuthenticationBackend
 	lock        sync.Mutex
-	sl          *stoppableListener.StoppableListener
+	listener    net.Listener
+	CnxType     ConnectionType
 	wg          sync.WaitGroup
 }
 
+func isTLSenabled() bool {
```
lebauce: This function is already defined in common/

config/config.go
```diff
@@ -246,3 +246,12 @@ func GetEtcdServerAddrs() []string {
 	}
 	return []string{"http://localhost:2379"}
 }
+
+func IsTLSenabled() bool {
```
lebauce: This could be moved to common/tls.go too

I would prefer to use a boolean in the configuration to enable/disable TLS, something like "ws_use_tls: true"
That makes debugging tls easier : SKYDIVE_WS_USE_TLS=false /usr/bin/skydive allinone ...

api/api.go
```diff
@@ -71,6 +71,9 @@ func (a *APIServer) RegisterAPIHandler(handler APIHandler) error {
 			Method: "GET",
 			Path:   "/api/" + name,
 			HandlerFunc: func(w http.ResponseWriter, r *auth.AuthenticatedRequest) {
+				if r.TLS != nil {
```
safchain: moving this if statement into a http wrapper in the http package will ease to maintain the tls code. Probably wrapping the router.

nplanel:
I need to do some cleanup as functions defined twice
IsTLSEnabled() is done in config and not common because common should not include config ...
I think ws_use_tls will confuse the user, you don't want tls don't enable the server, BTW with we can't listen normal an tls flow on the same port

api/api.go
nplanel: good idea

config/config.go
nplanel: I think ws_use_tls will confuse the user, you don't want tls don't enable the server, BTW with we can't listen normal an tls flow on the same port

common/service.go
```diff
@@ -71,3 +75,45 @@ func ServiceAddressFromString(addressPort string) (ServiceAddress, error) {
 		Port: port,
 	}, nil
 }
+
+func SetupTLSLoadCertificate(certPEM string) *x509.CertPool {
```
nplanel: ok

statics/js/websocket.js
nplanel: thanks, done

config/config.go
nplanel: Futhermore wss socket will not stop you ability to debug in the browser

nplanel:
run selenium-tests

nplanel:
run skydive-selenium-tests

nplanel:
run selenium

api/api.go
nplanel: done by using the Auth wrapper

nplanel: as the Auth should set the header too

nplanel:
run tests

nplanel:
run selenium-tests

lebauce:
run selenium-tests

nplanel:
run go-fmt

nplanel:
run go-fmt

nplanel:
run go-fmt

nplanel:
run tests

config/config.go
nplanel: I mean if the configuration file have TLS X509 certificate so we use it
If you don't want .... basically we don't need an extra flag/config

scripts/scale.sh
```diff
@@ -169,10 +170,12 @@ host_id: $NAME
 ws_pong_timeout: 15
 analyzer:
   X509_cert: $ANALYZER_CRT
+  X509_key: $ANALYZER_KEY
 agent:
   listen: 0.0.0.0:8081
   X509_cert: $AGENT_CRT
   X509_key: $AGENT_KEY
+  X509_Insecure: true
```
lebauce: nit: s/509_Insecure/509_insecure/ for consistency

scripts/scale.sh
```diff
@@ -302,6 +306,8 @@ graph:
   backend: $GRAPH
 agent:
   X509_cert: $AGENT_CRT
+  X509_key: $AGENT_KEY
+  X509_Insecure: true
```
lebauce: Same as above

http/client.go
```diff
@@ -52,8 +54,29 @@ func readBody(resp *http.Response) string {
 	return string(data)
 }
 
+func getHttpClient() *http.Client {
+	var client *http.Client
```
lebauce:
```
client := &http.Client{}
if config.IsTLSEnabled() == true {
    ...
    client.Transport = &http.Transport{TLSClientConfig: tlsConfig}
}
```

PR #296, opened 2017-07-07
Openflow rules on OpenVSwitch.

pierrecregut:
The extension adds the possibility to view Openflow rules on 
OpenVSwitch bridges inside skydive. This is the **back-end part**
and will require changes to the front-end to be user-friendly.

The extension is enabled when the OVS probe is enabled and
ovsof.enabled is set to true.
Rules are displayed as nodes attached to an OpenVSwitch
bridge. Metadata include filters, actions, table and priority.
Without change to the front-end rules are displayed as
default nodes (interface).

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -124,7 +124,20 @@ ovs:
   # at least locally
   # % sudo ovs-appctl -t ovsdb-server ovsdb-server/add-remote ptcp:6400:127.0.0.1
   # ovsdb: unix:///var/run/openvswitch/db.sock
-
+  oflow:
+    # Enable the parsing of openflow rules (disabled by default)
+    enable: false
+    # The probe can connect to remote bridge over TLS (ssl url). 
+    # The default value is empty for those options.
+    # Path to the private key file (TLS connection)
+    # key: /etc/ssl/private/agent.key
+    # Path to the certificate associated to the key (TLS connection)
+    # cert: /etc/ssl/certs/agent.crt
+    # Path to certificate authority validating bridge connections (TLS connection)
+    # ca: /etc/ssl/certs/ca.crt
+    address:
+      # Map translating bridge names into URL for remote connection
+      # - bridge: ssl:xxx.yyy.zzz.ttt:port
 docker:
   # url: unix:///var/run/docker.sock
 
--- a/topology/probes/ovs_of.go
+++ b/topology/probes/ovs_of.go
@@ -1,0 +1,538 @@
+/*
+ * Copyright (C) 2017 Orange.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package probes
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"os/exec"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"golang.org/x/net/context"
+
+	uuid "github.com/nu7hatch/gouuid"
+	"github.com/skydive-project/skydive/config"
+	"github.com/skydive-project/skydive/logging"
+	"github.com/skydive-project/skydive/topology/graph"
+	"github.com/socketplane/libovsdb"
+)
+
+// OvsOfProbe is the type of the probe retrieving Openflow rules on an Open Vswitch
+type OvsOfProbe struct {
+	sync.Mutex
+	Host         string                    // The host
+	Graph        *graph.Graph              // The graph that will receive the rules found
+	Root         *graph.Node               // The root node of the host in the graph
+	BridgeProbes map[string]*BridgeOfProbe // The table of probes associated to each bridge
+	Translation  map[string]string         // A translation table to find the url for a given bridge knowing its name
+	Certificate  string                    // Path to the certificate used for authenticated communication with bridges
+	PrivateKey   string                    // Path of the private key authenticating the probe.
+	CA           string                    // Path of the certicate of the Certificate authority used for authenticated communication with bridges
+	sslOk        bool                      // cert private key and ca are provisionned.
+}
+
+// BridgeOfProbe is the type of the probe retrieving Openflow rules on a Bridge
+type BridgeOfProbe struct {
+	Host       string           // The global host
+	Bridge     string           // The bridge monitored
+	UUID       string           // The UUID of the bridge node
+	Address    string           // The address of the bridge if different from name
+	BridgeNode *graph.Node      // the bridge node on which the rule nodes are attached.
+	OvsOfProbe *OvsOfProbe      // Back pointer to the probe
+	Rules      map[string]*Rule // The set of rules found so far.
+	cancel     context.CancelFunc
+}
+
+// Rule is an OpenFlow rule in a switch
+type Rule struct {
+	Cookie   uint64 // cookie value of the rule
+	Table    int    // table containing the rule
+	Priority int    // priority of rule
+	Filter   string // all the filters as a comma separated string
+	Actions  string // all the actions (comma separated)
+	InPort   int    // -1 is any
+	UUID     string // Unique id
+}
+
+// Event is an event as monitored by ovs-ofctl monitor <br> watch:
+type Event struct {
+	Rule   *Rule  // The rule modified
+	Date   int64  // the date of the event
+	Action string // the action taken
+	Bridge string // The bridge whtere it ocured
+}
+
+// ProtectCommas substitute commas with semicolon
+// when inside parenthesis
+func protectCommas(line string) string {
+	work := []rune(line)
+	var braces = 0
+	for i, word := range work {
+		switch word {
+		case '(':
+			braces++
+		case ')':
+			braces--
+		case ',':
+			if braces > 0 {
+				work[i] = ';'
+			}
+		}
+	}
+	return string(work)
+}
+
+// fillIn is a utility function that takes a splitted rule line
+// and fills a Rule/Event structure with it
+func fillIn(components []string, rule *Rule, event *Event) {
+	for _, component := range components {
+		keyvalue := strings.SplitN(component, "=", 2)
+		if len(keyvalue) == 2 {
+			key := keyvalue[0]
+			value := keyvalue[1]
+			switch key {
+			case "event":
+				if event != nil {
+					event.Action = value
+				}
+			case "actions":
+				rule.Actions = value
+			case "table":
+				table, err := strconv.ParseInt(value, 10, 32)
+				if err == nil {
+					rule.Table = int(table)
+				} else {
+					logging.GetLogger().Errorf("Error while parsing table of rule: %s", err.Error())
+				}
+			case "cookie":
+				v, err := strconv.ParseUint(value, 0, 64)
+				if err == nil {
+					rule.Cookie = v
+				} else {
+					logging.GetLogger().Errorf("Error while parsing cookie of rule: %s", err.Error())
+				}
+			}
+		}
+	}
+
+}
+
+// extractPriority parses the filter of a rule and extracts the priority if it exists.
+func extractPriority(rule *Rule) {
+	components := strings.Split(rule.Filter, ",")
+	rule.Priority = 32768 // Default rule priority.
+	for _, component := range components {
+		keyvalue := strings.SplitN(component, "=", 2)
+		if len(keyvalue) == 2 {
+			key := keyvalue[0]
+			value := keyvalue[1]
+			switch key {
+			case "priority":
+				priority, err := strconv.ParseInt(value, 10, 32)
+				if err == nil {
+					rule.Priority = int(priority)
+				} else {
+					logging.GetLogger().Errorf("Error while parsing priority of rule: %s", err.Error())
+				}
+			}
+		}
+	}
+}
+
+type noEventError struct{}
+
+func (e *noEventError) Error() string {
+	return "No Event"
+}
+
+// parseEvent transforms a single line of ofctl monitor :watch
+// in an event. The string must be terminated by a "\n"
+// Protected commas will be replaced.
+func parseEvent(line string, bridge string, prefix string) (Event, error) {
+	var result Event
+	var rule Rule
+
+	if line[0] != ' ' {
+		return result, &noEventError{}
+	}
+	if strings.ContainsRune(line, '(') {
+		line = protectCommas(line)
+	}
+	components := strings.Split(line[1:len(line)-1], " ")
+	fillIn(components, &rule, &result)
+	if len(components) < 2 {
+		return result, errors.New("Rule syntax")
+	}
+	tentative := len(components) - 1
+	if strings.HasPrefix(components[tentative], "actions=") {
+		tentative = tentative - 1
+	}
+	if !strings.HasPrefix(components[tentative], "cookie=") {
+		rule.Filter = components[tentative]
+	}
+	result.Rule = &rule
+	fillUUID(&rule, prefix)
+	result.Date = time.Now().Unix()
+	result.Bridge = bridge
+	return result, nil
+}
+
+// Generates a unique UUID for the rule
+// prefix is a unique string per bridge using bridge and host names.
+func fillUUID(rule *Rule, prefix string) {
+	id := prefix + rule.Filter + "-" + string(rule.Table) + "-" + string(rule.Cookie)
+	u, err := uuid.NewV5(uuid.NamespaceOID, []byte(id))
+	if err == nil {
+		rule.UUID = u.String()
+	}
+}
+
+// parseRule transforms a single line of ofctl dump-flow in a rule.
+// The line DOES NOT include the terminating newline. Protected commas will be replaced.
+func parseRule(line string) (Rule, error) {
+	var rule Rule
+	if len(line) == 0 || line[0] != ' ' {
+		return rule, errors.New("No rule: " + line)
+	}
+	if strings.ContainsRune(line, '(') {
+		line = protectCommas(line)
+	}
+	components := strings.Split(line[1:], ", ")
+	if len(components) < 2 {
+		return rule, errors.New("Rule syntax")
+	}
+	fillIn(components, &rule, nil)
+	tail := components[len(components)-1]
+	components = strings.Split(tail, " actions=")
+	if len(components) == 2 {
+		rule.Filter = components[0]
+		rule.Actions = components[1]
+	} else {
+		return rule, errors.New("Rule syntax split filter and actions")
+	}
+	return rule, nil
+}
+
+func makeFilter(rule *Rule) string {
+	if rule.Filter == "" {
+		return fmt.Sprintf("table=%d", rule.Table)
+	}
+	return fmt.Sprintf("table=%d,%s", rule.Table, rule.Filter)
+}
+
+// Execute exposes an interface to command launch on the OS
+type Execute interface {
+	ExecCommand(string, ...string) ([]byte, error)
+	ExecCommandPipe(context.Context, string, ...string) (io.Reader, error)
+}
+
+// RealExecute is the actual implementation given below. It can be overriden for tests.
+type RealExecute struct{}
+
+var executor Execute = RealExecute{}
+
+// ExecCommand executes a command on a host
+func (r RealExecute) ExecCommand(com string, args ...string) ([]byte, error) {
+	/* #nosec */
+	command := exec.Command(com, args...)
+	return command.CombinedOutput()
+}
+
+// ExecCommandPipe executes a command on a host and gives back a pipe to control it.
+func (r RealExecute) ExecCommandPipe(ctx context.Context, com string, args ...string) (io.Reader, error) {
+	/* #nosec */
+	command := exec.Command(com, args...)
+	out, err := command.StdoutPipe()
+	command.Stderr = command.Stdout
+	if err != nil {
+		return out, err
+	}
+	err = command.Start()
+	go func() {
+		<-ctx.Done()
+		erk := command.Process.Kill()
+		if erk != nil {
+			logging.GetLogger().Errorf("Cannot kill background process: %s", com)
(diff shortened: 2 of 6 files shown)


safchain:
Thanks for the PR Pierre :)

safchain:
ok to test

safchain:
add to whitelist

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+
+// completeRule completes the rule by looking at it again but with dump-flows. This gives back more elements such as priority.
+func completeRule(o *OvsOfProbe, event *Event) error {
+	oldrule := event.Rule
+	bridge := event.Bridge
+	// We want exactly n+1 items where n was the number of items in old filters. The reason is that now
+	// the priority is provided. Another approach would be to use the shortest filter as it is the more generic
+	expected := countElements(oldrule.Filter) + 1
+	filter := makeFilter(oldrule)
+	command, err1 := o.makeCommand([]string{"ovs-ofctl", "dump-flows"}, bridge, filter)
```
safchain: I would prefer to not rely on command line execution, is there any enough openflow library that could be used instead.

topology/probes/ovs_of_test.go
```diff
@@ -0,0 +1,266 @@
...
+	result := r.Results[r.count]
+	r.count++
+	return []byte(result), nil
+}
+
+func (r ExecuteForTest) ExecCommandPipe(ctx context.Context, com string, args ...string) (io.Reader, error) {
+	return strings.NewReader(r.Flow), nil
+}
+
+func compareSlice(a []string, b []string) bool {
```
safchain: deepEqual doesn't do the same thing ?

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+	id := prefix + rule.Filter + "-" + string(rule.Table) + "-" + string(rule.Cookie)
+	u, err := uuid.NewV5(uuid.NamespaceOID, []byte(id))
+	if err == nil {
+		rule.UUID = u.String()
+	}
+}
+
+// ParseRule transforms a single line of ofctl dump-flow in a rule.
+// The line DOES NOT include the terminating newline. Protected commas will be replaced.
+func ParseRule(line string) (Rule, error) {
```
safchain: parseRule

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+
+// fillIn is a utility function that takes a splitted rule line
+// and fills a Rule/Event structure with it
+func fillIn(components []string, rule *Rule, event *Event) {
+	for _, component := range components {
+		keyvalue := strings.SplitN(component, "=", 2)
+		if len(keyvalue) == 2 {
+			key := keyvalue[0]
+			value := keyvalue[1]
+			switch key {
```
safchain: you are not parsing the priority which could be useful ?

pierrecregut: Because unfortunately it is not provided when you listen to rule changes. You have to get the priority from another query on the installed rules.

topology/probes/ovs_of.go
pierrecregut: Yes because OVS format are different for each kind of tool it is hard to share it between modules.

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+type RealExecute struct{}
+
+var executor Execute = RealExecute{}
+
+// ExecCommand executes a command on a host
+func (r RealExecute) ExecCommand(com string, args ...string) ([]byte, error) {
+	var command *exec.Cmd
+	/* #nosec */
+	command = exec.Command(com, args...)
+	bytes, err := command.CombinedOutput()
```
pierrecregut: ok. Was there because of a debug output on bytes that I removed.

topology/probes/ovs_of.go
pierrecregut: This is a tough question because parsing informal string output is never a great option but:
- we need to parse any kind of openflow construct that may appear
- we only need to understand a small subset of the output given to the end-user
- there are a lot of openflow extensions used by OpenVSwitch and I do not know any Openflow library that understands all of them
- go has some Openflow libraries but not a great one that would be at least fully compliant with last version of the standard.
-  the way we listen to openflow rules changes (ovs-ofctl monitor) is as far as I am aware of an OVS extension. This is probably why they missed the priority in the output.

So I think it is hard to avoid the parsing of outputs. It will be the same for the output of ofproto/trace

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+		"Type":    "ofrule",
+		"Name":    name,
+		"cookie":  fmt.Sprintf("0x%x", rule.Cookie),
+		"table":   rule.Table,
+		"filters": rule.Filter,
+		"actions": rule.Actions,
+		"UUID":    rule.UUID,
+	}
+	ruleNode := g.NewNode(graph.GenID(), metadata)
+	g.Link(bridgeNode, ruleNode, graph.Metadata{"RelationType": "ownership"})
```
pierrecregut: Rules evolve independently. If you add them as metadata you loose this property and make the front-end much more complex because it will just see that something changed not which rule changed and how. Furthermore you will increase the trafic a lot if you have a lot of rules on a bridge.

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+			event, err := ParseEvent(line, probe.Bridge, prefix)
+			if err == nil {
+				switch event.Action {
+				case "ADDED":
+					err = completeRule(ofp, &event)
+					if err != nil {
+						logging.GetLogger().Error(err.Error())
+					}
+					rule := event.Rule
+					_, here := probe.Rules[rule.UUID]
```
pierrecregut: ok.

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+	return commandLine, nil
+}
+
+// OnOvsBridgeAdd is called when a bridge is added
+func (o *OvsOfProbe) OnOvsBridgeAdd(uuid string, row *libovsdb.RowUpdate) {
+	o.Lock()
+	defer o.Unlock()
+	bridgeName := row.New.Fields["name"].(string)
+	hostName := o.Host
+	_, ok := o.BridgeProbes[bridgeName]
```
pierrecregut: ok.

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+		logging.GetLogger().Errorf("Cannot add bridge %s@%s", bridgeName, hostName)
+	}
+}
+
+// OnOvsBridgeDel is called when a bridge is deleted
+func (o *OvsOfProbe) OnOvsBridgeDel(uuid string, row *libovsdb.RowUpdate) {
+	o.Lock()
+	defer o.Unlock()
+	bridgeName := row.New.Fields["name"].(string)
+	bridgeProbe := o.BridgeProbes[bridgeName]
```
pierrecregut: yes right !

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+func NewOvsOfProbe(g *graph.Graph, root *graph.Node, host string) *OvsOfProbe {
+	enable := config.GetConfig().GetBool("ovsof.enable")
+	if !enable {
+		return nil
+	}
+	translate := config.GetConfig().GetStringMapString("ovsof.address")
+	cert := config.GetConfig().GetString("ovsof.certificate")
+	pk := config.GetConfig().GetString("ovsof.key")
+	ca := config.GetConfig().GetString("ovsof.ca")
+	sudo := config.GetConfig().GetBool("ovsof.sudo")
```
pierrecregut: Yes may be not. But I have always wondered if we could find a configuration where root would not be needed as it would be better if we could limit the right of agents to a less powerful user if we want to deploy it on production clouds. At least sudo can be limited to a given command.

topology/probes/ovsdb.go
```diff
@@ -46,6 +46,7 @@ type OvsdbProbe struct {
 	Graph        *graph.Graph
 	Root         *graph.Node
 	OvsMon       *ovsdb.OvsMonitor
+	OvsOfProbe   *OvsOfProbe
```
pierrecregut: view answer above.

topology/probes/ovs_of_test.go
pierrecregut: ok.

topology/probes/ovs_of.go
safchain: "we need to parse any kind of openflow construct that may appear" 
that exactly why it is better to use the protocol way as protocols have versions not the command line.

OK so let's start with this first implementation we will recheck the Openflow way later.

topology/probes/ovs_of.go
safchain: In general I prefer to not adapt the backend to ease the development of the frontend. I'm just thinking about the API, how we will request this information.
 For the traffic maybe but if you have 30 rules you will get one update with the metadata and 30 for the node/link but yes will low down the traffic on rule update. 

That's said I'm OK with this approach :)

topology/probes/ovs_of.go
safchain: So, agents need to be root for namespaces so basically for most of deployments. As root is a requirement for the moment I propose to not introduce another config parameter.

topology/probes/ovsdb.go
safchain: Sorry Pierre which answer ? my comment was referring to the line 451

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+		}
+	} else {
+		commandLine = append(commandLine, bridge)
+	}
+	commandLine = append(commandLine, args...)
+	return commandLine, nil
+}
+
+// OnOvsBridgeAdd is called when a bridge is added
+func (o *OvsOfProbe) OnOvsBridgeAdd(uuid string, row *libovsdb.RowUpdate) {
```
pierrecregut: What is the gain ? The probe is really a subprobe. It cannot work without the information supplied on the topology by the ovsdb probe. A regular listener will just be slower because it will listen to all node events. At least, we are sure that the initial configuration of the bridge has been analyzed.

topology/probes/ovsdb.go
pierrecregut: Sorry forgot to commit my answer on using a graph listener. Using OvsMonitor would be the worst solution. I want to know that OvsdbProbe.OnOvsBridgeProbe is finished. And the best way to do that is calling my callback at the end of the other one. Using the OvsMonitor, I would not even be sure I am called after the ovsdb probe.

topology/probes/ovs_of.go
safchain: Right, It was to avoid to lookup the graph for the bridge node on each add/delRule, but maybe that something that can be done here passing the bridgeNode to the NewBridgeProbe function ?

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,503 @@
...
+	return nil
+}
+
+// addRule adds a rule to the graph and links it to the bridge.
+func addRule(rule *Rule, probe *BridgeOfProbe) {
+	g := probe.OvsOfProbe.Graph
+	g.Lock()
+	defer g.Unlock()
+	bridgeNode := probe.BridgeNode
+	if bridgeNode == nil {
```
safchain: it seems this is always true ?

pierrecregut: Yes It is missing the back assignment to probe.BridgeNode for the memoization to work.

topology/probes/ovs_of.go
safchain: is it not called by completeRule -> ParseRule -> fillIn ?

pierrecregut: You mean have it parsed and kept as separate meta data and not in the filter ? Ok. The status of priority is not the same as the status of cookie or table. You have to extract it from filters.

topology/probes/ovs_of.go
safchain: Or event better why not passing the bridge node directly from the ovsdb probe ? it will remove the LookupFirstNode to request a bridge no just created.

topology/probes/ovs_of.go
safchain: so that we could order by priority when requesting

pierrecregut: Ok. But I will not put it in fillIn.  priority is viewed as a filter (by ovs not by the standard) and the fact it is the first filter is accidental. There is a first level of key=values separated by spaces. An exception is filters which has no keyword. Filters are also key=value pairs but separated by comma. The syntax would have been a little less ugly if there was a filters= as there is an actions=. So the correct way to do it is to parse the filters.

topology/probes/ovs_of.go
pierrecregut: Yes, it is better to use the node built than the rowupdate. The change makes the code simpler.

safchain:
rerun unit-tests

safchain:
rerun tests

topology/probes/ovs_of.go
pierrecregut: From the point of vue of the extension and ovs-ofctl, an action or a filter is just a string separated by commas and is arlready ready to be given to the end-user. Only a few of them are parsed for giving better feedback (port filter, output on a port, drop, resubmit...) . At the level of Openflow these are just random sequence of bytes in a TLV format that MUST be parsed to be meaningful to the end-user. So we are much more dependent on changes at the protocol level than at the text level.

pierrecregut:
I have added two more commits. 
- One remove spurious error messages in the log of agents.
- The other correctly handle rules without priority (ie. with an implicit priority 32768).

topology/probes/ovs_of.go
pierrecregut: Addressed by  https://github.com/skydive-project/skydive/pull/296/commits/4bc5ad6309d57c1f34fd69548f398452c1df2d07

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,531 @@
...
+	}
+	extractPriority(oldrule)
+	if !done {
+		return fmt.Errorf("Cannot complete rule on %s@%s with filter %s", bridge, o.Host, filter)
+	}
+	return nil
+}
+
+// addRule adds a rule to the graph and links it to the bridge.
+func addRule(rule *Rule, probe *BridgeOfProbe) {
```
safchain: you pass BridgeOfProbe for addRule, delRule, monitor seems to be methods of BridgeOfProbe. why not 

func (probe *BridgeOfProbe) addRule ?

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,531 @@
...
+			if err == nil {
+				switch event.Action {
+				case "ADDED":
+					err = completeRule(ofp, &event)
+					if err != nil {
+						logging.GetLogger().Error(err.Error())
+					}
+					rule := event.Rule
+					if _, here := probe.Rules[rule.UUID]; !here {
+						probe.Rules[rule.UUID] = rule
```
safchain: seems to be not used ?

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,531 @@
...
+	g.Link(bridgeNode, ruleNode, graph.Metadata{"RelationType": "ownership"})
+}
+
+// delRule deletes a rule from the the graph.
+func delRule(rule *Rule, probe *BridgeOfProbe) {
+	g := probe.OvsOfProbe.Graph
+	g.Lock()
+	defer g.Unlock()
+
+	ruleNode := g.LookupFirstNode(graph.Metadata{"UUID": rule.UUID})
```
safchain: I think something is wrong here as you can have delete event not matching exactly the adding one, ex:

ovs-ofctl add-flow br-int table=55,ip,nw_dst=192.168.0.44,actions=drop
ovs-ofctl del-flows br-int table=55

In that case rule.UUID will differ and the rule added will remain

pierrecregut: I am not listeneing to del-flows but to what ovs-ofctl monitor watch: gives back. And it will give back the exact filter of all the rules removed each on a separate line.

topology/probes/ovs_of.go
pierrecregut: addRule is only called once (line 422)

safchain: ok got it sorry

topology/probes/ovs_of.go
safchain: mmm, I'm saying that because I got the issue, I will test this one more time but when I did test I was monitoring ovs and only seen one del event.

topology/probes/ovs_of.go
pierrecregut: addRule, delRule, monitor are private functions. There is no gain in making their call through an interface. And in fact there is a degradation, performance wise.

topology/probes/ovs_of.go
safchain: I gave another try with the same issue, I didn't investigate deeply yet but I suspect the parseEvent function.

my version of ovs 2.5.0 (fedora)

Steps to reproduce:

$ skydive client topology query --gremlin "G.v().Has('Type', 'ofrule').Count()"
1

$ ovs-ofctl add-flow br-int table=66,ip,nw_dst=192.168.0.1,actions=drop
$ skydive client topology query --gremlin "G.v().Has('Type', 'ofrule').Count()"
2

$ ovs-ofctl del-flows br-int "table=66"
$ skydive client topology query --gremlin "G.v().Has('Type', 'ofrule').Count()"
2

$ ovs-ofctl dump-flows br-int
NXST_FLOW reply (xid=0x4):
 cookie=0x0, duration=17558.233s, table=0, n_packets=3856, n_bytes=364548, idle_age=0, priority=0 actions=NORMAL

output of the monitor while doing the test:

$ ovs-ofctl monitor br-int watch:
NXST_FLOW_MONITOR reply (xid=0x2):
 event=ADDED table=0 cookie=0 actions=NORMAL
OFPT_ECHO_REQUEST (xid=0x0): 0 bytes of payload
NXST_FLOW_MONITOR reply (xid=0x0):
 event=ADDED table=66 cookie=0 ip,nw_dst=192.168.0.1
OFPT_ECHO_REQUEST (xid=0x0): 0 bytes of payload
NXST_FLOW_MONITOR reply (xid=0x0):
 event=DELETED reason=delete table=66 cookie=0 ip,nw_dst=192.168.0.1
OFPT_ECHO_REQUEST (xid=0x0): 0 bytes of payload

topology/probes/ovs_of.go
```diff
@@ -0,0 +1,531 @@
...
+					logging.GetLogger().Errorf("Error while parsing cookie of rule: %s", err.Error())
+				}
+			}
+		}
+	}
+
+}
+
+// extractPriority parses the filter of a rule and extracts the priority if it exists.
+func extractPriority(rule *Rule) {
```
safchain: priority doesn't seem to be correctly parsed, got 

"Name": "priority=0 =\u003e NORMAL"

topology/probes/ovs_of.go
safchain: just better readability (IMHO) and there is no difference in term of performance go deals perfectly with this.

topology/probes/ovs_of.go
pierrecregut: I have not removed priority from the filter. I have just filled a Priority attribute. I agree that priority as a filter member is a weird ovs notion. Note that Name is not a very useful notion (just filter => actions"). There is no real notion of Name and it is not used by my front-end code. But a node without this metadata may have not been considered as standard. Name="" ?

safchain: I would say that a node without a Name should work, if not that an issue and we should fix this in Skydive. Please tell us if you got error using node without Name

pierrecregut: I have pushed a fix with a small change in nodeTitle so that it does not fail when Name is undefined in graph-layout.js. The front-end will require a change to filter out rule nodes so that they are not displayed in the d3 view. So far I have used an explicit filter on the type (must not be 'ofrule') but I could also filter on the absence of Name which could be more extensible in the future. I have rebased to have an up-to-date front-end and it seems fine.

topology/probes/ovsdb.go
safchain: agreed

topology/probes/ovs_of.go
safchain: ok thx

topology/probes/ovs_of.go
safchain: yes it works

topology/probes/ovs_of.go
safchain: with your last fix it works for me

topology/probes/ovs_of.go
pierrecregut: Done in https://github.com/skydive-project/skydive/pull/296/commits/f293bca068724639314b1791b85ae94ec1c5ed63 but I am not convinced. The point is that these are low level functions and there is no interface to use at the caller site. For example monitor directly uses the layout of the probe, so it will not use an interface instead of the actual type for the receiver type when it calls addRule or delRule. As a result, there is no gain of abstraction converting functions to methods (even if there is no loss in performance because the type is known). The use of methods without interface is misleading.

safchain:
@pierrecregut not sure you saw my last comment because I'm not able to find it, maybe I forgot to validate it :) 
We need to support Go 1.6 for packaging purpose, it seems that you need to change a bit your patch to support this version. Please rebase your patch so that it will use the correct version of the jobs. Sorry for this change

etc/skydive.yml.default
```diff
@@ -124,7 +124,20 @@ ovs:
   # at least locally
   # % sudo ovs-appctl -t ovsdb-server ovsdb-server/add-remote ptcp:6400:127.0.0.1
   # ovsdb: unix:///var/run/openvswitch/db.sock
-
+  oflow:
+    # Enable the parsing of openflow rules (disabled by default)
+    enbable: false
```
safchain: typo: enable

lebauce:
Thanks @pierrecregut, that's very useful.

I'm a little bit concerned about using a node for every OpenFlow rule in terms of query :
to get the OpenFlow rule on a node means asking for all the children :
`G.V().Has('Name', 'br-int').Out().Has('priority', ....)`

The real advantage of using nodes and edges comes when you use actually use it as a graph. It would be nice to be able to issue queries like : give me all the OpenFlow rules that makes the packet go out on a specified OVS port. In this case, the node for the OpenFlow rule would have a child that is indeed the OVS port (with an link of type 'OpenFlow' instead of 'ownership'). Do you plan to implement such behavior in a future patch ?

My point is that having many children for a node (with no further dependance) may have a performance cost. I think it's acceptable but it would be nice if it also made the querying easier.

Otherwise, it looks good to me

pierrecregut:
Currently finding input and output ports for rules is done in the front-end. An earlier prototype of the feature written in python had it in the back-end. This could be easily moved to the back-end again to create more metadata or links. Note that we have special port values representing no port (eg. drop, resubmit) or all ports (normal, flood, all, etc). Something equivalent should be found for links or we need both metadata and links.

safchain:
@pierrecregut Hi, it looks good to me, thx for updating it. A squash and it will be ready to be merged

safchain:
rerun unit-tests

PR #297, opened 2017-07-07
Alert: Trigger alert when a successful evaluation changes

### LINKED ISSUE #281: Alert: Do not trigger the alert only one time
Currently, once an alert has been triggered, it won't be triggered anymore.

I think it would be better to trigger the alert as soon as the Gremlin query or the Javascript expression returns something different from nil/false/0.

It ends up having event driven alert very chatty, but we could workaround this behavior by only sending alert when the result of the Gremlin query/ JS expression has changed.

matrohon:
fixes #281

Change-Id: Ic8d97c36225b19e8ee045827b65f232e23b4fc76

### DIFF
--- a/alert/server.go
+++ b/alert/server.go
@@ -29,6 +29,7 @@ import (
 	"fmt"
 	"net/http"
 	"os/exec"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


lebauce:
add to whitelist

PR #301, opened 2017-07-11
logging : add zap logger

nplanel:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -104,7 +104,7 @@ func init() {
 	cfg.SetDefault("logging.backends", []string{"stderr"})
 	cfg.SetDefault("logging.level", "INFO")
 	cfg.SetDefault("logging.file.path", "/var/log/skydive.log")
-	cfg.SetDefault("logging.format", "%{color}%{time} %{id} %{shortfile} %{shortpkg} %{longfunc} > %{level:.4s} %{id:03x}%{color:reset} %{message}")
+	cfg.SetDefault("logging.encoder", "")
 
 	replacer := strings.NewReplacer(".", "_", "-", "_")
 	cfg.SetEnvPrefix("SKYDIVE")
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -165,11 +165,12 @@ logging:
     - stderr
   # - file
   # - syslog
+  # - json
   # configuration of the 'file' backend
   # file:
   #   path: /var/log/skydive.log
-  # format of the logging messages
-  # format: "%{color}%{time} %{id} %{shortfile} %{shortpkg} %{longfunc} > %{level:.4s} %{id:03x}%{color:reset} %{message}"
+  # configuration encoder could be for all backends or for specific one
+  # encoder: json
 
 auth:
   # specify the type of authentication mechanism: noauth, basic, keystone (default: noauth)
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -24,79 +24,277 @@ package logging
 
 import (
 	"fmt"
+	"log/syslog"
 	"os"
+	"runtime"
 	"strings"
 
-	"github.com/op/go-logging"
 	"github.com/skydive-project/skydive/config"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
-var logger *logging.Logger
+type Logger struct {
+	*zap.Logger
+	config zap.Config
+	id     string
+}
 
-func initLogger() (_ *logging.Logger, err error) {
-	cfg := config.GetConfig()
-	id := cfg.GetString("host_id") + ":" + cfg.GetString("logging.id")
+var logger *Logger
 
-	format := cfg.GetString("logging.format")
-	format = strings.Replace(format, "%{id}", id, -1)
+type level int
 
-	level, err := logging.LogLevel(cfg.GetString("logging.level"))
-	if err != nil {
-		return nil, err
+const (
+	CRITICAL level = iota
+	ERROR
+	WARNING
+	NOTICE
+	INFO
+	DEBUG
+)
+
+func (l *Logger) log(level level, format *string, args ...interface{}) {
+	s := l.Sugar()
+	fmt := l.id + " "
+	if format != nil {
+		fmt += *format
 	}
 
-	var backends []logging.Backend
-	var backend logging.Backend
+	switch level {
+	case CRITICAL:
+		s.Errorf(fmt, args...)
+	case ERROR:
+		s.Errorf(fmt, args...)
+	case WARNING:
+		s.Warnf(fmt, args...)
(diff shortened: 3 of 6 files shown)


logging/logging.go
```diff
@@ -24,77 +24,253 @@ package logging
...
+	fmt := l.id + " "
+	if format != nil {
+		fmt += *format
 	}
 
-	var backends []logging.Backend
-	var backend logging.Backend
+	switch level {
+	case CRITICAL:
+		s.Errorf(fmt, args)
```
safchain: args...

logging/logging.go
```diff
@@ -24,77 +24,253 @@ package logging
...
+	case ERROR:
+		s.Errorf(fmt, args)
```
safchain: ditto

logging/logging.go
```diff
@@ -24,77 +24,253 @@ package logging
...
+	case WARNING:
+		s.Warnf(fmt, args)
```
safchain: ditto

logging/logging.go
```diff
@@ -24,77 +24,253 @@ package logging
...
+		newCore,
+		zap.AddCallerSkip(2),
+		// uncomment the following line to get stacktrace on error messages
+		// zap.AddStacktrace(zapcore.ErrorLevel),
+		zap.AddStacktrace(zapcore.DPanicLevel),
+	)
+	logger = &Logger{
+		Logger: z,
+		config: c,
+		id:     cfg.GetString("host_id") + ":" + cfg.GetString("logging.id"),
```
safchain: you should append logging.id only if provided

logging/logging.go
```diff
@@ -24,77 +24,253 @@ package logging
...
+			w, err := syslog.New(syslog.LOG_CRIT, "")
 			if err != nil {
-				return nil, err
+				return err
 			}
+			backends = append(backends, zapcore.NewCore(consoleEncoder, zapcore.AddSync(w), msgPriority))
 		case "stderr":
-			backend = logging.NewLogBackend(os.Stderr, "", 0)
+			backends = append(backends, zapcore.NewCore(consoleEncoder, zapcore.Lock(os.Stderr), msgPriority))
+		case "json":
```
safchain: this is a bit weird as json log only to stdout. It seems that json should be an encoder ? maybe you can add a section to the logging one like:

logging:
  encoder: json

so that we can use json for all the backends

nplanel: It's more like you can have stderr, json, file and stderr running at the same time without interference
but your idea is good too

logging/logging.go
nplanel: didn't cost anything as it's evaluated only once and by default logging.id is ""

logging/logging.go
```diff
@@ -24,77 +24,253 @@ package logging
...
+		// zap.AddStacktrace(zapcore.ErrorLevel),
+		zap.AddStacktrace(zapcore.DPanicLevel),
+	)
+	logger = &Logger{
+		Logger: z,
+		config: c,
+		id:     cfg.GetString("host_id") + ":" + cfg.GetString("logging.id"),
 	}
-	logger.SetBackend(backendLevel)
+	defer logger.Sync()
```
nplanel: yeah just for fun during test :)

logging/logging.go
nplanel: done

logging/logging.go
nplanel: I did a mix of both ideas
logging.encoder
et par backend logging.syslog.encoder ....

comme ca tu peux choisir, je pense pas que du json dans un syslog ca a du sens, mais pq pas

logging/logging.go
safchain: my concern was not the cost but the output which will have a ':' with nothing following

etc/skydive.yml.default
```diff
@@ -175,11 +175,12 @@ logging:
...
   # - file
   # - syslog
+  # - json
   # configuration of the 'file' backend
   # file:
   #   path: /var/log/skydive.log
-  # format of the logging messages
-  # format: "%{color}%{time} %{id} %{shortfile} %{shortpkg} %{longfunc} > %{level:.4s} %{id:03x}%{color:reset} %{message}"
+  # configuration encoder could be for all backends or for specific one
+  #   encoder: json
```
safchain: not correctly indented, you added it as sub element of file,should sub element of logging

logging/logging.go
```diff
@@ -24,77 +24,262 @@ package logging
...
-	format := cfg.GetString("logging.format")
-	format = strings.Replace(format, "%{id}", id, -1)
+func initLogger() (err error) {
+	cfg := config.GetConfig()
+	backendLevel := getZapLevel(cfg.GetString("logging.level"))
 
-	level, err := logging.LogLevel(cfg.GetString("logging.level"))
-	if err != nil {
-		return nil, err
+	defaultEncoder := zapcore.NewConsoleEncoder(NewEncoderConfig())
```
safchain: encoder instead of defaultEncoder as you override it two lines after

config/config.go
```diff
@@ -103,7 +103,7 @@ func init() {
 	cfg.SetDefault("logging.backends", []string{"stderr"})
 	cfg.SetDefault("logging.level", "INFO")
 	cfg.SetDefault("logging.file.path", "/var/log/skydive.log")
-	cfg.SetDefault("logging.format", "%{color}%{time} %{id} %{shortfile} %{shortpkg} %{longfunc} > %{level:.4s} %{id:03x}%{color:reset} %{message}")
+	cfg.SetDefault("logging.encoder", "stderr")
```
safchain: stderr is not an encoder ? should be empty no ?

logging/logging.go
nplanel: it's at the beginning of the file like 
"host: the_message"  

with logging.id
"host:id the_message"

etc/skydive.yml.default
nplanel: could be both, but as default let's do it for logging

config/config.go
nplanel: yeah empty

logging/logging.go
```diff
@@ -24,79 +24,264 @@ package logging
...
+	return zapcore.EncoderConfig{
+		// Keys can be anything except the empty string.
+		TimeKey:        "ts",
+		LevelKey:       "level",
+		NameKey:        "name",
+		CallerKey:      "caller",
+		MessageKey:     "message",
+		StacktraceKey:  "stacktrace",
+		LineEnding:     zapcore.DefaultLineEnding,
+		EncodeLevel:    zapcore.CapitalLevelEncoder,
```
lebauce: s/CapitalLevelEncoder/CapitalColorLevelEncoder/

nplanel: done

Issue #304, opened 2017-07-14
Linux VLAN interfaces not linked to parent link


markgoddard:
Skydive does not show the relationship between Linux VLAN interfaces and their parent interfaces. For example, to create an interface on VLAN 42 on interface eth0:

    ip link add link eth0 name eth0.42 type vlan id 42
    ip link set eth0.42 up

If the skydive agent is run on this host, skydive analyzer will show no relationship between eth0 and eth0.42.

Tested on CentOS 7.3 and Ubuntu 16.04 using skydive-0.11.0 and master branch.

PR #305, opened 2017-07-14
Fix issue #304: orphaned VLAN interfaces

markgoddard:
Skydive does not show the relationship between Linux VLAN interfaces and
their parent interfaces. This is because the netlink topology probe
looks up the Vlan metadata field as a String, when adding the
child/parent relationship, but in fact Vlan is an integer. This change
fixes this issue by using the more generic GetField() function to look
up the Vlan metadata field.

### DIFF
--- a/topology/probes/netlink.go
+++ b/topology/probes/netlink.go
@@ -116,7 +116,7 @@ func (u *NetNsNetLinkProbe) handleIntfIsChild(intf *graph.Node, link netlink.Lin
 	}
 
 	if link.Attrs().ParentIndex != 0 {
-		if _, err := intf.GetFieldString("Vlan"); err == nil {
+		if _, err := intf.GetFieldInt64("Vlan"); err == nil {
 			u.linkIntfToIndex(intf, int64(int64(link.Attrs().ParentIndex)))
 		}
 	}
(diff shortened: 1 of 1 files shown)


topology/probes/netlink.go
```diff
@@ -116,7 +116,7 @@ func (u *NetNsNetLinkProbe) handleIntfIsChild(intf *graph.Node, link netlink.Lin
 	}
 
 	if link.Attrs().ParentIndex != 0 {
-		if _, err := intf.GetFieldString("Vlan"); err == nil {
+		if _, found := intf.GetField("Vlan"); found {
```
safchain: Maybe that would be better to fix the type of the vlan field to int64 line 375, so that we can use the GetFieldInt64 function here. This will help for Gremlin request to have the proper type.

markgoddard: Done

markgoddard:
Thanks for the review. It's my first commit to skydive, so not familiar with the conventions. I've updated the commit message as requested. The Vlan field is now converted to an int64 as recommended.

safchain:
@markgoddard Thanks for updating your PR!

PR #306, opened 2017-07-17
Build : move bash-completion out of build process

nplanel:
skydive-bash-completion.sh is embedded inside skydive binary, via skydive
bash-completion command.

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -20,10 +20,22 @@
  *
  */
 
-package main
+package cmd
 
-import "github.com/skydive-project/skydive/cmd"
+import (
+	"fmt"
 
-func main() {
-	cmd.RootCmd.GenBashCompletionFile("skydive-bash-completion.sh")
+	"github.com/spf13/cobra"
+)
+
+// BashCompletion skydive root command
+var BashCompletion = &cobra.Command{
+	Use:          "bash-completion",
+	Short:        "generate bash completion helper",
+	Long:         "generate bash completion helper (skydive-bash-completion.sh)",
+	SilenceUsage: true,
+	Run: func(cmd *cobra.Command, args []string) {
+		RootCmd.GenBashCompletionFile("skydive-bash-completion.sh")
+		fmt.Println("skydive-bash-completion.sh has been generated")
+	},
 }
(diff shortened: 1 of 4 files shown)


cmd/completion.go
```diff
@@ -20,10 +20,17 @@
...
 
-import "github.com/skydive-project/skydive/cmd"
+import "github.com/spf13/cobra"
 
-func main() {
-	cmd.RootCmd.GenBashCompletionFile("skydive-bash-completion.sh")
+// BashCompletion skydive root command
+var BashCompletion = &cobra.Command{
+	Use:          "bash-completion",
+	Short:        "generate skydive-bash-completion.sh bash helper",
```
lebauce: maybe simpling "generate bash completion helper"

cmd/completion.go
```diff
@@ -20,10 +20,17 @@
...
+	Long:         "generate skydive-bash-completion.sh bash helper",
+	SilenceUsage: true,
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		RootCmd.GenBashCompletionFile("skydive-bash-completion.sh")
```
lebauce: fmt.Printf("skydive-bash-completion.sh was generated")

cmd/completion.go
```diff
@@ -20,10 +20,17 @@
...
 
-func main() {
-	cmd.RootCmd.GenBashCompletionFile("skydive-bash-completion.sh")
+// BashCompletion skydive root command
+var BashCompletion = &cobra.Command{
+	Use:          "bash-completion",
+	Short:        "generate skydive-bash-completion.sh bash helper",
+	Long:         "generate skydive-bash-completion.sh bash helper",
+	SilenceUsage: true,
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
```
lebauce: It does nothing on my machine. You probably need to change PersistentPreRun to Run

cmd/completion.go
nplanel: done

cmd/completion.go
nplanel: done

cmd/completion.go
nplanel: done

PR #314, opened 2017-07-19
Ignore .vagrant directory

carroarmato0:
When launching virtual machines through vagrant, vagrant creates a hidden folder containing the state of those machines.

Signed-off-by: Christophe Vanlancker <christophe.vanlancker@inuits.eu>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,5 @@ vendor/gopkg.in/
 contrib/packaging/rpm/*.tar.gz
 flow/*.pb.go
 statics/bindata.go
-
+contrib/vagrant/.vagrant/
 .idea


lebauce:
I was wondering, why is this hidden folder added to the git history ?
If I create a file in the git repository but do not do "git add", this file is not added to the history

carroarmato0:
By adding the hidden folder to .gitignore, you avoid yourself or others to accidentally add stuff that don't belong in the git repo.  Prime example are people who use  **git add .**

If you use a terminal extension which displays the state of the repo, it can be annoying sometimes seeing that *something* has changes which you actually don't care about.

If you build an application from sources, you don't want to see generated artifacts as changes to your repo. :)

PR #315, opened 2017-07-20
common: Delete util for tree based keys

masco-enovance:
currently we have set and get utils for tree(a.b.c.d) based key
adding the delete for the same.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -301,6 +301,30 @@ func SetField(obj map[string]interface{}, k string, v interface{}) bool {
 	return true
 }
 
+// DelField detete a value in a tree based on dot key
+func DelField(obj map[string]interface{}, k string) interface{} {
+	components := strings.Split(k, ".")
+	o, ok := obj[components[0]]
+	if !ok {
+		return obj
+	}
+	if len(components) == 1 {
+		delete(obj, k)
+		return obj
+	}
+
+	object, ok := o.(map[string]interface{})
+	if !ok {
+		return obj
+	}
+	r := DelField(object, strings.SplitN(k, ".", 2)[1])
+	if len(r.(map[string]interface{})) == 0 {
+		delete(obj, components[0])
+	}
+
+	return obj
+}
+
 // GetField retrieve a value from a tree from the dot key like "a.b.c.d"
 func GetField(obj map[string]interface{}, k string) (interface{}, error) {
 	components := strings.Split(k, ".")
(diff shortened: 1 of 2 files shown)


masco-enovance:

topology/graph/graph.go
```diff
@@ -605,6 +605,7 @@ func (g *Graph) SetMetadata(i interface{}, m Metadata) bool {
 // DelMetadata delete a metadata to an associated edge or node
 func (g *Graph) DelMetadata(i interface{}, k string) bool {
 	var e *graphElement
+	ge := graphEvent{element: i}
```
safchain: you need to specify the type of the event with the kind field

masco-enovance: added

masco-enovance:
run selenium-tests

safchain:
run selenium-tests

masco-enovance:
run selenium-tests

safchain:
rerun tests

common/types.go
```diff
@@ -301,6 +301,32 @@ func SetField(obj map[string]interface{}, k string, v interface{}) bool {
...
+		return obj
+	}
+
+	object, ok := o.(map[string]interface{})
+	if !ok {
+		return obj
+	}
+	r := DelField(object, strings.SplitN(k, ".", 2)[1])
+	if len(r.(map[string]interface{})) != 0 {
+		obj[components[0]] = r
```
safchain: not sure you have to set obj[components[0]] to r as it should be already equal to r and since this is a pointer to need to re-set it ?
I'm even not sure that we should delete a key if the map is empty.

masco-enovance: yes. you are right. no need to assign that.
IMO no use to have an empty map, let me know is there any reason to keep the empty map

safchain: I'm OK to remove it. I found this a bit weird because you want to delete an entry and it will delete the map, but maybe for metadata it is better to remove it.

common/types.go
```diff
@@ -301,6 +301,30 @@ func SetField(obj map[string]interface{}, k string, v interface{}) bool {
...
+	if len(components) == 1 {
+		delete(obj, k)
+		return obj
+	}
+
+	object, ok := o.(map[string]interface{})
+	if !ok {
+		return obj
+	}
+	r := DelField(object, strings.SplitN(k, ".", 2)[1])
```
safchain: sorry didn't see that before but you need to check that r is map[string]interface{} before using len on it as it can be something else in case of line 318

PR #322, opened 2017-07-21
cmd: do not panic on config read error

safchain:
It is scaring to see a stack trace for
just a config file read error.


safchain:
rerun selenium-tests

safchain:
rerun selenium-tests

Issue #323, opened 2017-07-23
feature flow: enable SYN/FIN/RST capturing 


vperelmax:
Using this light-weight capturing we enable such features as tracking after a packet over time and detecting elements on the path with high latency.

vperelmax:
working on the feature

safchain:
merged

PR #299, opened 2017-07-07
flow: add raw packet to flow

safchain:

PR #324, opened 2017-07-23
flow: Enable SYN/FIN/RST capturing

vperelmax:
Capturing of SYN/FIN/RST timestamps
Additional detailes that are captured:
    1. Calculate packet length according to SEQ on TCP flows.
    2. Add support for elastic search scheme in date format for ABSynStart
    3. Use accurate clock from the packet info (and not from flow-table)
    4. Make extended capturing configurable via agent parameters in .yml
    5. Change tests to reflect changes in the tables that we create

Related issue: https://github.com/skydive-project/skydive/issues/323

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -24,6 +24,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/logging"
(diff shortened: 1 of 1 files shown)


safchain:
ok to test

safchain:
add to whitelist

safchain:
rerun tests

safchain:
Hi, a first feed back about the general approach before digging deeply into the code.

1. I think it would be better to not add specific tcp code for latency calculation. I would prefer to add timestamp for the first and second packet seen at each layer so that we can calculate latency for different kind of protocol. FlowLayer could be modified accordingly.

2. The retransmission part is really interesting but I would see that as a counter of the transport layer or a specific field.

3. I think we should not use the config in order to use or not such feature we should use the capture api so that we can active it on-demand and for specific traffic (bpf, etc). You can have a look at the PR #299 I added TableOpts that can be use for that.

4. For the synData I think that the PR #299 could be use too.

What do you think ?

PR #326, opened 2017-07-24
UI : Alerting on if capture count == 0

safchain:
Fixes Count according to Capture.ID and add warning to the UI when capture.count is equal to 0


safchain:

nplanel:
LGTM

masco-enovance:
tested by adding and removing node, working cool. thanks!

PR #332, opened 2017-07-27
New Flow proto for discussion

vperelmax:


flow/flow.proto
```diff
@@ -67,11 +67,44 @@ message ICMPLayer {
 	uint32 ID = 3;
 }
 
+//Option A 
+message TCPMetric {
+   string ABSynData = 9;
```
safchain: nit: should be bytes

flow/flow.proto
```diff
@@ -67,11 +67,44 @@ message ICMPLayer {
...
+   string BASynData = 10;
+   uint32 ABSynTTL = 11;
+   uint32 BASynTTL = 12;
+   int64 ABFinStart = 14;
+   int64 BAFinStart = 15;
+   int64 ABRstStart = 16;
+   int64 BARstStart = 17;
+   uint32 ABExpectedSeq = 18;
```
safchain: is it useful to keep this field in the DB or is it only for internal use ?

flow/flow.proto
```diff
@@ -67,11 +67,44 @@ message ICMPLayer {
...
+    uint32 BASynTTL = 12;
+    int64 ABFinStart = 14;
+    int64 BAFinStart = 15;
+    int64 ABRstStart = 16;
+    int64 BARstStart = 17;
+    uint32 ABExpectedSeq = 18;
+    uint32 BAExpectedSeq = 19;
+}
+
+//Option A
```
safchain: I would prefer the option B as metric are continuously updated and kept in the database. We could put Request/Reply in FlowLayer message, depending on the protocol we could fill them of not

PR #339, opened 2017-07-31
flow : add RTT

nplanel:
Calculated on first packet on each way like :
Latency = t1stBAPacket - t1stABPacket

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -55,6 +55,14 @@ const (
 	DefaultProtobufFlowSize = 500
 )
 
+// flowState is used internally to track states within the flow table.
+// it is added to the generated Flow struct by Makefile
+type flowState struct {
+	lastMetric       *FlowMetric
+	link1stPacket    int64
+	network1stPacket int64
+}
+
 // Packet describes one packet
 type Packet struct {
 	gopacket *gopacket.Packet
@@ -363,7 +371,7 @@ func (f *Flow) Init(key string, now int64, packet *gopacket.Packet, length int64
 	f.UpdateUUID(key, L2ID, L3ID)
 }
 
-// Update a flow metrics
+// Update a flow metrics and latency
 func (f *Flow) Update(now int64, packet *gopacket.Packet, length int64) {
 	f.Last = now
 	if updated := f.updateMetricsWithLinkLayer(packet, length); !updated {
@@ -388,7 +396,6 @@ func (f *Flow) newLinkLayer(packet *gopacket.Packet, length int64) {
 		B:        ethernetPacket.DstMAC.String(),
 		ID:       linkID(packet),
 	}
-
 	f.updateMetricsWithLinkLayer(packet, length)
 }
 
@@ -421,6 +428,14 @@ func (f *Flow) updateMetricsWithLinkLayer(packet *gopacket.Packet, length int64)
 		f.Metric.BABytes += length
 	}
 
+	if f.XXX_state.link1stPacket == 0 {
+		f.XXX_state.link1stPacket = (*packet).Metadata().Timestamp.UnixNano()
+	} else {
+		if (f.RTT == 0) && (f.Link.A == ethernetPacket.DstMAC.String()) {
+			f.RTT = (*packet).Metadata().Timestamp.UnixNano() - f.XXX_state.link1stPacket
+		}
+	}
+
 	return true
 }
 
@@ -442,7 +457,6 @@ func (f *Flow) newNetworkLayer(packet *gopacket.Packet) error {
 				ID:   uint32(layer.Id),
 			}
 		}
-
 		return f.updateMetricsWithNetworkLayer(packet)
 	}
 
@@ -463,7 +477,6 @@ func (f *Flow) newNetworkLayer(packet *gopacket.Packet) error {
 				ID:   uint32(layer.Id),
 			}
 		}
-
 		return f.updateMetricsWithNetworkLayer(packet)
 	}
 
@@ -485,6 +498,15 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
 			f.Metric.BAPackets++
 			f.Metric.BABytes += int64(ipv4Packet.Length)
 		}
+
+		// update RTT
+		if f.XXX_state.network1stPacket == 0 {
+			f.XXX_state.network1stPacket = (*packet).Metadata().Timestamp.UnixNano()
+		} else {
+			if (f.RTT == 0) && (f.Network.A == ipv4Packet.DstIP.String()) {
+				f.RTT = (*packet).Metadata().Timestamp.UnixNano() - f.XXX_state.network1stPacket
+			}
+		}
 		return nil
 	}
 	ipv6Layer := (*packet).Layer(layers.LayerTypeIPv6)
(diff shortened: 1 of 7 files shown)


nplanel:

flow/flow.go
```diff
@@ -363,13 +367,68 @@ func (f *Flow) Init(key string, now int64, packet *gopacket.Packet, length int64
...
+
+func (f *Flow) updateLatency(packet *gopacket.Packet) {
+	if f.Link != nil && f.Link.latencySet == false {
+		ethernetLayer := (*packet).Layer(layers.LayerTypeEthernet)
+		ethernetPacket, ok := ethernetLayer.(*layers.Ethernet)
+		if ok && (f.Link.A == ethernetPacket.DstMAC.String()) {
+			f.Link.Latency = (*packet).Metadata().Timestamp.UnixNano() - f.Link.Latency
+			f.Link.latencySet = true
+		}
+	}
```
vperelmax: Am I right that in the patch we assume that since the two sequential packets from two different directions are first in the flow - they must be request/reply? I am not sure if such assumption is valid... The "reply" might be something belonging to an old session within the same flow. Or for example A sends to B via two different TCP sockets SYN requests, then B replies to one of them and looses one of them. On the layer of Ethernet - what will we keep as a Latency? 
I do understand capturing first packets in TCP layer, but still the possible scenario is that A sends SYN and SYN and B replies. Still, we cannot be sure we receive SYN/ACK from B to the first or to the second packet from A. 
I would say that it is more accurate and informative to keep time-stemp for first A->B packet and first B->A packet. Still, I am not sure if for all layers it is valid to calculate based on it the latency.
General comment, I would call to it RTT rather than Latency, but maybe I miss something...
I think we should keep the whole first packet in both directions because it might give us a lot of info for diagnostics (at least we can do it as an option).
Please, tell me what you think about it because my patch in progress that captures SYN, SYN/ACK packets looks like related to your patch, so I will be glad to discuss it and come to some consensus :).
Thank you,
Lera

nplanel: Yes the patch assume Request/Response of the upper layer
In the scenario of multiple syn send by to open the connection, what is important is to capture the 1st syn and the first syn/ack, or rst/fin what ever scenario, take in account the syn packets (2nd and ...) make no much sense in latency / RTT calculation. we can call that RTT Network

I was planned in a futur patch to add the Application layer, in this way we could calculate the RTT Application, in case of http for example : the time between the first http request and 1st http response, it could be ICMP, ... 

An good measure is network health ratio that give you a good value of the current connection health
NHR = RTTApplication / RTTNetwork

I seen your patch but I think you try to address too many use-cases at once, like synflood, port closed, fake checksum ... I would prefer to address a problem by integrating an external flow enhancer (another patch under way) with an IDS like [prelude](https://www.prelude-siem.org/) or other existing tools, as there are many tcp attack or other attack already implemented and will not have a performance impact on Skydive, so only the connection problem/attack/... would be logged in the Sydive database
Meanwhile I could understand that you need the 1st rawpacket on each way, aim to debug a connection problem

nplanel: s/prelude/snort/ https://www.snort.org/

safchain:
@nplanel hi, two things

1. For state of a flow please have a look at #341

2. @vperelmax made a comment few days ago bringing interesting points. Could you please tell us what are you thoughts on this. I almost agree that we should have something analyzing more precisely the protocols as often as possible. Otherwise I'm OK with the RTT field in the metric schema.

flow/flow.proto
```diff
@@ -43,6 +43,8 @@ message FlowLayer {
   string A = 3;
   string B = 4;
   int64 ID = 5;
+  int64 RTT = 6;
+  bool RTTSet = 7;
```
lebauce: Not sure you need a boolean to keep track of this.

What about comparing RTTSet to 0 (or -1 if you want to handle the case of RTT in the same nanosecond)

nplanel: RTT is used as temporary as well for the first ABpacket timestamp, so need a boolean to keep track when the RTT is set and valid. I agree this could be done by using FlowStat

safchain: should be now removed

nplanel:
done

flow/flow.go
```diff
@@ -372,6 +380,35 @@ func (f *Flow) Update(now int64, packet *gopacket.Packet, length int64) {
...
 	}
+	f.updateRTT(packet)
+}
+
+func (f *Flow) updateRTT(packet *gopacket.Packet) {
+	if f.Link != nil && f.RTT == 0 {
+		ethernetLayer := (*packet).Layer(layers.LayerTypeEthernet)
+		ethernetPacket, ok := ethernetLayer.(*layers.Ethernet)
+		if ok && (f.Link.A == ethernetPacket.DstMAC.String()) {
+			f.RTT = (*packet).Metadata().Timestamp.UnixNano() - f.state.link1stPacket
```
safchain: I would prefer to move this in the updateMetricsWithLinkLayer and updateMetricsWithNetworkLayer, it will avoid the same cast switch/case

nplanel: done

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -67,6 +67,14 @@ const flowMapping = `
 			}
 		},
 		{
+			"rtt": {
```
safchain: you need to bump the index version

flow/flow.go
```diff
@@ -496,8 +518,18 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
 			f.Metric.BAPackets++
 			f.Metric.BABytes += int64(ipv6Packet.Length)
 		}
+
+		// update RTT
+		if f.state.network1stPacket == 0 {
+			f.state.network1stPacket = (*packet).Metadata().Timestamp.UnixNano()
+		} else {
+			if (f.RTT == 0) && (f.Network.A == ipv6Packet.DstIP.String()) {
+				f.RTT = (*packet).Metadata().Timestamp.UnixNano() - f.state.network1stPacket
```
safchain: could be nice to have a specific value when the RTT calculation is not possible ex: sFlow. What about -1 ?

flow/storage/elasticsearch/elasticsearch.go
nplanel: done, was an overlap from tcpmetics patch

flow/flow.go
nplanel: RTT 0 is not possible (valid value)
For the issue with SFlow, I just add a patch that used the original packet timestamp

nplanel:
run skydive-scale-tests

nplanel:

nplanel:

nplanel:
re-run tests

flow/flow.go
```diff
@@ -627,8 +659,8 @@ func (f *Flow) newTransportLayer(packet *gopacket.Packet) error {
 
 // PacketsFromGoPacket split original packet into multiple packets in
 // case of encapsulation like GRE, VXLAN, etc.
-func PacketsFromGoPacket(packet *gopacket.Packet, outerLength int64, t int64, bpf *BPF) *Packets {
-	flowPackets := &Packets{Timestamp: t}
+func PacketsFromGoPacket(packet *gopacket.Packet, outerLength int64, bpf *BPF) *Packets {
```
safchain: It seems you broke the replay

nplanel:
run tests

flow/flow.go
safchain: you can't remove the t argument it is used by the replay mechanism

nplanel: yeah I seen it's a big trick to force replay to works, so it's specific code for test only ...

safchain: No this is not for test purpose it is to be able to reinject with the exact same timing in order to simulate more or less real traffic

safchain:
need to be rebased :(

nplanel:
done

PR #341, opened 2017-07-31
flow: introduce flowState and use json omit for raw packet

safchain:
a flowState struct pointer is added to Flow generated struct
which can be used to keep internal flow status within the
flow table.

Metrics now makes use of this.


safchain:

PR #343, opened 2017-08-01
graph: send graph update only if metadata changes

masco-enovance:


topology/probes/netlink.go
```diff
@@ -689,6 +690,10 @@ Ready:
 						}
 						tr.Metadata["Statistics"] = u.statsToMap(stats)
 						lastMetrics := u.statsToMap(&metric)
+						if reflect.DeepEqual(tr.Metadata["Statistics"], lastMetrics) {
```
lebauce: I think you are comparing the wrong values (absolute vs relative)

You could do this instead :

changed := true
for i, v := range lastMetrics {
    if v != 0 {
        break
    } else if i == len(lastMetrics) - 1 {
        changed = false
    }
}
if !changed {
    u.Graph.Unlock()
    continue
}

masco-enovance: oops, you are correct i am comparing the wrong values. I have changed the code as you suggested with a small improvement. thanks!

PR #355, opened 2017-08-03
wsserver : don't report connection closed by client as error

nplanel:


http/wsserver.go
```diff
@@ -217,7 +217,7 @@ func (c *WSClient) readPump() {
 	for {
 		_, m, err := c.conn.ReadMessage()
 		if err != nil {
-			if e, ok := err.(*websocket.CloseError); ok && e.Code == websocket.CloseAbnormalClosure {
+			if e, ok := err.(*websocket.CloseError); ok && (e.Code == websocket.CloseAbnormalClosure || e.Code == websocket.CloseGoingAway) {
```
safchain: It is the behavior we wanted to have. Not sure we want to report an error on client left, for ex the WebUI

nplanel:
run skydive-selenium-tests

PR #358, opened 2017-08-04
ovsdb: link orphan ovs internal interface to bridge

safchain:
This occurs when an internal interface is pushed into
a namespace and the namespace is not seen by the netns
probe. In that case the netlink probe is not linking
the interface to the namespace. As the interface
can't be linked to the namespace this patch link it
to the ovs bridge.


safchain:
rerun tests

PR #361, opened 2017-08-04
Front-end for displaying Openflow rules

pierrecregut:
Modifications to the front-end required to display Openflow rules of an OVS switch in a separate tab and not as nodes of the graph.


safchain:
ok to test

safchain:
add to whitelist

statics/js/components/graph-layout.js
```diff
@@ -402,6 +409,7 @@ TopologyGraphLayout.prototype = {
   },
 
   onNodeAdded: function(node) {
+    if (this.hidden(node)) return;
```
safchain: specific field are added in  the _onNodeAdded function especially _metadata which is used at some places. Not adding it will trigger exceptions. I can either move this return in the _onNodeAdded or check where _metadata is used.

pierrecregut: It is used by onNodeUpdated. But the real solution is probably to add if(this.hidden(node)) return to both onNodeUpdated and onNodeDeleted. We want to have a tunable set of nodes completely ignored by a renderer. To make it complete I have also added it to onEdgeDeleted (remove if either end is hidden).

statics/js/components/graph-layout.js
```diff
@@ -195,7 +196,12 @@ TopologyGraphLayout.prototype = {
     }
   },
 
+  hidden: function(d) {
+    return  d.metadata.Type === "ofrule"
```
pierrecregut: agreed

statics/js/components/rule_detail.js
```diff
@@ -0,0 +1,489 @@
+/*
```
pierrecregut: done

statics/js/components/rule_detail.js
```diff
@@ -0,0 +1,489 @@
...
+        summary.port = rule.inPort;
+        break;
+    }
+    rule.outAction.push(summary);
+  }
+}
+/** Summarize the actions of a rule, filling the outAction field
+ *  @param rule: the rule to complete
+ */
+function summarize_actions(rule) {
```
pierrecregut: ok. Done for the file.

statics/js/components/rule_detail.js
```diff
@@ -0,0 +1,489 @@
...
+    $('.nav-tabs a[href="#T' + tab + '"]').tab('show');
+  };
+
+  /** switch_port_tab Changes the selected port in the UI.
+   * @param tabR the table
+   * @param tabP the port number
+   */
+  BridgeLayout.prototype.switch_port_tab = function (tabR, tabP) {
+    $('.nav-tabs a[href="#P' + tabR + '-' + tabP + '"]').tab('show');
+  }
```
pierrecregut: ok.

statics/js/components/rule_detail.js
```diff
@@ -0,0 +1,489 @@
...
+        file = 'resubmit.png';
+        break;
+      case ActionFlood:
+        file = 'flood.png';
+        break;
+      case ActionNormal:
+        file = 'flood.png';
+        break;
+      case ActionDrop:
+        file = 'drop.png';
```
pierrecregut: yes indeed !

statics/js/components/rule_detail.js
```diff
@@ -0,0 +1,489 @@
...
+
+  return BridgeLayout;
+}());
+
+/** Graphical component that represents a set of rules associated to a given
+ *  Openflow table and port and displayed as a single HTML table.
+ */
+Vue.component('rule-table-detail', {
+  template: '\
+              <table class="table table-bordered table-condensed">\
```
pierrecregut: I have adapted the look but not reused dynamic-table without documentation. I am a little afraid of using it on the summary field.

statics/js/components/rule_detail.js
```diff
@@ -0,0 +1,489 @@
...
+                        <td>\
+                            <table>\
+                                <tr v-for="act in rule.outAction">\
+                                    <td><img :src="layout.imgsrc(act.action)"\
+                                            width="16"\
+                                            height="16" /></td>\
+                                    <td v-on:mouseover="layout.mark(act.port)"\
+                                        v-on:mouseleave="layout.unmark(act.port)"\
+                                        v-on:click="layout.select(act.port)">{{layout.portname(act.port)}}</td>\
+                                    <td class="number"><a v-on:click="layout.switch_tab(act.table)">{{act.table}}</a></td>\
```
pierrecregut: Added new css styles. I have used the skyblue color used elsewhere

statics/js/components/rule_detail.js
```diff
@@ -0,0 +1,489 @@
...
+      type: Object,
+      required: true
+    }
+  },
+});
+
+/** Vue component showing the rules associated to a bridge */
+Vue.component('rule-detail', {
+  template: '\
+<div class="rules-detail">\
```
pierrecregut: done with a v-if

statics/js/components/rule_detail.js
```diff
@@ -0,0 +1,489 @@
...
+    <ul class="nav nav-tabs"\
```
pierrecregut: should be ok now (nav-pills + flow-ops-panel)

pierrecregut:
rerun selenium-tests

statics/js/components/rule-detail.js
```diff
@@ -0,0 +1,494 @@
...
+                        <tr v-for="act in rule.outAction">\
+                            <td><img :src="layout.imgsrc(act.action)"\
+                                    width="16"\
+                                    height="16" /></td>\
+                            <td v-on:mouseover="layout.mark(act.port)"\
+                                v-on:mouseleave="layout.unmark(act.port)"\
+                                v-on:click="layout.select(act.port)">\
+                                <span class="port-link">{{layout.portname(act.port)}}</span>\
+                            </td>\
+                            <td> <a class="table-link" v-on:click="layout.switchTab(act.table)">{{act.table}}</a></td>\
```
safchain: "click to table" seems to be not working anymore

statics/js/components/rule-detail.js
```diff
@@ -0,0 +1,494 @@
...
+  },
+});
+
+/** Vue component showing the rules associated to a bridge */
+Vue.component('rule-detail', {
+  template: '\
+<div class="rules-detail flow-ops-panel" v-if="Object.keys(layout.structured).length > 0">\
+    <ul class="nav nav-pills"\
+        role="tablist">\
+        <li class="nav-header"> <a>Table </a></li>\
```
safchain: I would not put this "useless tab" before it makes me think that I can click on it instead I'll prefix the others with "Table "
 like "Table 0", "Table 1" or maybe we can find something better.

statics/js/components/rule-detail.js
```diff
@@ -0,0 +1,494 @@
...
+ *  by this object
+*/
+var BridgeLayout = (function () {
+
+  /** Builds the bridge layout
+   * @param graph the global graph hosting the bridge
+   * @param bridge the bridge node whose rules are represented.
+   */
+  function BridgeLayout(graph, bridge, store) {
+    this.graph = graph;
```
safchain: Something that could be really nice would be to listen for graph event so that we could refresh the rule without to unselect/reselect the node. Nothing mandatory for now but that something we will have to do.

statics/js/components/rule-detail.js
pierrecregut: Indeed. Lost with the change from nav-tabs to nav-pills as the class was used to find the relevant link to select.

statics/js/components/rule-detail.js
pierrecregut: Then it must be done for ports too. I think it is a little heavy... Let us try again. The new code is more in the spirit of main menus with an explicit header. I used container-fluid so that we can have multiple lines when there are a lot of tables or ports.

pierrecregut:
I agree with the principle of the change but I would change a few things:
- I like having an arrow for output rather than nothing : fa-long-arrow-right 
- fa-filter does not convey the idea of flooding, rather the opposite. fa-volume-down is similar to the icon I used. As long as it is hard to think of a loud-speaker in this context, it should be ok. Or fa-rss ?

I have pushed the change with my suggestion. We can revert back if you prefer your icons. I have added another commit on top of the stack solving a previous comment.

statics/js/components/rule-detail.js
pierrecregut: Done. With also a way to unregister the callback at the end of the component lifecycle.

pierrecregut:
rerun selenium-tests

pierrecregut:

pierrecregut:
rerun functional-tests

safchain:
rerun selenium-tests

safchain:
run cdd-overview

safchain:
run cdd-overview

PR #373, opened 2017-08-10
flow enhancers : add Snort alert messages

nplanel:
Parse Snort message running with options "-A cmg -X" from a TCP socket
configured, for example snort.listen = 0.0.0.0:8888

The snort messages are saved in the flow, during the enhance step, thanks
to the flow pipeline. All snort messages are append to the flow.snort
field.

### DIFF
--- a/contrib/snort/snortSkydive.go
+++ b/contrib/snort/snortSkydive.go
@@ -1,0 +1,255 @@
+/*
+ * Copyright (C) 2017 Red Hat, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"os/signal"
+	"regexp"
+	"strconv"
+	"strings"
+	"sync/atomic"
+	"syscall"
+	"time"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	esclient "github.com/skydive-project/skydive/storage/elasticsearch"
+
+	"github.com/skydive-project/skydive/common"
+	"github.com/skydive-project/skydive/config"
+	"github.com/skydive-project/skydive/flow"
+	"github.com/skydive-project/skydive/logging"
+)
+
+// sudo -E /usr/bin/snort -A cmg -c /etc/snort/snort.lua -R snort3-community-rules/snort3-community.rules -i br-gre -X 2>/dev/null | go run contrib/snort/snortSkydive.go
+
+const snortMessageMapping = `
+{
+	"dynamic_templates": [
+		{
+			"timestamp": {
+				"match": "Timestamp",
+				"mapping": {
+					"type": "date", "format": "epoch_millis"
+				}
+			}
+		}
+	]
+}`
+
+const (
+	timestamp = iota
+	snortRAW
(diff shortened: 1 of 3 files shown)


flow/enhancers/snort.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+const sepHEX = "- -   - - - - - - - - - - - -  - - - - - - - - - - - -  - - - - - - - - -\n"
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
```
vperelmax: return (a<b) maybe?

etc/skydive.yml.default
```diff
@@ -104,6 +104,9 @@ agent:
     # stats_update: 1
   metadata:
     info: This is compute node
+  snort:
+    binary: /home/nplanel/Downloads/snort-3.0.0-a4/b/src/snort
```
mscherer: This is unused in the code, is this normal ?

flow/enhancers/snort.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+		if step == snortRAW && regexSnortRAW.MatchString(str) {
+			packetBytes, err = strconv.Atoi(regexSnortRAW.ReplaceAllString(str, "${bytes}"))
+			if err != nil {
+				return err
+			}
+			step = decodeHEX
+			continue
+		}
+		if step == decodeHEX {
```
mscherer: Wouldn't it open skydive to a potential dos ? For example, since the connection is not authenticated, a attacker could just connect, and feed skydive with a crafted message of unlimited hexadecimal line, resulting in a growing msg.Data until there is OOM and skydive get killed.

flow/enhancers/snort.go
```diff
@@ -0,0 +1,200 @@
...
+// readCMGX : read snort log with running with "-A cmg -X" options
+func (sfe *SnortFlowEnhancer) parseSnortCMGX(reader *bufio.Reader) error {
+	regexTimestamp := regexp.MustCompile("^(?P<timestamp>\\d\\d/\\d\\d-\\d\\d:\\d\\d:\\d\\d\\.\\d\\d\\d\\d\\d\\d) \\[.+?\\] \\[.+?\\] \\\"(?P<message>.+?)\\\" \\[.+?\\] \\[(?P<classification>.+?)\\].*\n")
+	regexSnortRAW := regexp.MustCompile("^snort\\.raw\\[(?P<bytes>\\d+)\\]:\n")
+
+	var packetBytes int
+	var msg *flow.SnortMessage
+	step := timestamp
+	for {
+		str, err := reader.ReadString('\n')
```
mscherer: Connection do not expire ever, isn't it also a issue, since that can lead to ressources exhaustion (since the socket are kept open until "something happen") ?

mscherer:
So, something that was discussed with @lebauce in the office (after he asked "hey, you likely know snort") is that what should happen for alerts not related to a specific flow. For example, snort has a preprocessor to detect portscan, who may not be linked to a specific flow.

Also, there is no doc (which I expect for a WIP, but still took us a while to figure what was the goal)

flow/enhancers/snort.go
nplanel: (a<b) return a boolean

etc/skydive.yml.default
nplanel: Yes I forgot to removed it

flow/enhancers/snort.go
nplanel: technically more or less yes, but take in account the cache will force timeout the entires that didn't match for amount of time.

flow/enhancers/snort.go
nplanel: The whole patch is more likely a PoC, so yeah the socket never timeout if client connection still exist.
The patch have some limitation, as it's related to a packet event on snort side, all is done is correlate packet snort event with skydive. Some snort event can't be correlate for the moment, like mac address correlation

contrib/snort/snortSkydive.go
```diff
@@ -0,0 +1,252 @@
...
+
+	go sfe.run()
+}
+
+func (sfe *SnortFlowEnhancer) run() {
+	reader := bufio.NewReader(os.Stdin)
+	err := sfe.parseSnortCMGX(reader)
+	if err != nil {
+		logging.GetLogger().Errorf("parse error : %v", err)
+		return
```
safchain: quit chan will never be filled

nplanel: done

nplanel:
run tests

nplanel:
run skydive-selenium-tests

nplanel:
run skydive-selenium-tests

nplanel:
run skydive-scale-tests

nplanel:
run skydive-scale-tests

PR #377, opened 2017-08-10
neutron: do not reuse metadata map as value.

safchain:
We shouldn't use a map as metadata value or
we shouldn't modify it after commit since
a map is a pointer modifying it can be reported
to the datastore especially in case of bulk
operations.

This patch increases a bit the error chan.


safchain:
run devstack-plugin

safchain:
run devstack

safchain:
run devstack

PR #379, opened 2017-08-11
ui: do not attach openstack vm tab to host

safchain:
This kind of interface are already linked to
the host through the chain of interfaces bridges,
moreover it makes the topology more readable.

### DIFF
--- a/statics/js/components/graph-layout.js
+++ b/statics/js/components/graph-layout.js
@@ -353,6 +353,8 @@ TopologyGraphLayout.prototype = {
     link.target.links[link.id] = link;
 
     if (link.metadata.RelationType == "ownership") {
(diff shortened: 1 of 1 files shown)


statics/js/components/graph-layout.js
```diff
@@ -609,6 +611,14 @@ TopologyGraphLayout.prototype = {
       .attr("xlink:href", this.managerImg(d));
   },
 
+  isNeutronNode: function(d) {
```
masco-enovance: nit- why separate function only for this? not for tun? is this will be used in many places in future?

safchain: yeah I thought we could use it later but maybe overkill for such comparison, will remove it

PR #380, opened 2017-08-11
probe: add route-table probe

masco-enovance:
this will display the route table info
in the metadata.

### DIFF
--- a/tests/topology_test.go
+++ b/tests/topology_test.go
@@ -909,3 +909,101 @@ func TestAgentMetadata(t *testing.T) {
 
 	RunTest(t, test)
 }
+
+//TestRouteTable tests route table update
+func TestRouteTable(t *testing.T) {
+	gopath := os.Getenv("GOPATH")
+	topology := gopath + "/src/github.com/skydive-project/skydive/scripts/simple.sh"
+
+	test := &Test{
+		mode: OneShot,
+
+		setupCmds: []helper.Cmd{
+			{fmt.Sprintf("%s start 124.65.91.42/24 124.65.92.43/24", topology), true},
+		},
+
+		tearDownCmds: []helper.Cmd{
+			{fmt.Sprintf("%s stop", topology), true},
+		},
+
+		checks: []CheckFunction{
+			func(c *CheckContext) error {
+				prefix := "g"
+				if !c.time.IsZero() {
+					prefix += fmt.Sprintf(".Context(%d)", common.UnixMillis(c.time))
+				}
+
+				node, err := c.gh.GetNode(prefix + ".V().Has('IPV4', '124.65.91.42/24')")
+				if err != nil {
+					return fmt.Errorf("Failed to find a node with IP 124.65.91.42/24")
+				}
+
+				routingTable := node.Metadata()["RoutingTable"].([]interface{})
+				noOfRoutingTable := len(routingTable)
+
+				helper.ExecCmds(t,
+					helper.Cmd{Cmd: "ip netns exec vm1 ip route add 124.65.92.0/24 via 124.65.91.42 table 2", Check: true},
+					helper.Cmd{Cmd: "sleep 5", Check: false},
+				)
+
+				node, err = c.gh.GetNode(prefix + ".V().Has('IPV4', '124.65.91.42/24')")
+				routingTable = node.Metadata()["RoutingTable"].([]interface{})
+				newNoOfRoutingTable := len(routingTable)
+
+				helper.ExecCmds(t,
+					helper.Cmd{Cmd: "ip netns exec vm1 ip route del 124.65.92.0/24 via 124.65.91.42 table 2", Check: true},
+					helper.Cmd{Cmd: "sleep 5", Check: false},
+				)
+				if newNoOfRoutingTable <= noOfRoutingTable {
+					return fmt.Errorf("Failed to add Route")
+				}
+				return nil
+			},
+		},
+	}
+	RunTest(t, test)
+}
+
+//TestRouteTableHistory tests route table update available in history
+func TestRouteTableHistory(t *testing.T) {
+	gopath := os.Getenv("GOPATH")
+	topology := gopath + "/src/github.com/skydive-project/skydive/scripts/simple.sh"
+
+	test := &Test{
+		mode: OneShot,
+
+		setupCmds: []helper.Cmd{
+			{fmt.Sprintf("%s start 124.65.75.42/24 124.65.76.43/24", topology), true},
+			{"sleep 5", false},
+			{"ip netns exec vm1 ip route add 124.65.75.0/24 via 124.65.75.42 table 2", true},
+		},
+
+		tearDownCmds: []helper.Cmd{
+			{fmt.Sprintf("%s stop", topology), true},
+		},
+
+		checks: []CheckFunction{
+			func(c *CheckContext) error {
+				prefix := fmt.Sprintf("g.Context(%d)", common.UnixMillis(time.Now()))
+				node, err := c.gh.GetNode(prefix + ".V().Has('IPV4', '124.65.75.42/24')")
+				if err != nil {
+					return fmt.Errorf("Failed to find a node with IP 124.65.75.42/24")
+				}
+				routingTable := node.Metadata()["RoutingTable"].([]interface{})
+				foundNewTable := false
+				for _, obj := range routingTable {
+					rt := obj.(map[string]interface{})
+					if (rt["Id"].(json.Number)).String() == "2" {
+						foundNewTable = true
+						break
+					}
+				}
+				if !foundNewTable {
+					return fmt.Errorf("Failed to get added Route from history")
+				}
+				return nil
+			},
+		},
+	}
+	RunTest(t, test)
+}
--- a/topology/probes/netlink.go
+++ b/topology/probes/netlink.go
@@ -87,6 +87,25 @@ type NetLinkProbe struct {
 	wg      sync.WaitGroup
 }
 
+// RouteTable describes a list of Routes
+type RoutingTable struct {
+	Id     int
+	Src    net.IP `json:"Src,omitempty"`
+	Routes []Route
+}
+
+// Route describes a route
+type Route struct {
+	Prefix   string    `json:"Prefix,omitempty"`
+	Nexthops []NextHop `json:"Nexthops,omitempty"`
+}
+
+// NextHop describes a next hop
+type NextHop struct {
+	Priority int    `json:"Priority,omitempty"`
+	Ip       net.IP `json:"Ip,omitempty"`
+}
+
 func (u *NetNsNetLinkProbe) linkPendingChildren(intf *graph.Node, index int64) {
 	// ignore ovs-system interface as it doesn't make any sense according to
 	// the following thread:
@@ -410,6 +429,10 @@ func (u *NetNsNetLinkProbe) addLinkToTopology(link netlink.Link) {
 	neighbors = append(neighbors, u.getNeighbors(attrs.Index, syscall.AF_INET6)...)
 	metadata["Neighbors"] = neighbors
 
+	if rt := u.getRoutingTable(link, syscall.RTA_UNSPEC); rt != nil {
+		metadata["RoutingTable"] = rt
+	}
+
 	if statistics := link.Attrs().Statistics; statistics != nil {
 		metadata["Statistics"] = u.statsToMap(statistics)
 	}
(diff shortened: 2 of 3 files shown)


topology/probes/netlink.go
```diff
@@ -346,6 +346,32 @@ func (u *NetNsNetLinkProbe) addLinkToTopology(link netlink.Link) {
...
+			routes[i] = map[string]interface{}{
+				"Src":   route.Src.String(),
+				"Dst":   dst,
+				"GW":    gw,
+				"Table": int64(route.Table),
+			}
+		}
+		metadata["Routes"] = routes
+	}
+
```
vperelmax: The code looks fine. 
Just a question, do we have unit tests for the netlink stuff? If we do, I think it will be a good idea to add some Unit test for this feature too.

vperelmax: Well, now I see that route.MPLSDst doesn't exist :(
(just added the code to my source...)

masco-enovance:
run tests

topology/probes/netlink.go
```diff
@@ -74,6 +74,26 @@ type NetLinkProbe struct {
...
 
+// RouteTable descripes a list of Routes
+type RouteTable struct {
+	id      int
+	ifIndex int
+	src     net.IP
+	routes  []*Route
+}
+
+// Route descripes a route
```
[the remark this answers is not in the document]
lebauce: You should really only store int64 and string in metadata.
To store the routes, you may also need to convert to []map[string]interface{}

This is required for Gremlin querying

topology/probes/netlink.go
```diff
@@ -74,6 +74,26 @@ type NetLinkProbe struct {
...
+type Route struct {
+	prefix   *net.IPNet
```
lebauce: same as above

topology/probes/netlink.go
```diff
@@ -346,6 +366,30 @@ func (u *NetNsNetLinkProbe) addLinkToTopology(link netlink.Link) {
...
+				nhop := new(NextHop)
+				nhop.ip = nexthop.Gw
+				nhop.priority = route.Priority
+				nh = append(nh, nhop)
+			}
+			r.nexthops = nh
+			routes = append(routes, r)
+		}
+		routeTable.routes = routes
+		metadata["RouteTable"] = routeTable
```
lebauce: Maybe "RoutingTable" instead ?

lebauce:
nit: can you change the commit message as it's not really a new probe in the Skydive jargon

topology/probes/netlink.go
```diff
@@ -74,6 +74,26 @@ type NetLinkProbe struct {
 	wg      sync.WaitGroup
 }
 
+// RouteTable describes a list of Routes
+type RouteTable struct {
```
safchain: RoutingTable ?

topology/probes/netlink.go
```diff
@@ -346,6 +366,30 @@ func (u *NetNsNetLinkProbe) addLinkToTopology(link netlink.Link) {
 		metadata["FDB"] = neighbors
 	}
 
+	routeList, err := u.handle.RouteList(link, netlink.FAMILY_ALL)
```
safchain: maybe put this into a specific function

safchain: maybe neighbor stuff should be moved too (another commit/patch)

masco-enovance:
rerun selenium-tests

masco-enovance:
rerun scale-tests

topology/probes/netlink.go
```diff
@@ -492,6 +540,23 @@ func getFamilyKey(family int) string {
 	return ""
 }
 
+func (u *NetNsNetLinkProbe) onRouteChanged(index int, route *RouteTable) {
+	u.Graph.Lock()
+	defer u.Graph.Unlock()
+
+	intf := u.Graph.LookupFirstChild(u.Root, graph.Metadata{"IfIndex": index})
```
masco-enovance: here I am getting nil. I have simple.sh script's topology. used this cmd to add a default route "sudo ip netns exec vm2 route add default gw 192.168.10.18 eth0"

topology/probes/netlink.go
```diff
@@ -492,6 +540,23 @@ func getFamilyKey(family int) string {
 	return ""
 }
 
+func (u *NetNsNetLinkProbe) onRouteChanged(index int, route *RoutingTable) {
+	u.Graph.Lock()
+	defer u.Graph.Unlock()
+
+	intf := u.Graph.LookupFirstChild(u.Root, graph.Metadata{"IfIndex": index})
```
masco-enovance: here I am getting nil. I have simple.sh script's topology. used this cmd to add a default route "sudo ip netns exec vm2 route add default gw 192.168.10.18 eth0"

safchain: I think you have to convert index to uint64 uint64(index)

topology/probes/netlink.go
```diff
@@ -769,6 +857,15 @@ func (u *NetNsNetLinkProbe) onMessageAvailable() {
 				continue
 			}
 			u.onAddressDeleted(addr, family, int64(ifindex))
+		case syscall.RTM_NEWROUTE:
+			fallthrough
+		case syscall.RTM_DELROUTE:
```
safchain: nit: case syscall.RTM_NEWROUTE, syscall.RTM_DELROUTE

topology/probes/netlink.go
```diff
@@ -492,6 +540,23 @@ func getFamilyKey(family int) string {
 	return ""
 }
 
+func (u *NetNsNetLinkProbe) onRouteChanged(index int64, route *RoutingTable) {
```
safchain: nit : could be nice to use "rt" for routing table and "route" for a route

topology/probes/netlink.go
```diff
@@ -79,6 +79,26 @@ type NetLinkProbe struct {
 	wg      sync.WaitGroup
 }
 
+// RouteTable describes a list of Routes
+type RoutingTable struct {
+	Id      int
+	IfIndex int
```
safchain: not sure we need to keep this as we attach the RT to the node which already as a IfIndex ?

topology/probes/netlink.go
```diff
@@ -79,6 +79,26 @@ type NetLinkProbe struct {
...
+	Src     net.IP
+	Routes  []Route
+}
+
+// Route describes a route
+type Route struct {
+	Prefix   net.IP
+	Nexthops []NextHop
```
safchain: I'm not able to see the nexthops in the WebUI for instance

topology/probes/netlink.go
```diff
@@ -79,6 +79,26 @@ type NetLinkProbe struct {
 	wg      sync.WaitGroup
 }
 
+// RouteTable describes a list of Routes
+type RoutingTable struct {
+	Id      int
+	IfIndex int
+	Src     net.IP
```
safchain: you should add 'omitempty' where it is needed so that no empty field indexed and reported by the API

topology/probes/netlink.go
masco-enovance: done

topology/probes/netlink.go
masco-enovance: done

topology/probes/netlink.go
masco-enovance: agree and removed

topology/probes/netlink.go
masco-enovance: done

topology/probes/netlink.go
masco-enovance: it was displayed only if it has multipath, now it fixed. thanks for pointing it out.

topology/probes/netlink.go
```diff
@@ -437,6 +460,35 @@ func (u *NetNsNetLinkProbe) addLinkToTopology(link netlink.Link) {
...
 
+func (u *NetNsNetLinkProbe) getRoutingTable(link netlink.Link) *RoutingTable {
+	routeList, err := u.handle.RouteList(link, netlink.FAMILY_ALL)
+	if err == nil && len(routeList) > 0 {
+		var routeTable = RoutingTable{Id: routeList[0].Table, Src: routeList[0].Src}
+		var routes []Route
+		for _, route := range routeList {
+			var r = Route{}
+			if route.Dst != nil {
+				r.Prefix = (*route.Dst).IP
```
safchain: We need to have the mask for the prefix, route.Dst.Mask ?

topology/probes/netlink.go
```diff
@@ -437,6 +460,35 @@ func (u *NetNsNetLinkProbe) addLinkToTopology(link netlink.Link) {
...
+				for _, nexthop := range route.MultiPath {
+					var nhop = NextHop{Ip: nexthop.Gw, Priority: route.Priority}
+					nh = append(nh, nhop)
+				}
+			} else {
+				var nhop = NextHop{Ip: route.Gw, Priority: route.Priority}
+				nh = append(nh, nhop)
+			}
+			r.Nexthops = nh
+			routes = append(routes, r)
```
safchain: I don't know why but multiple nexthops are reported in different routes having the same prefix

topology/probes/netlink.go
```diff
@@ -79,6 +79,25 @@ type NetLinkProbe struct {
 	wg      sync.WaitGroup
 }
 
+// RouteTable describes a list of Routes
+type RoutingTable struct {
+	Id     int
```
safchain: I tried to use multiple tables but it seems that only the default table is reported

tests/topology_test.go
```diff
@@ -885,3 +885,55 @@ func TestUserMetadata(t *testing.T) {
...
+					helper.Cmd{Cmd: "ip netns exec vm1 ip route add 124.65.92.0/24 dev eth0", Check: true},
+					helper.Cmd{Cmd: "sleep 5", Check: false},
+				)
+
+				node, err = c.gh.GetNode(prefix + ".V().Has('IPV4', '124.65.91.42/24')")
+				routingTable = node.Metadata()["RoutingTable"].(map[string]interface{})
+				newNoOfRoutes := len(routingTable["Routes"].([]interface{}))
+
+				helper.ExecCmds(t,
+					helper.Cmd{Cmd: "ip netns exec vm1 ip route del 124.65.92.0/24 dev eth0", Check: true},
```
safchain: for the CI error it is normal as in the replay mode the vm1 netns doesn't exist anymore. I will think about a solution to fix this design issue

topology/probes/netlink.go
masco-enovance: done

topology/probes/netlink.go
masco-enovance: I thought there will be one table, so I wrote a code according to that.
I feel if you review the "getRoutingTable" method we can verify whether the implementation is correct or not.
In my system, I am getting routes for two table id's but from netlink i am getting only one id for all the routes. PSB the command line output
-bash-4.2$ sudo ip netns exec vm1 ip route show table 255
broadcast 124.65.91.0 dev eth0  proto kernel  scope link  src 124.65.91.42 
local 124.65.91.42 dev eth0  proto kernel  scope host  src 124.65.91.42 
broadcast 124.65.91.255 dev eth0  proto kernel  scope link  src 124.65.91.42 
-bash-4.2$ sudo ip netns exec vm1 ip route show table 254
124.65.91.0/24 dev eth0  proto kernel  scope link  src 124.65.91.42 

from netlink i am getting only 254

Please correct me if i understood and conveyed anything wrong.

topology/probes/netlink.go
masco-enovance: if i am not getting the multipath/Nexthop i am providing the route.Gw value similarly how the netlink string() method provides. this may be the reason. if it is not required i can remove this part but in my system i am always getting multipath array is empty.

safchain:
Hi @masco-enovance can you rebase your patch please

safchain:
@masco-enovance you have to bump to netlink library to have the your patch working properly

topology/probes/netlink.go
```diff
@@ -621,6 +703,29 @@ func (u *NetNsNetLinkProbe) initialize() {
 	}
 }
 
+func (u *NetNsNetLinkProbe) getRoute(m []byte) ([]RoutingTable, error, int) {
```
lebauce: s/getRoute/getRoutingTable

topology/probes/netlink.go
```diff
@@ -483,6 +506,48 @@ func (u *NetNsNetLinkProbe) addLinkToTopology(link netlink.Link) {
...
+	routeList, err := u.handle.RouteListFiltered(netlink.FAMILY_ALL, routeFilter, netlink.RT_FILTER_OIF|netlink.RT_FILTER_TABLE)
+	if err == nil && len(routeList) > 0 {
+		for _, route := range routeList {
+			var routeTable RoutingTable
+			if rt, ok := routeTableList[route.Table]; ok {
+				routeTable = rt
+			} else {
+				routeTable = RoutingTable{Id: route.Table, Src: route.Src}
+			}
+			var r = Route{}
```
lebauce: nit: var r Route

PR #381, opened 2017-08-11
ui: bugfix - Unregister buttons close callback

pierrecregut:
The callback registered on the document must be removed
as soon as it fires otherwise there is a long list of
callbacks that yields to a small memory leak and a real performance
problem.

### DIFF
--- a/statics/js/components/buttons.js
+++ b/statics/js/components/buttons.js
@@ -100,13 +100,15 @@ Vue.component('button-dropdown', {
     var self = this;
     // close the popup if we click elsewhere
     // from the target search if any parent has the id
-    $(document).on('click', function(event) {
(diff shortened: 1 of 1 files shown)


safchain:
ok to test

safchain:
run cdd-overview

statics/js/components/buttons.js
```diff
@@ -100,13 +100,15 @@ Vue.component('button-dropdown', {
     var self = this;
     // close the popup if we click elsewhere
     // from the target search if any parent has the id
-    $(document).on('click', function(event) {
+    var close_callback = function(event) {
```
eonpatapon: It should be writen in camelCase

PR #382, opened 2017-08-13
flow: SYN/FIN detection

vperelmax:
Added fixes after first round of reviews

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -104,6 +104,7 @@ agent:
     # stats_update: 1
   metadata:
     info: This is compute node
+  # capture_syn: true
 
 sflow:
   # Default listening address is 127.0.0.1
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -37,6 +37,7 @@ import (
 	"github.com/google/gopacket/layers"
 
 	"github.com/skydive-project/skydive/common"
+	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/logging"
 )
 
@@ -340,7 +341,6 @@ func (f *Flow) LinkType() (layers.LinkType, error) {
 	return 0, errors.New("LinkType unknown")
 }
 
-// Init a flow based on flow key and gopacket
 func (f *Flow) Init(key string, now int64, packet *gopacket.Packet, length int64, nodeTID string, parentUUID string, L2ID int64, L3ID int64) {
 	f.Start = now
 	f.Last = now
@@ -366,9 +366,11 @@ func (f *Flow) Init(key string, now int64, packet *gopacket.Packet, length int64
 // Update a flow metrics
 func (f *Flow) Update(now int64, packet *gopacket.Packet, length int64) {
 	f.Last = now
-
 	if updated := f.updateMetricsWithLinkLayer(packet, length); !updated {
 		f.updateMetricsWithNetworkLayer(packet)
+	}
+	if f.TCPFlowMetric != nil {
+		f.updateTCPMetrics(packet)
 	}
 }
 
@@ -499,6 +501,85 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
 	return errors.New("Unable to decode the IP layer")
 }
 
+func (f *Flow) updateTCPMetrics(packet *gopacket.Packet) error {
+	// capture content of SYN packets
+	//bypass if not TCP
+	if f.Network == nil || f.Transport == nil || f.Transport.Protocol != FlowProtocol_TCPPORT {
+		return nil
+	}
+	var metadata *gopacket.PacketMetadata
+	if metadata = (*packet).Metadata(); metadata == nil {
+		return nil
+	}
+
+	tcpLayer := (*packet).Layer(layers.LayerTypeTCP)
+	tcpPacket, ok := tcpLayer.(*layers.TCP)
+	if !ok {
+		logging.GetLogger().Notice("Capture SYN unable to decode TCP layer. ignoring")
+		return nil
+	}
+
+	// we capture SYN, FIN & RST
+	if !(tcpPacket.SYN || tcpPacket.FIN || tcpPacket.RST) {
+		return nil
+	}
+	var srcIP string
+	var timeToLive uint32
+	switch f.Network.Protocol {
+	case FlowProtocol_IPV4:
+		ipv4Layer := (*packet).Layer(layers.LayerTypeIPv4)
+		ipv4Packet, ok := ipv4Layer.(*layers.IPv4)
+		if !ok {
+			return errors.New("Unable to decode IPv4 Layer")
+		}
+		srcIP = ipv4Packet.SrcIP.String()
+		timeToLive = uint32(ipv4Packet.TTL)
+	case FlowProtocol_IPV6:
+		ipv6Layer := (*packet).Layer(layers.LayerTypeIPv6)
+		ipv6Packet, ok := ipv6Layer.(*layers.IPv6)
+		if !ok {
+			return errors.New("Unable to decode IPv4 Layer")
+		}
+		srcIP = ipv6Packet.SrcIP.String()
+		timeToLive = uint32(ipv6Packet.HopLimit)
+	default:
+		logging.GetLogger().Notice("Capture SYN unknown IP version. ignoring")
+		return nil
+
+	}
+
+	captureTime := common.UnixMillis(metadata.CaptureInfo.Timestamp)
+
+	switch {
+	case tcpPacket.SYN:
+		if f.Network.A == srcIP {
+			if f.TCPFlowMetric.ABSynStart == 0 {
+				f.TCPFlowMetric.ABSynStart = captureTime
+				f.TCPFlowMetric.ABSynTTL = timeToLive
+			}
+		} else {
+			if f.TCPFlowMetric.BASynStart == 0 {
+				f.TCPFlowMetric.BASynStart = captureTime
+				f.TCPFlowMetric.BASynTTL = timeToLive
+			}
+		}
+	case tcpPacket.FIN:
+		if f.Network.A == srcIP {
+			f.TCPFlowMetric.ABFinStart = captureTime
+		} else {
+			f.TCPFlowMetric.BAFinStart = captureTime
+		}
+	case tcpPacket.RST:
+		if f.Network.A == srcIP {
+			f.TCPFlowMetric.ABRstStart = captureTime
+		} else {
+			f.TCPFlowMetric.BARstStart = captureTime
+		}
+	}
+
+	return nil
+}
+
 func (f *Flow) newTransportLayer(packet *gopacket.Packet) error {
 	var transportLayer gopacket.Layer
 	var ok bool
@@ -528,6 +609,10 @@ func (f *Flow) newTransportLayer(packet *gopacket.Packet) error {
 		transportPacket, _ := transportLayer.(*layers.TCP)
 		f.Transport.A = strconv.Itoa(int(transportPacket.SrcPort))
 		f.Transport.B = strconv.Itoa(int(transportPacket.DstPort))
+		if config.GetConfig().GetBool("agent.capture_syn") {
+			f.TCPFlowMetric = &TCPMetric{}
+			return f.updateTCPMetrics(packet)
+		}
 	case FlowProtocol_UDPPORT:
 		transportPacket, _ := transportLayer.(*layers.UDP)
 		f.Transport.A = strconv.Itoa(int(transportPacket.SrcPort))
@@ -544,7 +629,6 @@ func (f *Flow) newTransportLayer(packet *gopacket.Packet) error {
 // case of encapsulation like GRE, VXLAN, etc.
 func PacketsFromGoPacket(packet *gopacket.Packet, outerLength int64, t int64, bpf *BPF) *Packets {
 	flowPackets := &Packets{Timestamp: t}
-
 	if (*packet).Layer(gopacket.LayerTypeDecodeFailure) != nil {
 		logging.GetLogger().Errorf("Decoding failure on layerpath %s", layerPathFromGoPacket(packet))
 		logging.GetLogger().Debug((*packet).Dump())
@@ -572,7 +656,6 @@ func PacketsFromGoPacket(packet *gopacket.Packet, outerLength int64, t int64, bp
 
 	// length of the encapsulation header + the inner packet
 	topLayerLength := outerLength
-
 	var start int
 	var innerLength int
 	for i, layer := range packetLayers {
@@ -607,6 +690,9 @@ func PacketsFromGoPacket(packet *gopacket.Packet, outerLength int64, t int64, bp
 
 	if len(flowPackets.Packets) > 0 {
 		p := gopacket.NewPacket(packetData[start:], topLayer.LayerType(), gopacket.NoCopy)
+		if metadata := (*packet).Metadata(); metadata != nil {
+			p.Metadata().CaptureInfo = metadata.CaptureInfo
+		}
 		flowPackets.Packets = append(flowPackets.Packets, Packet{gopacket: &p, length: 0})
 	} else {
 		flowPackets.Packets = append(flowPackets.Packets, Packet{gopacket: packet, length: outerLength})
@@ -697,6 +783,34 @@ func (i *ICMPLayer) GetFieldInt64(field string) (int64, error) {
 	}
 }
 
+// GetFieldInt64 returns the value of a ICMP field
+func (i *TCPMetric) GetFieldInt64(field string) (int64, error) {
+	if i == nil {
+		return 0, common.ErrFieldNotFound
+	}
+
+	switch field {
+	case "ABSynTTL":
+		return int64(i.ABSynTTL), nil
+	case "BASynTTL":
+		return int64(i.BASynTTL), nil
+	case "ABSynStart":
+		return i.ABSynStart, nil
+	case "BASynStart":
+		return i.BASynStart, nil
+	case "ABFinStart":
+		return i.ABFinStart, nil
+	case "BAFinStart":
+		return i.BAFinStart, nil
+	case "ABRstStart":
+		return i.BARstStart, nil
+	case "BARstStart":
+		return i.BARstStart, nil
+	default:
+		return 0, common.ErrFieldNotFound
+	}
+}
+
 // GetFieldString returns the value of a Flow field
 func (f *Flow) GetFieldString(field string) (string, error) {
 	fields := strings.Split(field, ".")
@@ -770,6 +884,8 @@ func (f *Flow) GetFieldInt64(field string) (_ int64, err error) {
 		return f.Metric.GetFieldInt64(fields[1])
 	case "LastUpdateMetric":
 		return f.LastUpdateMetric.GetFieldInt64(fields[1])
+	case "TCPFlowMetric":
+		return f.TCPFlowMetric.GetFieldInt64(fields[1])
 	case "Link":
 		return f.Link.GetFieldInt64(fields[1])
 	case "Network":
--- a/flow/flow.proto
+++ b/flow/flow.proto
@@ -80,6 +80,17 @@ message RawPacket {
   bytes Data = 3;
 }
 
+message TCPMetric {
+  int64 ABSynStart = 1;
+  int64 BASynStart = 2;
+  uint32 ABSynTTL = 3;
+  uint32 BASynTTL = 4;
+  int64 ABFinStart = 5;
+  int64 BAFinStart = 6;
+  int64 ABRstStart = 7;
+  int64 BARstStart = 8;
+}
+
 message Flow {
 /* Flow Universally Unique IDentifier
    flow.UUID is unique in the universe, as it should be used as a key of an
@@ -105,6 +116,9 @@ message Flow {
 	FlowMetric LastUpdateMetric = 31;
 /* Total amount of data for the whole flow duration */
 	FlowMetric Metric = 32;
+
+/* Flow metric specific to the TCP Protocol only and optional */
+  TCPMetric TCPFlowMetric = 38;
 
   int64 Start = 10;
   int64 Last = 11;
--- a/flow/flow_test.go
+++ b/flow/flow_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/google/gopacket"
 	"github.com/google/gopacket/layers"
 	"github.com/google/gopacket/pcap"
+	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/filters"
 )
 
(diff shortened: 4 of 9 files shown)


masco-enovance:
ok to test

flow/flow.go
```diff
@@ -240,7 +243,7 @@ func networkID(p *gopacket.Packet) int64 {
 	return id
 }
 
-// NewFlow creates a new empty flow
+// NewFlow returns extended with memory-only attributes Flow instance
```
nplanel: "// NewFlow creates a new empty flow" is enough

etc/skydive.yml.default
```diff
@@ -104,6 +104,8 @@ agent:
     # stats_update: 1
   metadata:
     info: This is compute node
+  tcp_len_by_seq: true
```
nplanel: I understand that your patch for tcp_len_by_seq, counting flow bytes based on the tcp len and avoid counting the retransmission. Why leaving the flow bytes as it's done today and add a TCPRetransmissionLen in TCPFlowMetric
as we can store these datas and see number of retransmitted packets and bytes in a separated TCP only statistics

I would rather prefer to have TCP stack counters like, overlap, retransmission that we could add in TCP Metric, and enable them by default. Basically add counter like wireshark TCP analysis.
Another interesting subject would be IP fragmentation

What do you think ?

flow/flow.go
```diff
@@ -499,6 +545,167 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
 	return errors.New("Unable to decode the IP layer")
 }
 
+func (f *Flow) updateMetricsSynCapture(packet *gopacket.Packet) error {
```
nplanel: updateTCPMetrics() name will fit nicer

flow/flow.go
```diff
@@ -499,6 +545,167 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
...
+		f.Metric.ABBytes += int64(length)
+		f.state.aBExpectedSeq = newExpectedSeq
+	} else {
+		f.Metric.BAPackets += int64(1)
+		f.Metric.BABytes += int64(length)
+		f.state.bAExpectedSeq = newExpectedSeq
+	}
+
+	if transportPacket.FIN {
+		logging.GetLogger().Notice("FIN", f.state.aBExpectedSeq, f.state.bAExpectedSeq, f.Metric.ABBytes, f.Metric.BABytes)
```
nplanel: Notice is too verbose, you will see 1 log per tcp flow
As you don't dump the flowid I think you can remove this line

flow/flow.go
```diff
@@ -528,6 +735,13 @@ func (f *Flow) newTransportLayer(packet *gopacket.Packet) error {
 		transportPacket, _ := transportLayer.(*layers.TCP)
 		f.Transport.A = strconv.Itoa(int(transportPacket.SrcPort))
 		f.Transport.B = strconv.Itoa(int(transportPacket.DstPort))
+		if config.GetConfig().GetBool("agent.capture_syn") {
+			f.TCPFlowMetric = &TCPMetric{
+				ABSynStart: int64(0),
```
nplanel: don't need to initialize at 0, it's already the default in go

flow/flow.go
```diff
@@ -697,6 +911,50 @@ func (i *ICMPLayer) GetFieldInt64(field string) (int64, error) {
...
 }
 
+// GetStringField returns the value of a ICMP field
+func (i *TCPMetric) GetStringField(field string) (string, error) {
+	if i == nil {
+		return "", common.ErrFieldNotFound
+	}
+
+	switch field {
+	case "ABSynData":
```
nplanel: This is already addressed by @safchain patch #299

flow/flow.go
```diff
@@ -499,6 +545,167 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
...
+			}
+			expectedSeq = f.state.bAExpectedSeq
+		}
+	} else {
+		logging.GetLogger().Error("Error retrieving IP layer")
+	}
+
+	length := uint32Distance(transportPacket.Seq, expectedSeq)
+
+	if transportPacket.Seq < TCPMaxSegmentSize {
```
nplanel: I would prefer to see a logic that track the TCP segment, that will address the TCP partial segment retransmission, etc ...
Add this sliding window logic tracking in flowState
With these information, we could count number of restransmitted and overlaped packets and bytes easily and correctly
In the current version you only retransmission when the restransmit packet if old by TCPMaxSegmentSize, that it different than TCP window size.

flow/flow.go
```diff
@@ -499,6 +545,167 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
...
+		logging.GetLogger().Notice("Overlap", transportPacket.Seq, expectedSeq)
+	}
+
+	if length > 0x80010000 {
```
nplanel: why 0x80010000 ? -TCPMaxSegmentSize ?

flow/flow.go
vperelmax: OK

flow/flow.go
vperelmax: OK

flow/flow.go
vperelmax: It looks like additional feature... 
In this feature we don't detect retransmittings and with BPF it is even impossible to detect them this way, we just calculate length of data that passes in the session in a more exact way using sequence numbers, rather than just calculating passed bytes, because not always we look at all the packets that pass (only SYN/FIN/RST, if BPF is used). 
Its my fault that I have added retransmit to the title, I have removed it.
I think that thing that you propose to calculate is more complex to do and I propose to open it as an issue (plan for the next feature).

flow/flow.go
vperelmax: OK

flow/flow.go
vperelmax: OK

etc/skydive.yml.default
vperelmax: My patch doesn't deal so much with retransmissions. It refers more to the case when for example we capture only partial traffic (such as SYN/FIN) and then using sequence numbers we calculate traffic length. 
Thing that you propose looks like good thing to have, but I see it as rather big and independent feature. So, I propose opening it as a new open issue and starting discussing it via our meetings.

flow/flow.go
vperelmax: OK, u r right, I will remove it

flow/flow.go
vperelmax: It wont pass compilation in go

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -81,6 +81,54 @@ const flowMapping = `
 					"type": "date", "format": "epoch_millis"
 				}
 			}
+		},
+		{
```
nplanel: just declare these with simple regex like match by modify the "start"
like this : 
		{
			"start": {
				"match": "*Start",
				"mapping": {
					"type": "date", "format": "epoch_millis"
				}
			}
		},

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -122,6 +170,14 @@ const metricMapping = `
 					"type": "date", "format": "epoch_millis"
 				}
 			}
+		},
```
nplanel: I think you forgot to remove this section

flow/storage/orientdb/orientdb.go
```diff
@@ -67,28 +67,74 @@ func flowMetricToDocument(flow *flow.Flow, metric *flow.FlowMetric) orient.Docum
 	}
 }
 
+func flowTCPMetricToDocument(flow *flow.Flow, tcp_metric *flow.TCPMetric) orient.Document {
+
+	if tcp_metric != nil {
+		return orient.Document{
+			"@class":     "TCPMetric",
+			"@type":      "d",
+			"Start":      flow.LastUpdateStart,
```
nplanel: TCPMetrics don't have Start and Last

flow/storage/orientdb/orientdb.go
```diff
@@ -67,28 +67,74 @@ func flowMetricToDocument(flow *flow.Flow, metric *flow.FlowMetric) orient.Docum
...
+	if tcpMetric != nil {
+		flowDoc = orient.Document{
+			"@class":             "Flow",
+			"UUID":               flow.UUID,
+			"LayersPath":         flow.LayersPath,
+			"Application":        flow.Application,
+			"LastUpdateMetric":   lastMetricDoc,
+			"Metric":             metricDoc,
+			"TCPFlowMetric":      tcpMetric,
+			"Start":              flow.Start,
```
nplanel: same here "TCPMetrics don't have Start and Last"

flow/storage/orientdb/orientdb.go
```diff
@@ -391,6 +437,30 @@ func New() (*OrientDBStorage, error) {
 		}
 	}
 
+	if _, err := client.GetDocumentClass("TCPMetric"); err != nil {
+		class := orient.ClassDefinition{
+			Name: "TCPMetric",
+			Properties: []orient.Property{
+				{Name: "Start", Type: "LONG", Mandatory: true, NotNull: true},
```
nplanel: same here "TCPMetrics don't have Start and Last"

flow/flow.go
```diff
@@ -367,9 +407,15 @@ func (f *Flow) Init(key string, now int64, packet *gopacket.Packet, length int64
 func (f *Flow) Update(now int64, packet *gopacket.Packet, length int64) {
 	f.Last = now
 
-	if updated := f.updateMetricsWithLinkLayer(packet, length); !updated {
+	if f.state.lenBySeq {
+		f.updateMetricsBySeq(packet)
+	} else if updated := f.updateMetricsWithLinkLayer(packet, length); !updated {
+		logging.GetLogger().Error("Update metrics with network\n")
```
nplanel: please remove this log

etc/skydive.yml.default
nplanel: Ok I understand your point
But if you don't take in account overlapping, retransmission, reordering packets, you flow tcp size metric would not be correct.

So let's do this patch step by step, and just focus on the syn/fin/rst flag and ttl only, this part sound good for me except few exception.
Could you remove the "unrelated" parts to tcp len calculation, I will prefer to see that in a separate patch

It's a pity that we have a big timezone difference, I will try catch you on IRC tonight (morning for you)

flow/flow.go
nplanel: yes, the idea was have a signed distance

flow/flow_test.go
```diff
@@ -79,6 +79,12 @@ func TestFlowJSON(t *testing.T) {
 		},
```
vperelmax: Done

flow/storage/orientdb/orientdb.go
vperelmax: OK, uploading change

flow/storage/orientdb/orientdb.go
vperelmax: OK

flow/flow.go
```diff
@@ -770,6 +890,10 @@ func (f *Flow) GetFieldInt64(field string) (_ int64, err error) {
 		return f.Metric.GetFieldInt64(fields[1])
 	case "LastUpdateMetric":
 		return f.LastUpdateMetric.GetFieldInt64(fields[1])
+	case "TCPFlowMetric":
+		if f.TCPFlowMetric != nil {
+			return f.TCPFlowMetric.GetFieldInt64(fields[1])
```
nplanel: you can call directly the return f.TCPFlowMetric.GetFieldInt64( ....  as you do the test already line 794

flow/flow.proto
```diff
@@ -80,6 +80,17 @@ message RawPacket {
   bytes Data = 3;
 }
 
+message TCPMetric {
+	int64 ABSynStart = 7;
```
nplanel: why not start protobuf ID at 1

flow/flow.proto
```diff
@@ -106,6 +117,9 @@ message Flow {
 /* Total amount of data for the whole flow duration */
 	FlowMetric Metric = 32;
 
+/* Floe metric specific to the TCP Protocol only and optional */
+    TCPMetric TCPFlowMetric = 38;
```
nplanel: use ID 33 as it's free

flow/flow_test.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/google/gopacket"
 	"github.com/google/gopacket/layers"
 	"github.com/google/gopacket/pcap"
+	"github.com/skydive-project/skydive.working.at.6.10/config"
```
nplanel: s|"github.com/skydive-project/skydive.working.at.6.10/config"|"github.com/skydive-project/skydive/config"|

flow/storage/orientdb/orientdb.go
```diff
@@ -67,28 +67,72 @@ func flowMetricToDocument(flow *flow.Flow, metric *flow.FlowMetric) orient.Docum
...
-		"TrackingID":         flow.TrackingID,
-		"L3TrackingID":       flow.L3TrackingID,
-		"ParentUUID":         flow.ParentUUID,
-		"NodeTID":            flow.NodeTID,
-		"ANodeTID":           flow.ANodeTID,
-		"BNodeTID":           flow.BNodeTID,
-		"RawPacketsCaptured": flow.RawPacketsCaptured,
+	tcpMetric := flowTCPMetricToDocument(flow, flow.TCPFlowMetric)
+	var flowDoc orient.Document
+	if tcpMetric != nil {
```
nplanel: don't do a big if here, just do after flowDoc declaration
if tcpMetric != nil {
   flowDoc["TCPFlowMetric"] = tcpMetric
}

flow/storage/orientdb/orientdb.go
```diff
@@ -391,6 +435,28 @@ func New() (*OrientDBStorage, error) {
 		}
 	}
 
+	if _, err := client.GetDocumentClass("TCPMetric"); err != nil {
+		class := orient.ClassDefinition{
+			Name: "TCPMetric",
+			Properties: []orient.Property{
+				{Name: "ABSynStart", Type: "LONG", Mandatory: false, NotNull: false},
```
nplanel: put the NotNull to true for all TCPMetric

flow/flow_test.go
```diff
@@ -273,9 +285,8 @@ func validateAllParentChains(t *testing.T, table *Table) {
 
 func flowsFromPCAP(t *testing.T, filename string, linkType layers.LinkType, bpf *BPF) []*Flow {
 	table := NewTable(nil, nil, NewEnhancerPipeline(), "", TableOpts{})
-
+	table.SetSynCapture(true)
```
nplanel: as the flag is global to the agent, and will be set only by the user, set the config flag directly here for the purpose of the test, so place this code in your unit test (TestFlowSimpleSynFin)

config.GetConfig().Set("agent.capture_syn", true)
defer config.GetConfig().Set("agent.capture_syn", false)

flow/table.go
```diff
@@ -349,6 +349,11 @@ func (ft *Table) flowPacketToFlow(packet *Packet, parentUUID string, t int64, L2
 	return flow
 }
 
+// Enable/Disable capturing SYN packets
+func (ft *Table) SetSynCapture(enableSynCapture bool) {
```
nplanel: remove this class method, see comment above

flow/flow_test.go
```diff
@@ -190,21 +202,22 @@ func fillTableFromPCAP(t *testing.T, table *Table, filename string, linkType lay
...
+			p.Metadata().CaptureInfo = ci
 			if p.ErrorLayer() != nil {
 				t.Fatalf("Failed to decode packet with layer path '%s': %s\n", layerPathFromGoPacket(&p), p.ErrorLayer().Error())
 			}
 			pcapPacketNB++
 			if strings.Contains(layerPathFromGoPacket(&p), "DecodeFailure") {
 				t.Fatalf("GoPacket decode this pcap packet %d as DecodeFailure :\n%s", pcapPacketNB, p.Dump())
 			}
-
+			config.GetConfig().Set("agent.capture_syn", true)
```
nplanel: why forcing the flag here, old debuging trace I assume ...

flow/flow.go
```diff
@@ -499,6 +501,91 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
 	return errors.New("Unable to decode the IP layer")
 }
 
+func (f *Flow) updateTCPMetrics(packet *gopacket.Packet) error {
+	// capture content of SYN packets
+	//bypass if not TCP
+	if f.Transport == nil || f.Transport.Protocol != FlowProtocol_TCPPORT {
```
vperelmax: OK

flow/flow.go
```diff
@@ -499,6 +501,91 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
...
+		}
+		srcIP = ipv6Packet.SrcIP.String()
+		timeToLive = uint32(ipv6Packet.HopLimit)
+	default:
+		logging.GetLogger().Notice("Capture SYN unknown IP version. ignoring")
+		return nil
+
+	}
+
+	var captureTime int64
```
vperelmax: OK

flow/flow.go
```diff
@@ -607,6 +698,7 @@ func PacketsFromGoPacket(packet *gopacket.Packet, outerLength int64, t int64, bp
 
 	if len(flowPackets.Packets) > 0 {
 		p := gopacket.NewPacket(packetData[start:], topLayer.LayerType(), gopacket.NoCopy)
+		p.Metadata().CaptureInfo = packet_capture_info
```
vperelmax: I do it here because, p is a packet copy that is created and scoped only here and NewPacket is not our function (we cant change it) and it doesn't copy the metadata that we need (the timestemps), so we loose here the data. But you are right, there is some mess in the code, I will fix it and also add check for metadata.

flow/flow.go
vperelmax: Done

flow/flow.proto
vperelmax: Done

flow/flow.proto
vperelmax: 33 belongs to NodeTID

flow/flow_test.go
vperelmax: dont need it at all, I remove it

flow/flow_test.go
vperelmax: right :)

flow/flow_test.go
vperelmax: I tried, it didnt work. Somehow in the test it sees a different config variable... I did exactly what you type and printed the valuse right before calling table after and inside the code of table, and I see that they work on different config instances... So this was my hack to overcome the problem.
This function is not totally not needed, I think in the future we might want to add some UI button that could change SYN/FIN capturing flag, then it will be useful not only in the tests

flow/table.go
vperelmax: Please, see my comment above.

flow/storage/orientdb/orientdb.go
vperelmax: OK

flow/storage/orientdb/orientdb.go
vperelmax: OK

vperelmax:
I see that in my environments the tests pass, I think there is some problem with the environment, can anyone help with it?thanks,
-L.

flow/table.go
vperelmax: OK, I removed it.
The bug was that I used two different folders of config, now, everything works well.
Thanks!

flow/flow_test.go
vperelmax: OK

vperelmax:
Hi Nicolas,
I see that fmt test falls when running "make vet", it is falling on code that wasn't changed in the patch... please, advise what I should do about it.
Thanks,
-L.

flow/flow.proto
```diff
@@ -80,6 +80,17 @@ message RawPacket {
   bytes Data = 3;
 }
 
+message TCPMetric {
+	int64 ABSynStart = 1;
```
nplanel: ident this block like others, prefix with 2 spaces

flow/flow.proto
```diff
@@ -106,6 +117,9 @@ message Flow {
 /* Total amount of data for the whole flow duration */
 	FlowMetric Metric = 32;
 
+/* Floe metric specific to the TCP Protocol only and optional */
+  TCPMetric TCPFlowMetric = 38;
```
nplanel: ID 33

flow/flow_test.go
```diff
@@ -759,3 +776,38 @@ func TestGREMPLS(t *testing.T) {
 
 	validatePCAP(t, "pcaptraces/gre-mpls-icmpv4.pcap", layers.LinkTypeEthernet, nil, expected)
 }
+
+func TestFlowSimpleSynFin(t *testing.T) {
+	flows := flowsFromPCAP(t, "pcaptraces/simple-tcpv4.pcap", layers.LinkTypeEthernet, nil)
+	// In test pcap SYNs happen at 2017-03-21 10:58:23.768977 +0200 IST
+	synTimestemp := int64(1490086703768)
```
nplanel: s/Timestemp/Timestamp/g

flow/metrics.go
```diff
@@ -25,12 +25,27 @@ package flow
...
+		ABBytes:   fm.ABBytes,
+		BAPackets: fm.BAPackets,
+		BABytes:   fm.BABytes,
+	}
+}
+
+// Copy extended metric
+func (fm *TCPMetric) Copy() *TCPMetric {
+	return &TCPMetric{
+
```
nplanel: remove this line

flow/storage/orientdb/orientdb.go
```diff
@@ -67,11 +67,32 @@ func flowMetricToDocument(flow *flow.Flow, metric *flow.FlowMetric) orient.Docum
 	}
 }
 
+func flowTCPMetricToDocument(flow *flow.Flow, tcp_metric *flow.TCPMetric) orient.Document {
+
```
nplanel: remove this line

flow/table.go
```diff
@@ -208,7 +208,6 @@ func (ft *Table) updateMetric(f *Flow, start, last int64) {
 	} else {
 		f.LastUpdateStart = f.Start
 	}
-
```
nplanel: don't remove this line, cosmetic

flow/flow.proto
vperelmax: NodeTID is 33, I think it is impossible in the same struct to have two elements with the same ID

flow/flow_test.go
vperelmax: Done

flow/flow.proto
vperelmax: Done

flow/metrics.go
vperelmax: Done

flow/storage/orientdb/orientdb.go
vperelmax: Done

PR #384, opened 2017-08-14
performance counters : add performance and troubleshooting counters

nplanel:
Available via http on agent and analyzer instance , ex : http://localhost:8082/perfcounters

### DIFF
--- a/common/perfcounters.go
+++ b/common/perfcounters.go
@@ -1,0 +1,140 @@
+/*
+ * Copyright (C) 2017 Red Hat, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package common
+
+import (
+	"expvar"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/abbot/go-http-auth"
+	"github.com/paulbellamy/ratecounter"
+)
+
+var perfMap map[string]*expvar.Map
+
+// PerfCounterIntRate describe a performance counter
+type PerfCounterIntRate struct {
+	count      expvar.Int
+	in         expvar.Int
+	rate       expvar.Int
+	rateMinute *ratecounter.RateCounter
+	avg        expvar.Float
+	avgMinute  *ratecounter.AvgRateCounter
+	startTime  time.Time
+}
+
+// from expvar go 1.7
+func expvarHandler(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	fmt.Fprintf(w, "{\n")
+	first := true
+	expvar.Do(func(kv expvar.KeyValue) {
+		if !first {
+			fmt.Fprintf(w, ",\n")
+		}
+		first = false
+		fmt.Fprintf(w, "%q: %s", kv.Key, kv.Value)
+	})
+	fmt.Fprintf(w, "\n}\n")
+}
+
+func expvarHandlerAuth(w http.ResponseWriter, r *auth.AuthenticatedRequest) {
+	expvarHandler(w, &r.Request)
+}
+
+// PerfcountersHandler return an http handler to show the values
+func PerfcountersHandler() http.Handler {
+	return http.HandlerFunc(expvarHandler)
+}
+
+// PerfcountersHandlerAuth retur, an http authenticated handler
+func PerfcountersHandlerAuth() auth.AuthenticatedHandlerFunc {
+	return expvarHandlerAuth
+}
+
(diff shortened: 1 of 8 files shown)


nplanel:
let's say it's for developers when user have issue, but I can add a special flag in the configuration file
I write these in mind to track our futur test on scalelab

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -154,13 +154,16 @@ const rawPacketMapping = `
 // ElasticSearchStorage describes an ElasticSearch flow backend
 type ElasticSearchStorage struct {
 	client *esclient.ElasticSearchClient
+	perf   map[string]*common.PerfCounterIntRate
```
safchain: I would prefer to have a global maybe in the perfcounter package itself so that less intrusion in the current code

common/perfcounters.go
```diff
@@ -0,0 +1,130 @@
...
+// Epilog of the performance counter
+func (p *PerfCounterIntRate) Epilog(delta int64) {
+	p.avgMinute.Incr(time.Since(p.startTime).Nanoseconds())
+	p.avg.Set(p.avgMinute.Rate())
+	p.in.Add(-delta)
+}
+
+func newOrGetPerfMap(name string) *expvar.Map {
+	if m, ok := perfMap[name]; ok {
+		return m
```
safchain: shouldn't raise an error instead of returning the map ? if by mistake we use the same name ?

http/server.go
```diff
@@ -224,9 +234,17 @@ func NewServer(host string, serviceType common.ServiceType, addr string, port in
 		Auth:        auth,
 	}
 
+	router.PathPrefix("/statics").HandlerFunc(server.serveStatics)
 	router.HandleFunc("/login", server.serveLogin)
 	router.HandleFunc("/", server.serveIndex)
 
+	router.Handle("/debug/vars", common.PerfcountersHandler())
```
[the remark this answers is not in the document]
nplanel: yes I can do that, sideeffect : a perfcounter will be created per entry as well, as RegisterRoutes create it

flow/table.go
```diff
@@ -392,20 +401,30 @@ func (ft *Table) Run() {
 	for atomic.LoadInt64(&ft.state) == common.RunningState {
 		select {
 		case now := <-expireTicker.C:
+			ft.perf["Expire"].Prolog(1)
```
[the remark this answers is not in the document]
nplanel: it's more readable to have it all around the same place

flow/table.go
```diff
@@ -392,20 +401,30 @@ func (ft *Table) Run() {
...
 		case query, ok := <-ft.query:
 			if ok {
+				ft.perf["Query"].Prolog(1)
 				ft.reply <- ft.onQuery(query)
+				ft.perf["Query"].Epilog(1)
 			}
 		case now := <-nowTicker.C:
 			ft.tableClock = common.UnixMillis(now)
 		case packets := <-ft.PacketsChan:
+			ft.perf["Packets"].Prolog(int64(len(packets.Packets)))
```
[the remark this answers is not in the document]
nplanel: agree to be pluggable, 
I'm sure that len(slice) is cost less in go : https://blog.golang.org/go-slices-usage-and-internals

common/perfcounters.go
nplanel: good question, that's the default of expvar, but in our code (I don't have the example right now) we would like refer to the same map/counters (and it's legit) so that why I did that like this (newOrGet)

flow/table.go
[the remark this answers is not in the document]
nplanel: but I agreed that we can move it inside the function

flow/storage/elasticsearch/elasticsearch.go
nplanel: you perfer to have a kind of global tree ? (and remove the perf per instance ?)
rootPerf["storage.elasticsearch"]["StoreFlows"].Prolog(...

http/server.go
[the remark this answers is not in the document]
nplanel: done

safchain:
This project could help to bring optional counters : https://github.com/deferpanic/goweave

nplanel:
run tests

nplanel:

nplanel:

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -161,13 +161,16 @@ const rawPacketMapping = `
...
 }
 
 // StoreFlows push a set of flows in the database
 func (c *ElasticSearchStorage) StoreFlows(flows []*flow.Flow) error {
 	if !c.client.Started() {
 		return errors.New("ElasticSearchStorage is not yet started")
 	}
+	//@perf c.perf["StoreFlows"].Prolog(int64(len(flows)))
```
safchain: I would prefer to add one tag/comment before the function definition, your perl script would read this tag and would inject these two lines at the beginning of the function using the name of the function+the path. I think it is too intrusive to add things in the middle of the functions. if we need to provide more fine grained stuff to the users then we should split functions but that is another story.

nplanel: I think tagging the line give us flexibility that we want to inject the code that, specially in the case of inside loop (see flow/table.go)
In the case of StoreFlows for example, we didn't want to count StoreFlows call/timing when the client is not connected

nplanel:

flow/storage/elasticsearch/elasticsearch.go
safchain: Let's start with something simple and non intrusive, here it is really too intrusive. If we want to get fine grained metrics let us the profiler otherwise something simple will be fine and if at some point we need more flexibility we will improve it

nplanel: keep in mind this patch is to measure internal performances during runtime more like a monitoring tool ... profiling we let known already on calls and stacktrace of functions
Where do you feel this is intrusive ? in all case the code will not run on production/normal build/release

safchain:
Close for now. Let's re-open it if needed

PR #386, opened 2017-08-14
inject: add support for injecting payload with icmp packets (by length, to allow stress tests)

eranra:
…h, to allow stress tests)

### DIFF
--- a/statics/js/components/inject-form.js
+++ b/statics/js/components/inject-form.js
@@ -36,6 +36,10 @@ Vue.component('inject-form', {
           <label for="inject-count">ICMP Identifier</label>\
           <input id="inject-id" type="number" class="form-control input-sm" v-model="id" min="0" />\
         </div>\
+        <div class="form-group">\
+          <label for="payload-length">Payload length</label>\
+          <input id="payload-length" type="number" class="form-control input-sm" v-model="payloadlength" min="0" />\
+        </div>\
       </div>\
       <div v-if="type === \'tcp4\' || type === \'tcp6\'">\
         <div class="form-group">\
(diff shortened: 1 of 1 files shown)


masco-enovance:
ok to test

nplanel:
please use lowcase in patch subject, for log comments consistency

eranra:
fixed

nplanel:
LGTM
@masco-enovance could you test this patch ?

statics/js/components/inject-form.js
```diff
@@ -36,6 +36,10 @@ Vue.component('inject-form', {
           <label for="inject-count">ICMP Identifier</label>\
           <input id="inject-id" type="number" class="form-control input-sm" v-model="id" min="0" />\
         </div>\
+        <div class="form-group">\
+          <label for="payload-length">Payload length</label>\
```
masco-enovance: in this the user can't set the custom payload msg. i would like to have both payload length and payload with select option. so the user can select either one of that. what you think?

eranra: I Just want to add the option to increase the size of the packets at this time (and ICMP packets are sufficient for stress testing). 

Totally agree that in future enhancements we should allow "custom payload". My thoughts are that this can be injected by allowing a user to point into something like the output of the file that one gets when capturing packets. So the custom injection can be similar to a 'reverse' of the capture process. But this is much bigger, and not part of this commit. I would wait for Sylvain to finish the new format of the capture packets process before attempting to write this 'custom injection'. For now this commit is a stand alone thing that allows us to create stress in the system and check performance ... nice addition with small amount of code.

masco-enovance:
in future it needs to extend for tcp protocol also.

PR #387, opened 2017-08-14
typo: in comments, should be resource instead of ressource 

eranra:


masco-enovance:
ok to test

nplanel:
could you fix the other issues like with a simple grep I found : 

api/alert.go:// Name returns ressource name "alert"
api/handler.go:// WatcherCallback callback called by the ressources watcher
api/handler.go:// BasicStoppableWatcher basic implementation of a ressources watcher
api/packet_injector.go:// RegisterPacketInjectorAPI registers a new packet injector ressource in the API

eranra:
Fixed

PR #398, opened 2017-08-21
ui: expand all childs in a group

masco-enovance:
added a dedicated button for node to expand
and collapse all the child groups in the node group.

### DIFF
--- a/statics/css/skydive.css
+++ b/statics/css/skydive.css
@@ -190,7 +190,7 @@ li.capture-item {
   padding-bottom: 5px;
 }
 
-.capture-action {
+.capture-action, .node-action {
   float: right;
   cursor: pointer;
   font-size: 1.2em;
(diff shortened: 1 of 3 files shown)


masco-enovance:
run tests

masco-enovance:
run coverage

nplanel:
run skydive-coverage

masco-enovance:
run coverage

vperelmax:
Looks, fine to me.
Small comment: I think the icon and its placement near "metadatas" is not intuitive...
Maybe, we should add button with name "Collapse/Expand" similar to general button. also it is better for each thing to have as unique way as possible in the UI.

vperelmax:
By the way, the code is written very nice!
Another small note/suggestion: maybe we should add some way to give to user by maybe double click on "+" icon of the button to explicitly say how many levels we would like to open.
Just thinking aloud :)

masco-enovance:
Added a button in topology window too. it will appear only after selecting any one node.

safchain:
Maybe we should now use only images in the toolbar, it will take less place and will be more consistent but that another story! :)

PR #401, opened 2017-08-24
flow : add TCP and IP high level metric

nplanel:


flow/flow.proto
```diff
@@ -89,6 +89,11 @@ message TCPMetric {
   int64 BAFinStart = 6;
   int64 ABRstStart = 7;
   int64 BARstStart = 8;
+
+  int64 IPFragments = 9;
+  int64 TCPSegmentOutOfOrder = 10;
+  int64 TCPRestransmissionPackets = 11;
+  int64 TCPRestransmissionBytes = 12;
 }
```
vperelmax: Why do we add fields that currently are not calculated?
I mean Retransmitting, OutOfOrder and so on? I think it is better to add them only when we do implement this part

flow/flow.go
```diff
@@ -610,7 +617,7 @@ func (f *Flow) newTransportLayer(packet *gopacket.Packet) error {
 		f.Transport.A = strconv.Itoa(int(transportPacket.SrcPort))
 		f.Transport.B = strconv.Itoa(int(transportPacket.DstPort))
 		if config.GetConfig().GetBool("agent.capture_syn") {
-			f.TCPFlowMetric = &TCPMetric{}
+			f.TCPIPMetric = &TCPIPMetric{}
 			return f.updateTCPMetrics(packet)
```
vperelmax: I think here we should check if BPF filter is on before calculating fragments, because this code is valid onkly when we do collect all the packets.
What do you think?

flow/flow.go
```diff
@@ -496,6 +499,10 @@ func (f *Flow) updateMetricsWithNetworkLayer(packet *gopacket.Packet) error {
 			f.Metric.BAPackets++
 			f.Metric.BABytes += int64(ipv6Packet.Length)
 		}
+		ipv6Frag := (*packet).Layer(layers.LayerTypeIPv6Fragment)
+		if _, ok := ipv6Frag.(*layers.IPv6Fragment); ok {
+			f.TCPIPMetric.IPFragments++
+		}
```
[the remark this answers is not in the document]
nplanel: in case of ipv4 it's part of the protocol header (a flag)
in case of ipv6 it's an optional header on top of ipv6

flow/flow.go
safchain: checking BPF will not tell us that we are going to get or not all the packets, for ex: filtering only port 80 packets.

flow/flow.go
```diff
@@ -369,7 +369,7 @@ func (f *Flow) Update(now int64, packet *gopacket.Packet, length int64) {
 	if updated := f.updateMetricsWithLinkLayer(packet, length); !updated {
 		f.updateMetricsWithNetworkLayer(packet)
 	}
-	if f.TCPFlowMetric != nil {
+	if f.TCPIPMetric != nil {
```
safchain: TCPMetric ?

nplanel:
we need to integrate and use latest gopacket as it include IP defragment and TCP reassembly

nplanel:
https://github.com/google/gopacket/blob/master/examples/statsassembly/main.go

flow/flow.go
```diff
@@ -693,6 +743,49 @@ func PacketsFromGoPacket(packet *gopacket.Packet, outerLength int64, t int64, bp
 		return flowPackets
 	}
 
+	var ipFragStat IPFragmentStat
+	if IPv4Defragmentation {
+		ip4Layer := (*packet).Layer(layers.LayerTypeIPv4)
```
safchain: I'm not sure we should have this here. I would prefer to put this in network layer

nplanel:
run tests

nplanel:
run tests

nplanel:

nplanel:
run tests

nplanel:
run skydive-selenium-tests

nplanel:
has been merged in #904

PR #403, opened 2017-08-25
openflow: bugfix - race with ovs-ofctl and functional tests.

### LINKED ISSUE #370: ovs_of: functional tests are needed

pierrecregut:
Race between ovs-ofctl create and ovs-ofctl monitor when the bridge appears in ovsdb database

- added a functional test on openflow probe. It may exhibit the
problem
- added info logging (as for other probes) to make functional
  tests log usable.

closes #370

### DIFF
--- a/topology/probes/ovs_of.go
+++ b/topology/probes/ovs_of.go
@@ -292,29 +292,36 @@ func launchOnSwitch(cmd []string) (string, error) {
 	return "", err
 }
 
-// launchContinuousOnSwitch launches  a stream producing command on a given switch
+// launchContinuousOnSwitch launches  a stream producing command on a given switch. The command is resilient
+// and is relaunched until the context explicitely cancels it.
 func launchContinuousOnSwitch(ctx context.Context, cmd []string) (<-chan string, error) {
 	var cout = make(chan string, 10)
-	out, err := executor.ExecCommandPipe(ctx, cmd[0], cmd[1:]...)
-	if err != nil {
-		close(cout)
-		return cout, err
-	}
-	reader := bufio.NewReader(out)
-	var line string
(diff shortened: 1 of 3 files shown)


topology/probes/ovs_of.go
```diff
@@ -292,29 +292,35 @@ func launchOnSwitch(cmd []string) (string, error) {
...
-	reader := bufio.NewReader(out)
-	var line string
 
 	go func() {
-		for {
-			line, err = reader.ReadString('\n')
-			if err == io.EOF {
+		for ctx.Err() == nil {
+			out, err := executor.ExecCommandPipe(ctx, cmd[0], cmd[1:]...)
+			if err != nil {
```
nplanel: log the error here like
logging.GetLogger().Errorf("Can't execute command %v", cmd)

topology/probes/ovs_of.go
```diff
@@ -494,6 +502,7 @@ func (o *OvsOfProbe) OnOvsBridgeAdd(bridgeNode *graph.Node) {
 		return
 	}
 	bridgeProbe, err := o.NewBridgeProbe(hostName, bridgeName, uuid, bridgeNode)
+	logging.GetLogger().Infof("Probe added for %s on %s", bridgeName, hostName)
```
nplanel: log the uuid as well

topology/probes/ovs_of.go
pierrecregut: Yes missing logging.

topology/probes/ovs_of.go
pierrecregut: Yes and made more consistent with error case.

PR #407, opened 2017-08-30
govendor : use a local storage to speed up govendor sync

nplanel:


nplanel:
BTW old URL is down

nplanel:

safchain:
We shouldn't use gist for that as we can host this file on our servers like before

nplanel:
done

nplanel:

scripts/ci/install-go-deps.sh
```diff
@@ -16,7 +16,7 @@ go get -f -u github.com/tebeka/go2xunit
 export PATH=$PATH:$GOPATH/bin
 
 # speedup govendor sync command
-curl -o /tmp/vendor.tgz http://46.231.132.68:8080/v1/AUTH_0ec9e4f4f3044236b4d18536ccfcb182/skydive/vendor/vendor.tar.gz
+curl -o /tmp/vendor.tgz curl -o /tmp/vendor.tgz http://ci.skydive.network/vendor.tar.gz
```
masco-enovance: look like some copy paste error? curl -o /tmp/vendor.tgz is repeating two times.

nplanel: yep that it

lebauce:
I'm not sure it useful : govendor already has a cache. The problem is that it's in $GOPATH/.cache/govendor and Jenkins cleans the workspace for every build.

So we could simply do a symlink in $GOPATH/.cache/govendor to ~/.govendor

scripts/ci/install-go-deps.sh
```diff
@@ -16,15 +16,6 @@ go get -f -u github.com/tebeka/go2xunit
...
-go get -f -u github.com/kardianos/govendor
-govendor sync -n | perl -pe 's|fetch \"(.*)\"$|vendor/\1|g' | sort -u > vendor.fetch.list
-cat vendor.fetch.list | xargs tar -xvzf /tmp/vendor.tgz --exclude "vendor/vendor.json"
-# remove installed
-find vendor/ -mindepth 2 -type f | xargs dirname | sort -u > vendor.installed.list
-echo "package to be removed/cleanup"
-diff -u vendor.fetch.list vendor.installed.list | grep '^\+v' | perl -pe 's|^\+(.*)|\1|' | tee /dev/stdout | xargs -n 1 rm -rf
-rm -f vendor.fetch.list vendor.installed.list
-popd
+mkdir $HOME/.govendor $GOPATH/.cache
```
lebauce: mkdir -p

scripts/ci/install-go-deps.sh
```diff
@@ -16,15 +16,6 @@ go get -f -u github.com/tebeka/go2xunit
...
+rm -rf $GOPATH/.cache/govendor
```
lebauce: The workspace is cleaned for every build so this line is useless

nplanel: I known it's useless on our CI, but will works all the time, in case using locally

scripts/ci/install-go-deps.sh
nplanel: useless go get above, will create $GOPATH

safchain: I think -p is useful otherwise the mkdir will fail on $HOME/.govendor on the second run

Issue #317, opened 2017-07-20
feature: filter: add ability to filter out topology sub-graph based on Gremlin expression


vperelmax:
We would like to give the ability to user to express using Gremlin language which part of sub-graph to demonstrate.
This feature will help to focus on the relevant part of graph in highly complex network topologies.
Using it we would like for example to show all the paths that connect between two specific hosts or only links of specific layer (such as application layer links and nodes connected via such links).

vperelmax:
Related pull request: https://github.com/skydive-project/skydive/pull/408

PR #408, opened 2017-08-30
feature: filter: Gremlin-based Topology Filter

vperelmax:
Example for test:
1) Add to skydive.yml this section:
  fabric:
         - TOR1[Name=tor1] -> [color=red] TOR1_PORT1[Name=port1, MTU=1500]
         - TOR1[Name=tor1] -> [color=red] TOR1_PORT2[Name=port2, MTU=1500]
         - TOR1[Name=tor1] -> [color=red] TOR1_PORT3[Name=port3, MTU=1500]
2) Try following expression: G.V().Has("Name", Regex("port"))
Expected result: three nodes <port1, port2, port3>

Relates to issue: #317

### DIFF
--- a/analyzer/server.go
+++ b/analyzer/server.go
@@ -170,6 +170,9 @@ func (s *Server) initialize() (err error) {
 	tr.AddTraversalExtension(topology.NewTopologyTraversalExtension())
 	tr.AddTraversalExtension(ftraversal.NewFlowTraversalExtension(tableClient, s.Storage))
 
+	topologyFilter := traversal.NewTraversalGraphFilter(tr)
+	s.TopologyServer.Graph.SetFilter(topologyFilter)
+
 	s.AlertServer = alert.NewAlertServer(alertAPIHandler, s.WSServer, tr, s.EtcdClient)
 
 	piClient := packet_injector.NewPacketInjectorClient(s.WSServer)
--- a/analyzer/server.go
+++ b/analyzer/server.go
@@ -176,8 +176,9 @@ func (s *Server) initialize() (err error) {
 
 	api.RegisterTopologyAPI(s.HTTPServer, tr)
 
-	api.RegisterPacketInjectorAPI(piClient, s.TopologyServer.Graph, s.HTTPServer)
-
+	if config.GetConfig().GetBool("analyzer.inject_enabled") {
+		api.RegisterPacketInjectorAPI(piClient, s.TopologyServer.Graph, s.HTTPServer)
+	}
 	api.RegisterPcapAPI(s.HTTPServer, s.Storage)
 
 	api.RegisterConfigAPI(s.HTTPServer)
--- a/cmd/client/capture.go
+++ b/cmd/client/capture.go
@@ -31,7 +31,7 @@ import (
 	"github.com/skydive-project/skydive/flow"
 	"github.com/skydive-project/skydive/logging"
 	"github.com/skydive-project/skydive/validator"
-
+	"github.com/skydive-project/skydive/config"
 	"github.com/spf13/cobra"
 )
 
@@ -71,6 +71,10 @@ var CaptureCreate = &cobra.Command{
 		}
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		if !config.GetConfig().GetBool("analyzer.capture_enabled") {
+			logging.GetLogger().Critical("Violation of disabled capturing is detected!")
+			os.Exit(1)
+		}
 		client, err := api.NewCrudClientFromConfig(&AuthenticationOpts)
 		if err != nil {
 			logging.GetLogger().Critical(err.Error())
@@ -83,9 +87,14 @@ var CaptureCreate = &cobra.Command{
 		capture.Type = captureType
 		capture.Port = port
 		capture.HeaderSize = headerSize
-		capture.RawPacketLimit = rawPacketLimit
 		capture.ExtraTCPMetric = extraTCPMetric
 		capture.SocketInfo = socketInfo
+
+		if !config.GetConfig().GetBool("analyzer.packet_capture_enabled") {
+			capture.RawPacketLimit = 0
+		} else {
+			capture.RawPacketLimit = rawPacketLimit
+		}
 		if err := validator.Validate(capture); err != nil {
 			logging.GetLogger().Error(err.Error())
 			os.Exit(1)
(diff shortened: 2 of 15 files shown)


vperelmax:
This patch adds new functionality to UI.
Now user should be able to filter graph topology using Gremlin syntax.
In this version we support only Gremlin expressions that filter nodes. Then edges that connect nodes that are both filtered are also added to the topology.

masco-enovance:
ok to test

vperelmax:
The change relates to issue #317

topology/graph/graph.go
```diff
@@ -1058,12 +1064,15 @@ func (g *Graph) String() string {
 
 // MarshalJSON serialize the graph in JSON
 func (g *Graph) MarshalJSON() ([]byte, error) {
+	nodes := g.GetFilteredNodes()
```
safchain: I think we should use the SubGraph step #417, we could use it as a prefix in the WebUI ?

topology/graph/message.go
```diff
@@ -79,6 +79,10 @@ func UnmarshalWSMessage(msg shttp.WSMessage) (string, interface{}, error) {
 			}
 			context.TimeSlice = common.NewTimeSlice(i, i)
 		}
+		switch v := m["topologyFilter"].(type) {
```
safchain: The issue with the current approach is that we only filter on Sync message if a node appears after a sync it will appear on the WebUI even if not matching the filter.

We should add something in the forwarding message mechanism checking if the node belong to the filter. if we use the subgraph step it will be easy (I hope)

1. get the new node
2. get the subgraph 
  G....SubGraph()
3 check if the node is present
  G...SubGraph().V(<ID of node>)
4 if present let forward the message

statics/js/components/capture-form.js
```diff
@@ -80,17 +80,25 @@ Vue.component('capture-form', {
...
-      <button type="button"\
-              id="create-capture"\
-              class="btn btn-primary"\
-              v-else\
-              @click="visible = !visible">\
-        Create\
-      </button>\
+      <div v-else>\
+          <button type="button" id="create-capture" class="btn btn-primary" @click="visible = !visible"> Create</button>\
+          <button type="button" id="capture-all" class="btn btn-primary" @click="captureall">All</button>\
```
safchain: this button and the following one should be removed from this commit

statics/js/components/capture-form.js
```diff
@@ -80,17 +80,25 @@ Vue.component('capture-form', {
...
     </transition>\
   ',
 
   data: function() {
+    $.when(this.$getConfigValue('analyzer.capture_enabled').
+                        then(function(captureEnabled) {
+                        try {
+                            if (captureEnabled.localeCompare("false") === 0) {
+                                document.getElementById("create-capture").disabled = true;
+                                document.getElementById("capture-all").disabled = true;
```
safchain: should be removed too

statics/js/components/capture-form.js
```diff
@@ -80,17 +80,25 @@ Vue.component('capture-form', {
...
+          <button type="button" id="create-capture" class="btn btn-primary" @click="visible = !visible"> Create</button>\
+          <button type="button" id="capture-all" class="btn btn-primary" @click="captureall">All</button>\
+          <button type="button" id="capture-bpf" class="btn btn-primary" @click="captureallbpf">AllBPF</button>\
+      </div>\
     </transition>\
   ',
 
   data: function() {
+    $.when(this.$getConfigValue('analyzer.capture_enabled').
+                        then(function(captureEnabled) {
```
safchain: nit: indentation should be fixed

statics/js/components/capture-form.js
```diff
@@ -80,17 +80,25 @@ Vue.component('capture-form', {
...
+                        try {
+                            if (captureEnabled.localeCompare("false") === 0) {
+                                document.getElementById("create-capture").disabled = true;
```
safchain: maybe set it as disabled by default

flow/ondemand/client/client.go
```diff
@@ -95,6 +96,9 @@ func (o *OnDemandProbeClient) registerProbes(nodes []interface{}, capture *api.C
 	toRegister := func(node *graph.Node, capture *api.Capture) (nodeID graph.Identifier, host string, register bool) {
 		o.graph.RLock()
 		defer o.graph.RUnlock()
+		if !config.GetConfig().GetBool("analyzer.capture_enabled") {
```
safchain: I don't think this check is needed at this point

statics/js/components/capture-form.js
```diff
@@ -80,17 +80,25 @@ Vue.component('capture-form', {
...
         </form>\
       </div>\
-      <button type="button"\
-              id="create-capture"\
-              class="btn btn-primary"\
-              v-else\
-              @click="visible = !visible">\
-        Create\
-      </button>\
+      <div v-else>\
```
safchain: We should do the same thing for injection, I mean removing the content in case of disabled

statics/js/components/capture-form.js
```diff
@@ -80,17 +80,25 @@ Vue.component('capture-form', {
...
+          <button type="button" id="create-capture" class="btn btn-primary" @click="visible = !visible"> Create</button>\
```
safchain: We should add a tittle explaining that the button is disabled because of ACL

cmd/client/capture.go
```diff
@@ -71,6 +71,10 @@ var CaptureCreate = &cobra.Command{
 		}
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		if !config.GetConfig().GetBool("analyzer.capture_enabled") {
```
safchain: I don't think we should put this test here as it depends on the config file read by the client and the config of the client could allow the capture creation. Check on the API side should be enough

analyzer/server.go
```diff
@@ -176,8 +176,9 @@ func (s *Server) initialize() (err error) {
 
 	api.RegisterTopologyAPI(s.HTTPServer, tr)
 
-	api.RegisterPacketInjectorAPI(piClient, s.TopologyServer.Graph, s.HTTPServer)
-
+	if config.GetConfig().GetBool("analyzer.inject_enabled") {
```
safchain: I propose to add a function in order to return if the resource accepts write operation : https://github.com/skydive-project/skydive/blob/master/api/handler.go#L46, (ex: IsEnabled). So that in the Create/Delete functions we check whether the handler is allowed to modify the resource or not. If not we could return http.StatusForbidden.

safchain:
Already merged

Issue #409, opened 2017-08-31
In v0.12.0 inject packet check depends on netns is a error desgin


jonnary:
1.In v0.12.0 we change the inject packet and binding it to netns.This cause if the interface not in netns, it will can't be inject packet. 
For exampe, In openstack devstack enviroment, the vm's tap/qvo interfaces all not IPV4 ip in the interface, but we can see the Neutron ips in ui, 
2.This situation in v0.11.0 we can inject packet by skydive client

jonnary:
Have discussed with masco.

ji-xuepeng:
I have the same problem in v0.12.0
In my openstack devstack environment, the virtual ports of the VMs do not belong to any of the network namespaces, and the packet injection fails. 
In fact, nothing is shown when ip netns command is executed on the compute node.

safchain:
The issue is not related to netns. In fact tap doesn't have IPV4 but we report the IPV4 that the VM will use in the Neutron metadata section. 

Before this patch https://github.com/skydive-project/skydive/commit/f010aec562200aef936c1781a3904e64b8d20968
metadata of Neutron were replicated to the first level of metadata that's why the packet injector was able to find an IP. This was an error since the tap interface in reallity doesn't have an IP.

I think this should be fixed in the WebUI, adding something to help user to spoof the IP from the Neutron metadata

ji-xuepeng:
@safchain, I can not  inject packet successfully also when I use v0.11.0.  In my environment, there are two VMs created by openstack. OVS is uesed in the  environment. 
VM1:
IP:10.10.0.11
MAC:fa:16:3e:95:b0:b8
Interface:tapb3238674-88 bridge     qbrb3238674-88 virtio      fa:16:3e:95:b0:b8
VM2:
IP:10.10.0.6
IP:fa:16:3e:ae:10:c6
Interface:tapeec17485-2d bridge     qbreec17485-2d virtio      fa:16:3e:ae:10:c6

root@compute1:~# brctl show
bridge name     bridge id               STP enabled     interfaces
qbrb3238674-88          8000.36b39499a407       no              qvbb3238674-88
                                                        tapb3238674-88
qbreec17485-2d          8000.aa3935cbd0c4       no              qvbeec17485-2d
                                                        tapeec17485-2d
When I ping from vm1 to vm2, I can capture ICMP messages normally, as shown below:
root@compute1:~# tcpdump -vv -i tapb3238674-88
[...]

PR #419, opened 2017-09-06
packet-inject: fix injection outside of netns

safchain:
Closes: #409


safchain:
Test and other fixes will come

PR #420, opened 2017-09-06
pi: allow user to provide IP from UI

masco-enovance:


masco-enovance:
run selenium-tests

statics/js/components/inject-form.js
```diff
@@ -22,10 +22,13 @@ Vue.component('inject-form', {
                        id="inject-src"\
                        attr="id"\
                        v-model="node1"></node-selector>\
+        <input id="src-ip" v-if="noSrcIP" class="form-control input-sm inject-target" v-model="srcIP" placeholder="No IP in src node, please enter"/>\
```
safchain: I think it could be nice add a way to force the IP as for most of the interface with the IPV6 stack have a link-local address.

masco-enovance: so we need to get the IP always from user? if no IP given by user should we have to throw a form error?

safchain: not always but having a way (button or something else) to allows the user to force the IP.

masco-enovance: got your point. thanks!

PlutoCode: I think it's better to show ip in the box when this node has ip addr
This will  looks more clearly

statics/js/components/inject-form.js
```diff
@@ -22,10 +22,13 @@ Vue.component('inject-form', {
                        id="inject-src"\
                        attr="id"\
                        v-model="node1"></node-selector>\
+        <input id="src-ip" class="form-control input-sm inject-target" v-model="srcIP" placeholder="Src IP"/>\
         <node-selector placeholder="To"\
+                       class="inject-target"\
                        attr="id"\
                        id="inject-dst"\
                        v-model="node2"></node-selector>\
+        <input id="dst-ip" class="form-control input-sm" v-model="dstIP" placeholder="Dst IP"/>\
```
safchain: I uploaded a draft in order to make it a bit easier to understand (IHMO) https://da.gd/jHPbO. We could fill the IP text input with the auto selected IP if present and let the user override it.

masco-enovance: with "form-group" class, we can't have 100% width. so I will do the require modification with input-group and submit the code.

safchain: I think that is not a big issue if the input doesn't reach the right end of the screen (IPs are short) but if you can have the IP (label) and the input text wider it will be perfect.

masco-enovance:
rerun tests

safchain:
rerun tests

safchain:
run cdd-overview

PR #428, opened 2017-09-11
ui: expand/collapse nodes by level

masco-enovance:


statics/js/components/topology.js
```diff
@@ -263,25 +264,26 @@ var TopologyComponent = {
...
-      this.layout.collapse(this.collapsed);
-    },
-
     toggleExpandAll: function(node) {
       this.layout.toggleExpandAll(node);
     },
 
+    expandNextLevel: function() {
+	this.layout.collapseByLevel(this.layout.collapseLevel, !true, this.layout.groups);
+        this.layout.collapseLevel++;
```
eonpatapon: Feels to me that this.layout.collapseLevel and this.layout.groups should be handled internally by this.layout,
here we should just need to call this.layout.collapseByLevel(true || false)

safchain: I agree with @eonpatapon that the collapseLevel should be part of the layout. maybe you could add two function one for increase the other for decrease in the layout.

statics/js/components/topology.js
```diff
@@ -37,6 +35,10 @@ var TopologyComponent = {
               </i>\
             </span>\
           </button>\
+          <button id="expand" type="button" class="btn btn-primary" \
+                  title="Expand" @click="expandNextLevel">++</button>\
```
safchain: I would prefer to use an awesome font like "fa-expand" and "fa-compress"

statics/js/components/topology.js
```diff
@@ -27,8 +27,6 @@ var TopologyComponent = {
           </button>\
           <button id="zoom-reset" type="button" class="btn btn-primary" \
```
safchain: It is not related to your patch but I think we will have too much button and this one seems to be useless. The "fit" one is quite similar and more useful. Maybe it is time to remove it ?

statics/js/components/graph-layout.js
```diff
@@ -915,6 +916,34 @@ TopologyGraphLayout.prototype = {
...
+//        } else if (!groups[i].parent) {
+          } else {
+          this.uncollapseGroup(groups[i]);
+        }
+      }
+      this.update();
+    } else {
+      var i;
+      for (i = groups.length - 1; i >= 0; i--) {
+        this.collapseByLevel((level-1), collapse, this.getGroupArray(groups[i].children));
```
safchain: Object.values(groups[i].children)) instead of adding a new function

statics/js/components/graph-layout.js
```diff
@@ -915,6 +916,34 @@ TopologyGraphLayout.prototype = {
...
+          this.collapseGroup(groups[i]);
+//        } else if (!groups[i].parent) {
+          } else {
+          this.uncollapseGroup(groups[i]);
+        }
+      }
+      this.update();
+    } else {
+      var i;
+      for (i = groups.length - 1; i >= 0; i--) {
```
safchain: nit: for (var i =

statics/js/components/graph-layout.js
```diff
@@ -915,6 +916,34 @@ TopologyGraphLayout.prototype = {
     this.update();
   },
 
+  collapseByLevel: function(level, collapse, groups) {
+    if (level == 0) {
+      var i;
```
safchain: move var at the beginning of the function as you use it in if and else and var scope is function

statics/js/components/topology.js
masco-enovance: done

statics/js/components/topology.js
masco-enovance: yes. will remove it.

statics/js/components/topology.js
masco-enovance: yes, it will be awesome to have the icon but to 'expand the node tree' functionality, I am using those buttons. I can't use the same it will confuse the user :(

statics/js/components/graph-layout.js
masco-enovance: done

statics/js/components/graph-layout.js
masco-enovance: previous declaration of 'i' can be used here, so just removing the var i

statics/js/components/graph-layout.js
masco-enovance: done

statics/js/components/topology.js
```diff
@@ -37,6 +33,10 @@ var TopologyComponent = {
               </i>\
             </span>\
           </button>\
+          <button id="expand" type="button" class="btn btn-primary" \
+                  title="Expand" @click="toggleCollapseByLevel(false)">++</button>\
```
safchain: I think we can reuse the same icon as it means expand and the context is different or at least we can find something else. I think using ++/-- break the look and feel and lead to think adding/deleting.

masco-enovance: if we use the same icon it will confuse the user. I will do one thing, use the icon for step by step expand and collapse. use a word button for complete tree expand as of now. later we can change with the appropriate icon. what do you think?

safchain: what do you mean by "tree expand" I didn't get this point

masco-enovance: i am using fa-expand and fa-collapse icons here(https://github.com/skydive-project/skydive/blob/master/statics/js/components/topology.js#L32) so I can't use the same icons instead of "++" and "--". my question was shall i use fa icons for ++ and -- and use text in https://github.com/skydive-project/skydive/blob/master/statics/js/components/topology.js#L32

safchain: ok got your point. I don't find this really confusing but I would propose to keep the tree expand as it is right now and to use new icons for '++'/'--'. What do you think about this
http://jsfiddle.net/3k3u5qs4/
it combines two icons

safchain:
Will change text to icon in a coming PR

PR #430, opened 2017-09-11
Ipv4range gremlin step

safchain:


flow/traversal/traversal.go
```diff
@@ -310,8 +310,18 @@ func paramsToFilter(params ...interface{}) (*filters.Filter, error) {
 			return nil, errors.New("keys should be of string type")
 		}
 
-		if v, ok := params[i+1].(string); ok && (k == "Network" || k == "Link" || k == "Transport") {
-			filter = filters.NewOrFilter(filters.NewTermStringFilter(k+".A", v), filters.NewTermStringFilter(k+".B", v))
+		if k == "Network" || k == "Link" || k == "Transport" {
```
nplanel: cool :+1:

matrohon:
I'm having issues using the client with this patch :
```
skydive client --analyzer 127.0.0.1:8082 topology query --gremlin "G.V().Has('IPV4', IPV4Range('10.193.0.0/16'))"

2017-09-12T10:29:43.034+0200	ERROR	client/topology.go:61 glob.func18	t460: 400 Bad Request: error parsing regexp: unexpected ): `^(10\.193\.)([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]\.)[0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9])$`
```

safchain:
@matrohon the predicate is not currently working with V() step but I will add another commit for that. Otherwise there is a issue when using /16. going to fix that. Thanks for testing

matrohon:
using a /24 instead of the /16 doesn't trigger the error, but the node is not retrieved

safchain:
@matrohon  the /24 /16 should be fixed now, still need to address list value

safchain:
@matrohon this version should work with list on IPs so with V() step

matrohon:
@safchain quickly tested, it works like a charm, thanks 👍

PR #432, opened 2017-09-11
flow: report afpacket stats as for pcap

safchain:


lebauce:
It seems the stats goroutine is not stopped properly so the stats are updated after the capture has been removed.

In addition, the metadata still has a "Capture" dict so the WebUI still displays the red button.

I think the bugs were already there with pcap but we do not use the "pcap" capture type very often, as AFPacket is the default

PR #441, opened 2017-09-13
govendor : use latest version of github.com/kardianos/govendor

nplanel:


nplanel:
which version of go did you use ? (I use go 1.9)

nplanel:
same issue with go 16

safchain:
version 1.8.3
@masco-enovance could you please test this to see if we you the same result as Nicolas

safchain:
Seems to be OK regarding to the CI resutls

Issue #426, opened 2017-09-10
Add Gremlin Union step


eranra:
In order to allowed support for even simple use cases, for example get all nodes connected to a single node and the node itself (i.e. myself + all nodes surrounding me one edge away) there is a need to add the union step to Gremlin.

So for example in this case, it will be possible to execute::
G.V().**union**(Has('Name', 'host1'),Has('Name', 'host1').Out())

safchain:
G.V().Has('Name', within('host1, 'host2').Out() should work

safchain:
G.V().Has('Name', within('host1, 'host2')).Out()

eranra:
I tried this ::

G.V().Has('Name', within('kube-dal10-crc4d48704bff84452b2a7ad78340377aa-w3.cloud.ibm', 'kube-dal10-crc4d48704bff84452b2a7ad78340377aa-w3.cloud.ibm')).Out()


On the demo env, but it;'s not working ... to explain ... I want to see the host iteslf, and all the connected interfaces etc components one edge away ... so I have only one host not two ... in your example you explained how to see two hosts, this is not what I'm trying to solve.

safchain:
sorry I mis-read your comment. So maybe this could be used:

G.V().Has('Name', 'host1').OutE().BothV ? but you will need this #416

eranra:
Should work ... I need to check ... I ran::
G.V().Has('Name', 'kube-dal10-crc4d48704bff84452b2a7ad78340377aa-w3.cloud.ibm').OutE().OutV()
and 
G.V().Has('Name', 'kube-dal10-crc4d48704bff84452b2a7ad78340377aa-w3.cloud.ibm').OutE().InV()

and both looks good ... so I assume you are correct and BothV() will solve (maybe need also to add the dedup stage). In any event I think that the flexibility we will get with Union will be useful forward and the syntax will be more understandable.

BTW; There is another option which is to extend InV and outV to include also the node itself  ... this might be another nice alternative.

eranra:
Waiting for #416 to see this working.

safchain:
We will add the as/select steps to address the union

safchain:
Closing it for now as select/as has been added and handles the use cases

PR #450, opened 2017-09-20
Add support for gremlin union (or) step

rkhachatryan:
Hello,
I'd like to provide an implementation of UNION/OR step as was proposed in #426  (see [docs](http://gremlindocs.spmallette.documentup.com/#or) )

I think explicit OR step will give more abilities than existing/upcoming steps, for example to combine results of different steps: `G.V().Or(Has('Value', 2).In(), Has('Value', 3).Out())`
Also using explicit OR may be more convenient.
Further NOT and AND steps can be implemented based on this changes.

Overview of changes
1. added new step, it's scanning and execution
1. added new method to scan for nested steps in parentheses
1. changed parsing to parse nested steps
1. added unit tests, changed existing - extracted some common code

(tried to keep changes minimal to not affect existing functionality)

NOTES: 
1. added scan method is able to scan strings with arbitrary depth and also deals with escaping
1. infix notation not implemented
1. underscore in not needed and not interpreted in the beginning of nested sequences


safchain:
ok to test

safchain:
@rkhachatryan Thx for submitting this :) we will have a deep look at it. Meanwhile I'm adding you to the whitelist for the CI so that if you push updates, jobs will be triggered automatically.

safchain:
add to whitelist

rkhachatryan:

safchain:

eonpatapon:
I think this step name should be just or(), not union().

If you look at the gremlin (tinkerpop) doc union() is a branching step that yieds multiple traversals whereas or() is a filter step for the current traversal

safchain:

rkhachatryan:
Thanks for reviewing!

You are right, I didn't make difference between these two steps. And currently this PR implements UNION rather than OR.

Implementing OR will require keeping track of which elements produce which elements when executing the each step.

So now I see the following options:
1. Keep track of which graphElement from initialStep has outputs and which hasn't. To do this we'll need to modify each Gremlin Step because they transform one set of elements to another. This will require a lot of changes.
1. For each graphElement from initial step create a fake step and Exec GremlinTraversalSequence to determine whether it has outputs or not. This may cause problems in queries like g.V().or(_().V(2))
1. Just rename current step to Union

I propose just to change the name to union (and remove OR token).

safchain:
Yes keeping the union behavior sounds good. Could you just convert the WrapperGraphTraversalStep to something like GraphTraversalSet which could be

type GraphTraversalSet struct {
  vs []GraphTraversalV
  es []GraphTraversalE
  err error
}

so that we fill this structure with the results of the union. We could add "sub" steps for this step later like subgraph ?

rkhachatryan:
When we build this struct we only have a collection of GraphTraversalStep as a result of executing GremlinTraversalSequence s. 

That can be not only GraphTraversalE/V but also GraphTraversalShortestPath or something else.

If the next step needs to differentiate these types then why not to do it there in step.Exec method?

safchain:
Right for the collection of GraphTraversalStep. Your approach looks good to me.

safchain:
@rkhachatryan do you have any news on this ?

rkhachatryan:
@safchain, right now I don't have time to work on this, maybe on weekend.

safchain:
@rkhachatryan no problem :)

safchain:
@rkhachatryan Hi, did/do you have time to work on this ? can we help you maybe in order to have it merge as it is a nice feature ?

safchain:
Closing it for now since as/select steps have been introduced handling the use cases

PR #472, opened 2017-09-29
api: fix segfault in dot output with short ids

lebauce:


lebauce:
run skydive-coverage

lebauce:
run skydive-coverage

PR #483, opened 2017-10-03
ci : add required dpdk packages to build skydive

nplanel:


nplanel:
we need to relay on a uptodate dpdk version

PR #489, opened 2017-10-09
Gremlin filtering

lebauce:


lebauce:
run devstack

lebauce:
run packaging-tests

lebauce:
run vagrant-tests

lebauce:

lebauce:

lebauce:

vperelmax:
General main comments:
 - I dont agree that each gremlin expression should be finished by SubGraph() 
 - I am also not sure about the need in change in passing gremlin filter via http header, in our change we didnt add any special mechanism and sinse we here keep for each subscriber its query - such thing is a new code without a real need
- I explained it - in the commit of Diff - I dont think it worth all the code, additional computation and memory spent in order to send to each subscriber its diff rather than its subgraph, which is anyway small (and if it is not small - we spend too much memory on the copies)
-We don't need the change of separating in gremlin graph and sending to it a graph at the query time. In my patch I didnt do it and it worked well. Anyway  we always run the query on the same original graph of analyzer and not the subscriber copies.

Generally my impression - we add too much code for a relatively small feature and it might be easier that I will merge only small part of it into our filter change (only the part of keeping query of subscriber and update of subscribers upon add/remove of a node). Alternative way which I still think is legal - to add in JS of client upon node update event received from the analyzer - call for sync. We tried it and it didnt look like too big overhead and keeps the code changes minimal.

I think all these points should be discussed.

Some additional features that we support - regular expressions (I am not sure that we have uploaded the change) - but this thing is ignorable, I agree.

lebauce:
@vperelmax The PR is indeed pretty big but the biggest part is only to allow a client to send its service type and filter though the URL query (?x-client-type=webui&x-gremlin-filter=....)

The filtering itself is pretty small : https://github.com/vperelmax/skydive/commit/704caf563098c865b507f49436c15ac0ffa5af3e

lebauce:

analyzer/topology_server.go
```diff
@@ -214,13 +244,41 @@ func (t *TopologyServer) OnWSJSONMessage(c shttp.WSSpeaker, msg *shttp.WSJSONMes
 	// this kind of message usually comes from external clients like the WebUI
 	if msgType == graph.SyncRequestMsgType {
 		t.Graph.RLock()
-		context, status := obj.(graph.GraphContext), http.StatusOK
-		g, err := t.Graph.WithContext(context)
+		context, status := obj.(graph.SyncRequestMsg), http.StatusOK
```
safchain: I would change context to syncMsg

analyzer/topology_server.go
```diff
@@ -214,13 +244,41 @@ func (t *TopologyServer) OnWSJSONMessage(c shttp.WSSpeaker, msg *shttp.WSJSONMes
...
+			t.subscribers[c.GetHost()] = &topologySubscriber{graph: g, ts: ts, gremlinFilter: context.GremlinFilter}
+		}
+
+		if subscriber, found := t.subscribers[c.GetHost()]; found {
+			gts, err := subscriber.ts.Exec()
+			if err != nil {
+				logging.GetLogger().Errorf("Graph: unable to get a graph with filter %s: %s", subscriber.gremlinFilter, err.Error())
+				result, status = nil, http.StatusBadRequest
+			}
+			result = gts.Values()[0]
```
safchain: I think we should check that we have a least something in the array ?

PR #495, opened 2017-10-10
http: accept parameters as http headers or query parameters

lebauce:


lebauce:
run skydive-scale-tests

config/config.go
```diff
@@ -273,3 +273,13 @@ func IsTLSenabled() bool {
 	}
 	return false
 }
+
+func GetURL(protocol string, addr string, port int, path string) *url.URL {
```
nplanel: would make sense to move it in common or http package

PR #496, opened 2017-10-10
neutron: do not block anymore waiting for neutron

safchain:
This patch removes also the cache using a map instead

Closes: #148


safchain:
run devstack-plugin

safchain:
run devstack

safchain:
run devstack

PR #501, opened 2017-10-11
Fix red button remaining 

safchain:


flow/ondemand/client/client.go
```diff
@@ -206,12 +208,27 @@ func (o *OnDemandProbeClient) onNodeEvent() {
 
 // OnNodeAdded graph event
 func (o *OnDemandProbeClient) OnNodeAdded(n *graph.Node) {
-	o.onNodeEvent()
+	// a node comes up with already a capture, this could be due to a re-connect of
+	// an agent. Check if the capture is still active.
+	if id, err := n.GetFieldString("Capture.ID"); err == nil {
+		for _, capture := range o.captures {
```
lebauce: I think you can do 
```
if capture, found := o.captures[id]; found {
    return
}
```

lebauce: By the way, it seems we don't take an RLock to access in o.captures (it may already have been the case before your patch)

safchain:
run scale-tests

safchain:
run scale-tests

safchain:
run scale-tests

safchain:
run scale-tests

PR #504, opened 2017-10-13
Neutron: add subnet cidr to neutron ips

zhuzeyu:
IPs of Neutron don't have cidr, it is unfavorable for reading
We can add subnet cidr to the ips

### DIFF
--- a/topology/probes/neutron.go
+++ b/topology/probes/neutron.go
@@ -35,6 +35,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/provider"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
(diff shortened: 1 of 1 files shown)


topology/probes/neutron.go
```diff
@@ -163,7 +164,12 @@ func (mapper *NeutronProbe) retrieveAttributes(portMd portMetadata) (*attributes
 
 	var IPs []string
 	for _, element := range port.FixedIPs {
-		IPs = append(IPs, element.IPAddress)
+		subnet, err := subnets.Get(mapper.client, element.SubnetID).Extract()
+		if err != nil {
+			return nil, err
+		}
+		neutron_ips := element.IPAddress + "/" + strings.Split(subnet.CIDR, "/")[1]
```
safchain: thanks for submitting this. could you please just change neutron_ips to neutronIPs mainly because the golinter will raise a warning. Otherwise it is ok for me

zhuzeyu: okay， it is my fault

safchain:
@zhuzeyu thanks ;)

PR #536, opened 2017-10-26
socketinfo: disable it by default and apply it only on tcp

safchain:
This enhancer has a huge impact on performances especially
with many new flows


safchain:
rerun skydive-scale-tests

PR #538, opened 2017-10-26
websocket: add AddClient for wsIncomerPool to not loose the type

safchain:
Using the same AddClient implementation for WSPool and wsIncomerPool
leads to loose the original while add the listener. Then the on
disconnected callback is not the right one but always the one of
WSPool.

This causes to not disconnect the WebSocket Flow client.


vperelmax:
Wow! Great!!! the bug was really annoying!!! :)

PR #548, opened 2017-10-31
ui: node select should disable after error click

masco-enovance:
in capture select function should be disabled
after an error click


masco-enovance:

PR #555, opened 2017-11-02
topology: fix ShortestPathTo by not testing path longest that we found already

nplanel:


topology/graph/graph.go
```diff
@@ -702,6 +702,14 @@ func (g *Graph) StartMetadataTransaction(i interface{}) *MetadataTransaction {
 }
 
 func (g *Graph) lookupShortestPath(n *Node, m Metadata, path []*Node, v map[Identifier]bool, em Metadata) []*Node {
+	var shortest int
```
safchain: no need to add a new function you can add this in LookupShortestPath

safchain:

PR #558, opened 2017-11-06
Find: Highlight selected nodes

vperelmax:
Initial Implementation
Find on graph selected via Gremlin Expression sub-graph and highlight it.

### DIFF
--- a/statics/js/components/topology.js
+++ b/statics/js/components/topology.js
@@ -14,8 +14,8 @@ var TopologyComponent = {
 	          <slider v-if="history" :min="timeRange[0]" :max="timeRange[1]" \
 	                v-model="time" :info="topologyTimeHuman"></slider>\
 	          <div class="form-group input-sm" style="width: 450px">\
-	              <label for="topology-filter">Filter</label>\
-	              <input list="topology-filter-list" placeholder="e.g. Name : TOR" \
+	              <label for="topology-filter" style="width: 33px">Filter</label>\
+	              <input list="topology-filter-list" placeholder="e.g. g.V().Has(,)" \
 	              @click="topologyFilterClear" \
 	              id="topology-filter" type="text" style="color: black;width: 350px" \
 	              v-model="topologyFilter" @keyup.enter="topologyFilterQuery"></input>\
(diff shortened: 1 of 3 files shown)


statics/js/components/topology.js
```diff
@@ -309,6 +332,31 @@ var TopologyComponent = {
...
+    highlightSelectedNodes: function(bool) {
+      var self = this;
+
+      this.$topologyQuery(this.topologySelect)
+        .then(function(nodes) {
+          nodes.forEach(function(n) {
+            for (var i in n.Nodes) {
+                var myNode = n.Nodes[i];
+                if (bool) {
+                    self.$store.commit("highlight", myNode.ID);
```
safchain: you should use the self.layout.highlightNodeID function as it is the layout which has to handle this.

statics/js/components/topology.js
```diff
@@ -122,6 +135,7 @@ var TopologyComponent = {
       timeRange: [-120, 0],
       collapsed: false,
       topologyFilter: "",
+      topologySelect: "",
```
safchain: nit: I would not use "topologySelect" as it lead to think that we select something. Here it is just matter of highlighting

statics/js/components/topology.js
vperelmax: sorry for a stupid question - do you mean without $store? I thought I want to do it asynchronously. Actually copied the code from capturing.

safchain: There is a dedicated function here: 
https://github.com/skydive-project/skydive/blob/master/statics/js/components/topology.js#L164

I think we need to change the capture stuff but that another story. So if you wanted to stay with your implementation that is OK for me I will change it in my "sliding panel" patch.

vperelmax:
I have issue with cleaning old selection before running a new one - I will fix it and upload once again.

statics/js/components/topology.js
vperelmax: I will change it

vperelmax: It looks like it has bug.. I will check it - the subscribe function sends to highlightNodeID one parameter, while highlightNodeID in layout receives two parameters and doesnt actually use the second one...
Maybe my misunderstanding of JS :(

vperelmax:
1) The issue was resolved and I have fixed also everything to "highlight"
2) Now user should not use the suffix "SubGraph()" in both filter and highlight requests and the selection is done on the flight (changing state of the graph every time the user types ")")

Issue #556, opened 2017-11-05
Filtrer: ShortestPath is unsupported


vperelmax:
Cant filter by ShortestPath()

Possible issue with list of lists returned by ShortestPath and unacceptable by SubGraph()

safchain:
The shortestpath step should implement at least the SubGraph step.

PR #562, opened 2017-11-07
Fix: Add ability to SubGraph to receive output of ShortestPath

vperelmax:
Small change to add ability for SubGraph() to work on the output
of ShortestPathTo that returns list of lists

To check the functionality run:
g.V().Has("Name", "port1").ShortestPathTo(Metadata("Name", "port2"))
as a filter or find

Bug Fix: #556


vperelmax:
Added Unit test covering current fix.

PR #564, opened 2017-11-07
fabric: add the ability to set l2OnlyLink with -->

matrohon:


scripts/scale.sh
```diff
@@ -325,10 +325,13 @@ analyzer:
     fabric:
 EOF
 
-	TOTAL_AGENT=$(( $AGENT_NUM + $AGENT_STOCK ))
+	TOTAL_AGENT=$(( $AGENT_NUM ))
```
safchain: the scale test is failing, I didn't check yet why but you shouldn't remove the AGENT_STOCK. It is used to provision agent ports that will be used later by the test in order to start more agents

topology/probes/fabric.go
```diff
@@ -162,10 +162,16 @@ func NewFabricProbe(g *graph.Graph) *FabricProbe {
 
 	list := config.GetConfig().GetStringSlice("analyzer.topology.fabric")
 	for _, link := range list {
-		pc := strings.Split(link, "->")
-		if len(pc) != 2 {
-			logging.GetLogger().Errorf("FabricProbe link definition should have two endpoint: %s", link)
-			continue
+		l2OnlyLink := false
+		pc := strings.Split(link, "-->")
```
safchain: could be nice to add an example in the default config file

matrohon: ack

scripts/scale.sh
matrohon: Ack, I didn't understand the meaning of AGENT_STOCK, and it was complicating the displayed graph.

safchain:
run cdd-overview

scripts/scale.sh
```diff
@@ -327,8 +327,11 @@ EOF
 
 	TOTAL_AGENT=$(( $AGENT_NUM + $AGENT_STOCK ))
   for AGENT_I in $( seq $TOTAL_AGENT ); do
-		echo "      - TOR1 -> TOR1_PORT_$AGENT_I" >> $TEMP_DIR/$NAME.yml
-		echo "      - TOR1_PORT_$AGENT_I -> *[Name=agent-$AGENT_I]/eth0" >> $TEMP_DIR/$NAME.yml
+		echo "      - TOR -> TOR_PORT_TOR_$AGENT_I" >> $TEMP_DIR/$NAME.yml
```
lebauce: Is it possible to make it an option of the scale.sh ? We use it a lot for demo and it makes the topology quite complex

lebauce: Is it possible to make it an option of the scale.sh ? We use it a lot for demo and it makes the topology quite complex

matrohon: You want it to be on an argument of the scale.sh script?

safchain: yes, having this extra TOR topology config optional

matrohon: I just pushed a simpler version still uses "-->".
A feel like a better approach would be to use UT.

safchain: So you can modify the scale_test.go to add check on the layer2 link you used.

safchain:
run cdd_overview

safchain:
run cdd-overview

scripts/scale.sh
```diff
@@ -327,8 +327,8 @@ EOF
 
 	TOTAL_AGENT=$(( $AGENT_NUM + $AGENT_STOCK ))
   for AGENT_I in $( seq $TOTAL_AGENT ); do
-		echo "      - TOR1 -> TOR1_PORT_$AGENT_I" >> $TEMP_DIR/$NAME.yml
-		echo "      - TOR1_PORT_$AGENT_I -> *[Name=agent-$AGENT_I]/eth0" >> $TEMP_DIR/$NAME.yml
+		echo "      - TOR -> TOR_PORT_$AGENT_I" >> $TEMP_DIR/$NAME.yml
+		echo "      - TOR_PORT_$AGENT_I --> *[Name=agent-$AGENT_I]/eth0" >> $TEMP_DIR/$NAME.yml
```
safchain: could you please change them here too: 
https://github.com/skydive-project/skydive/blob/master/devstack/plugin.sh#L158
https://github.com/skydive-project/skydive/blob/master/contrib/vagrant/setup-analyzer.sh#L26

matrohon: done

safchain:

safchain:

PR #567, opened 2017-11-09
ui: add capture type in capture form

masco-enovance:


statics/js/components/capture-form.js
```diff
@@ -162,12 +184,51 @@ Vue.component('capture-form', {
           self.resetQueryNodes();
           self.queryNodes = nodes;
           self.highlightQueryNodes(true);
+          if (nodes.length == 1) {
+            self.typeAllowed = true;
+            self.nodeType = nodes[0].Metadata["Type"];
```
safchain: better to write nodes[0].Metadata.Type

statics/js/components/capture-form.js
```diff
@@ -162,12 +184,51 @@ Vue.component('capture-form', {
...
         .fail(function() {
           self.resetQueryNodes();
+          self.typeAllowed = true;
         });
-    }
+    },
+
+    node1: function(newNode) {
+      var node = this.$store.state.currentNode;
+      this.nodeType = node.metadata["Type"];
```
safchain: same here nodes[0].Metadata.Type

statics/js/components/capture-form.js
```diff
@@ -162,12 +184,51 @@ Vue.component('capture-form', {
...
+    node2: function(newNode) {
+      if (newNode) {
+        this.typeAllowed = false;
+      } else {
+        this.typeAllowed = true;
+      }
+    },
 
+    mode: function(newMode) {
+      if (newMode == "selection") {
```
safchain: better to use === instead of ==

statics/js/components/capture-form.js
```diff
@@ -179,6 +240,25 @@ Vue.component('capture-form', {
...
       this.socketInfo = false;
       this.visible = false;
+      this.captureType = "";
+    },
+
+    checkQuery: function(query) {
+      var self = this;
+      this.$topologyQuery(query)
+        .then(function(nodes) {
+          if (nodes.length == 1) {
```
safchain: use ===

statics/js/components/capture-form.js
```diff
@@ -179,6 +240,25 @@ Vue.component('capture-form', {
...
+            self.nodeType = nodes[0].Metadata["Type"];
```
safchain: Metadata.Type

statics/js/components/capture-form.js
```diff
@@ -53,6 +53,13 @@ Vue.component('capture-form', {
             </div>\
             <div id="one" class="panel-collapse collapse">\
               <fieldset class="form-group">\
+                <div class="form-group" v-if="typeAllowed">\
+                  <label for="capture-type">Capture Type</label>\
+                  <select id="capture-type" v-model="captureType" class="form-control input-sm">\
+                    <option disabled value="">Select capture type</option>\
+                    <option v-for="option in options"> {{ option }}</option>\
```
safchain: nit: remove extra space before {{ option }}

statics/js/components/capture-form.js
```diff
@@ -114,6 +124,18 @@ Vue.component('capture-form', {
 
   computed: {
 
+    options: function() {
+      var options = {};
+      var allowedTypes = ["ovsbridge", "device", "internal", "veth", "tun", "bridge", "dummy", "gre", "bond", "can", "hsr", "ifb", "macvlan", "macvtap", "vlan", "vxlan", "gretap", "ip6gretap", "geneve", "ipoib", "vcan", "ipip", "ipvlan", "lowpan", "ip6tnl", "ip6gre", "sit", "dpdkport"];
```
safchain: could be nice to use a common list as this one is already defined in node_selector.js, what do you think about moving them to api.js ?

statics/js/components/capture-form.js
```diff
@@ -162,12 +184,51 @@ Vue.component('capture-form', {
           self.resetQueryNodes();
           self.queryNodes = nodes;
           self.highlightQueryNodes(true);
+          if (nodes.length == 1) {
+            self.typeAllowed = true;
+            self.nodeType = nodes[0].Metadata["Type"];
+          } else if (nodes.length > 1) {
+            self.typeAllowed = false;
+          } else {
+            self.typeAllowed = true;
```
safchain: I think we should only set it to true if nodes.length == 1, otherwise if the nodes selected are not on the same layer2 we will get no node and will be able to select something while the capture is not matching anything

statics/js/components/capture-form.js
```diff
@@ -53,6 +53,13 @@ Vue.component('capture-form', {
             </div>\
             <div id="one" class="panel-collapse collapse">\
               <fieldset class="form-group">\
+                <div class="form-group" v-if="typeAllowed">\
```
safchain: I would prefer to let the div displayed but marking the "select" input as disabled

statics/js/components/capture-form.js
```diff
@@ -162,12 +184,51 @@ Vue.component('capture-form', {
           self.resetQueryNodes();
           self.queryNodes = nodes;
           self.highlightQueryNodes(true);
+          if (nodes.length == 1) {
```
safchain: nit: ===

statics/js/components/capture-form.js
```diff
@@ -162,12 +184,51 @@ Vue.component('capture-form', {
...
+            self.nodeType = nodes[0].Metadata["Type"];
+          } else if (nodes.length > 1) {
+            self.typeAllowed = false;
+          } else {
+            self.typeAllowed = true;
+          }
         })
         .fail(function() {
           self.resetQueryNodes();
+          self.typeAllowed = true;
```
safchain: I think it should be false by default as we can't determine the node types

statics/js/components/capture-form.js
```diff
@@ -105,6 +112,9 @@ Vue.component('capture-form', {
       userQuery: "",
       mode: "selection",
       visible: false,
+      captureType: "",
+      nodeType: "",
+      typeAllowed: true,
```
safchain: I would rather put it as false by default, because we can allow the user the select the type only if we know which nodes are selected.

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: we discussed to have an API to retrieve this list from common.go in another patch, so temporarily i defined here. if you want me this to move to api.js, i am ok with it.

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
safchain: right, but meanwhile I think it is better to have one definition

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: done

statics/js/components/capture-form.js
masco-enovance: sure Sylvain will do.

masco-enovance:
run selenium-tests

PR #569, opened 2017-11-10
graph : use Dijkstra shortestpath algorithm

nplanel:


lebauce:
It seems you don't need to add the GetNodesMap method to the backend but only to the file graph.go : the 3 implementations all do : get all the nodes and create a map of it

nplanel:
done

PR #580, opened 2017-11-14
ui: fix bandwidth display

safchain:


matrohon:
I just tested the patch. Now the text is changed when the threshold is bypassed. But if the last link change its text to "warning" style, other links that should be in "alert" style get their changed to "warning" style too.
👎

matrohon:
it works in my testbed 👍

PR #584, opened 2017-11-14
ui: fix toggling issue in metadata

masco-enovance:


statics/js/components/object-detail.js
```diff
@@ -54,8 +54,16 @@ Vue.component('object-detail', {
     collapsedByDefault: function(key) {
       if (key === "FDB" || key === "Neighbors") return true;
       return false;
+    },
+
+    getTarget: function(key) {
+      this.randomKey = key + Math.floor(Math.random() * 200);
```
safchain: what do you think about increment a value, random can still generate the same value.

masco-enovance: agree, will change!

masco-enovance: @safchain I am not able to do an increment a value, as soon as i add this line "this.$store.commit('increment');" in getTarget function and corresponding code in app.js. getTarget is called continuously and hanging.

safchain: @masco-enovance ok, I will try to figure out why

safchain: @masco-enovance if you rebase I think you can now use path + key as key like for the tranformer stuff, no more need to random nor to increment

safchain: @masco-enovance nope it was stupid :( but here something which seems to work https://da.gd/0AFbC

masco-enovance: if another array has the same key at the same index it is not working, in Neighbors, we have 'state' at index 4 and it is not working. so I have used the path also to generate a unique ID. Hope this will work for all.

Issue #578, opened 2017-11-13
UI: link doesn't change after bypassing bandwidth thresholds


matrohon:
The color of the link/text should change when the bandwidth bypass the thresholds set in the conf file :

```
  bandwidth_absolute_active: 1
  bandwidth_absolute_warning: 2
  bandwidth_absolute_alert: 100
```

but nothing change.
After some debugging, it seems that css classes "link-label-warning/alert" are not applied, even when required conditions are met.

matrohon:
this patch makes the bw text message to change :

https://github.com/Orange-OpenSource/skydive/commit/4aa6f9f4f02c61a69a02c81ba73bd584f8dd2679

but every bw texts gets changed when a threshold is bypassed, not only the one of the concerned link

matrohon:
since 75ee89d98cdd7114cc8deede6128e2714ee65456, the color of text and line gets changed when threshold are bypassed.

However, I still have some strange behaviors : 
- the color of the line between a fabric port and a host port is never changed;
- some line color stay red, even when the amount of data goes under the alarm threshold.

PR #588, opened 2017-11-15
ui: reset link color when bandwidth below threshold

safchain:
Bug: #578


matrohon:
👍 it works on my setup

PR #589, opened 2017-11-16
Api python

safchain:


lebauce:
nack because of the term "pythonic"

safchain:
@lebauce arf I've been unmasked :)

PR #606, opened 2017-11-22
ui: add preference pane

masco-enovance:


masco-enovance:
run selenium-tests

statics/js/components/topology.js
```diff
@@ -234,6 +238,7 @@ var TopologyComponent = {
     });
 
     this.setFilterFromConfig();
+    this.topologyFilterQuery();
```
masco-enovance: this call not applying the filter, not sure from where we need to call this function. needs @safchain blessings :)

safchain: @masco-enovance it does for me

masco-enovance: oh, cool. thanks sylvain!

safchain: It does for me, what do you mean by not called, you never pass through "mounted" or the function is called but the filter is not applied ?

safchain:
@masco-enovance regarding to the preferences itself I would allow the user to define:

- List of filter with two fields, name and filter instead of default. This list will be appended to the one coming from the config file.
- Changing the bandwidth threshold that are currently coming from the config file.

For the order of item in the right menu I would prefer preference, documentation then logout

For the panel itself I would align everything to the left.

Thanks

statics/js/components/topology.js
masco-enovance: yeah, I mean that the filter is not applied.

statics/js/components/topology.js
```diff
@@ -174,12 +174,16 @@ var TopologyComponent = {
   ',
 
   data: function() {
+    var filter = "";
+    if (typeof(Storage) !== "undefined") {
+      filter = localStorage.getItem("defaultFilter");
+    }
```
eonpatapon: Instead of using localStorage directly, we should look at vuex alternatives:
- https://github.com/robinvdvleuten/vuex-persistedstate
- https://github.com/crossjs/vuex-localstorage

So that we can have 2 vues stores, one persistent, one temporary (the one we already have).

In this case, if the topologyFilter is stored in a persistent vuex store, we can react on it in all components.

eonpatapon: In fact with https://github.com/robinvdvleuten/vuex-persistedstate it is possible to specify which keys from the store are persisted, so only one store is required.

masco-enovance:
run unit-tests

statics/js/components/preference.js
```diff
@@ -0,0 +1,98 @@
...
+  template: '\
+    <form class="form-preference" @submit.prevent="save">\
+      <h2> Preferences </h2>\
+      <div class="form-group">\
+        <label for="collapse-index">Collapse Index</label>\
+        <input id="collapse-index" type="number" class="form-control input-sm" v-model="collapseIndex" min="0" />\
+      </div>\
+      <div class="form-group">\
+        <div class="row">\
+          <label for="default-filter">Filters</label>\
```
safchain: the label is not aligned with the other labels

statics/js/components/preference.js
```diff
@@ -0,0 +1,98 @@
...
+          <button class="btn btn-primary btn-round-xs btn-xs" type="button" @click="addFilter">+</button>\
```
safchain: I would prefer to have the "+" button below the list of fields

statics/js/components/topology.js
safchain: I think we can start the current approach if masco is fine with it, then we will improve this.

statics/js/components/preference.js
```diff
@@ -0,0 +1,98 @@
...
+
+var PreferenceComponent = {
+
+  name: 'Preference',
+
+  mixins: [notificationMixin],
+
+  template: '\
+    <form class="form-preference" @submit.prevent="save">\
+      <h2> Preferences </h2>\
```
safchain: I would remove the title as we already know that we are editing the preferences, we clicked on the tab and the tab is selected.

statics/js/components/preference.js
```diff
@@ -0,0 +1,98 @@
...
+      <div class="form-group">\
+        <label for="collapse-index">Collapse Index</label>\
```
safchain: It could be nice to add another line below this one with explanation of the parameter or to add a "?" picto on the right of the label displaying a tooltip.

statics/js/components/topology.js
masco-enovance: I am thinking these values are no need to available for all the vue components and not a big fan of using more external libs unless it is really needed.

statics/css/skydive.css
```diff
@@ -533,3 +533,29 @@ ul.dropdown-menu .fa.pull-right {
...
+.filter-field {
+  margin-top: 5px;
+}
+
+.button-holder {
+  text-align: center;
+}
+
+.btn-round-xs{
+border-radius: 11px;
```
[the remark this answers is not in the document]
masco-enovance: done

statics/js/components/preference.js
masco-enovance: done

statics/js/components/preference.js
masco-enovance: added a help text

statics/js/components/preference.js
masco-enovance: done

statics/js/components/preference.js
masco-enovance: moved

statics/js/components/preference.js
```diff
@@ -0,0 +1,98 @@
...
+      collapseIndex: ci,
+      filters: filters,
+      bwThreshold: bwThreshold,
+    };
+  },
+
+  methods: {
+
+    save: function() {
+      if (this.bwThreshold !== "" && this.bwThreshold !== "absolute" && this.bwThreshold !== "relative") {
```
safchain: we need to handle all the BW configuration parameters

masco-enovance: done

statics/js/components/graph-layout.js
```diff
@@ -1094,13 +1100,37 @@ TopologyGraphLayout.prototype = {
                  'analyzer.bandwidth_absolute_alert']
      };
 
+    if (typeof(Storage) !== "undefind") {
```
safchain: typo: undefined

statics/js/components/graph-layout.js
```diff
@@ -1094,13 +1100,37 @@ TopologyGraphLayout.prototype = {
...
+      var cfgValues = {
+      absolute: [this.getLocalValue("bandwidthAbsoluteActive"),
+                 this.getLocalValue("bandwidthAbsoluteWarning"),
+                 this.getLocalValue("bandwidthAbsoluteAlert")],
+      relative: [this.getLocalValue("bandwidthRelativeActive"),
+                 this.getLocalValue("bandwidthRelativeWarning"),
+                 this.getLocalValue("bandwidthRelativeAlert")]
+      };
+    } else {
+      var cfgValues = {
```
safchain: can you move the declaration before the if so that we don't use var in if/else statement which is more readable

statics/js/components/preference.js
```diff
@@ -0,0 +1,133 @@
...
+
+var PreferenceComponent = {
+
+  name: 'Preference',
+
+  mixins: [notificationMixin],
+
+  template: '\
+    <form class="form-preference" @submit.prevent="save">\
+      <div class="form-group">\
```
safchain: maybe remove this one as not used yet or not working

statics/js/components/preference.js
```diff
@@ -0,0 +1,133 @@
...
+              <input class="form-control" v-model="filter.filter"/>\
+            </div>\
+          </div>\
+        </div>\
+        <button class="btn btn-primary btn-round-xs btn-xs" type="button" @click="addFilter">+</button>\
+      </div>\
+      <div class="form-group">\
+        <label for="bw-threshold">Bandwidth Threshold</label>\
+        <a><i class="fa fa-question help-text" aria-hidden="true" title="bandwidth threshold mode (relative/absolute)"></i></a>\
+        <input id="bw-threshold" type="text" class="form-control input-sm" v-model="bwThreshold"/>\
```
safchain: it would be better to have a list instead of having to enter the value

statics/js/components/preference.js
```diff
@@ -0,0 +1,133 @@
...
+  data: function() {
+    var filters = [{name:"", filter:""}];
+    
+    f = JSON.parse(localStorage.getItem("filters"));
+    if (f && f.length > 0) filters = f;
+
+    return {
+      collapseIndex: localStorage.collapseIndex,
+      filters: filters,
+      bwThreshold: localStorage.bandwidthThreshold,
```
safchain: We should return the default values from config file is provided

statics/js/components/preference.js
```diff
@@ -0,0 +1,133 @@
...
+      </div>\
+      <div class="form-group">\
+        <label for="bw-threshold">Bandwidth Threshold</label>\
+        <a><i class="fa fa-question help-text" aria-hidden="true" title="bandwidth threshold mode (relative/absolute)"></i></a>\
+        <input id="bw-threshold" type="text" class="form-control input-sm" v-model="bwThreshold"/>\
+      </div>\
+      <div class="form-group">\
+        <label for="bw-update-interval">Bandwidth Update Interval</label>\
+        <a><i class="fa fa-question help-text" aria-hidden="true" title="bandwidth udadate interval(in seconds)"></i></a>\
+        <input id="bw-update-interval" type="number" class="form-control input-sm" v-model="bwUpdatePeriod"/>\
```
safchain: you don't reset the timer and it will not be easy, I propose to remove this parameter for now

statics/js/components/graph-layout.js
```diff
@@ -1094,13 +1100,37 @@ TopologyGraphLayout.prototype = {
...
       .then(function(period, src, threshold) {
-        b.updatePeriod = period[0] * 1000; // in millisec
-        b.bandwidth_threshold = threshold[0];
+        if (up = self.getLocalValue("bandwidthUpdatePeriod")) {
+          b.updatePeriod = up * 1000; //in millisec
+        } else {
+          b.updatePeriod = period[0] * 1000;
+        }
+        if (localStorage.bandwidthThreshold) {
+          b.bandwidth_threshold = localStorage.bandwidthThreshold;
```
safchain: There is a bug in the master branch, it should be b.bandwidthThreshold, can you fix it in your patch

statics/js/components/preference.js
```diff
@@ -0,0 +1,133 @@
...
+      </div>\
+      <div class="form-group">\
+        <label for="bw-abs-active">Bandwidth Absolute Active</label>\
```
safchain: no need to have twice the parameters, one for both Absolute and Relative is enough. maybe you could show/hide fields according to the mode.

statics/js/components/preference.js
```diff
@@ -0,0 +1,133 @@
...
+        <input id="bw-abs-active" type="number" class="form-control input-sm" v-model="bwAbsActive" min="0"/>\
+      </div>\
+      <div class="form-group">\
+        <label for="bw-abs-warning">Bandwidth Absolute Warning</label>\
```
safchain: could you add a help note saying that the value is in kbps

statics/js/components/preference.js
masco-enovance: done, nice idea!

statics/js/components/preference.js
masco-enovance: removed

statics/js/components/preference.js
masco-enovance: done

statics/js/components/preference.js
masco-enovance: removed

statics/js/components/preference.js
masco-enovance: done

statics/js/components/graph-layout.js
masco-enovance: done

statics/js/components/graph-layout.js
masco-enovance: done

statics/js/components/graph-layout.js
masco-enovance: done

statics/js/components/preference.js
masco-enovance: if we pass a config value it will be changing the local value every time we go to preference pane.

statics/js/components/preference.js
```diff
@@ -0,0 +1,132 @@
...
+            </div>\
+          </div>\
+        </div>\
+        <button class="btn btn-primary btn-round-xs btn-xs" type="button" @click="addFilter">+</button>\
+      </div>\
+      <div class="form-group">\
+        <label for="bw-threshold">Bandwidth Threshold</label>\
+        <a><i class="fa fa-question help-text" aria-hidden="true" title="bandwidth threshold mode (relative/absolute)"></i></a>\
+        <select id="be-threshold" v-model="bwThreshold" class="form-control input-sm">\
+          <option value="">None</option>\
```
safchain: This can't be None it's either Absolute or Relative the default is Relative, but we should use the value coming from the config as default. If the storage is not filled with fill the Panel with the values from the API

masco-enovance: done. now if there is no value in local storage, it will take from the config file.

statics/js/components/preference.js
```diff
@@ -0,0 +1,142 @@
...
+        </div>\
+        <div class="form-group">\
+          <label for="bw-abs-alert">Bandwidth Absolute Alert</label>\
+          <a><i class="fa fa-question help-text" aria-hidden="true" title="Alert threshold in kbps"></i></a>\
+          <input id="bw-abs-alert" type="number" class="form-control input-sm" v-model="bwAbsAlert" min="0"/>\
+        </div>\
+      </div>\
+      <div v-if="bwThreshold == \'relative\'">\
+        <div class="form-group">\
+          <label for="bw-rel-active">Bandwidth Relative Active</label>\
```
safchain: could you add a help note for these fields too saying that the value is a ratio between 0-1 or something similar

statics/js/components/preference.js
```diff
@@ -0,0 +1,142 @@
...
+
+  template: '\
+    <form class="form-preference" @submit.prevent="save">\
+      <div class="form-group">\
+        <label for="default-filter">Filters</label>\
+        <a><i class="fa fa-question help-text" aria-hidden="true" title="filter and highlight gremlin queries, displayed in the left panel"></i></a>\
+        <div v-for="filter in filters">\
+          <div class="form-group">\
+            <div class="input-group">\
+              <label class="input-group-addon">Name: </label>\
```
safchain: We should add a "delete" button right to this field in order to allow the user to delete previously defined filter

masco-enovance: nice idea! added a button to delete.

statics/js/components/preference.js
masco-enovance: done

masco-enovance:
rerun compile-tests

PR #610, opened 2017-11-23
ui: bug fix for time icon click with no data

masco-enovance:
now if we click the time icon with no data in date and time fields
it is showing Live in bottom of the screen.

the functionality is changed to show the current time


masco-enovance:
run scale-tests

PR #612, opened 2017-11-24
common : factorize IPv4/v6 normalize IP/Address

nplanel:

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -224,9 +224,22 @@ func IPv6Supported() bool {
 	return true
 }
 
-// NormalizeIPForUrl returns a string normalized that can be used in URL. Brackets
-// will be used for IPV6 addresses.
-func NormalizeIPForUrl(ip net.IP) string {
+// NormalizeAddrForUrl returns a string normalized that can be used in URL. Brackets
+// will be used for IPV6 addresses. Accept net.IP or string as input.
+func NormalizeAddrForURL(addrI interface{}) string {
+	var ip net.IP
+	switch addrI.(type) {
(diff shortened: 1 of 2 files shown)


common/types.go
```diff
@@ -225,8 +225,20 @@ func IPv6Supported() bool {
 }
 
 // NormalizeIPForUrl returns a string normalized that can be used in URL. Brackets
-// will be used for IPV6 addresses.
-func NormalizeIPForUrl(ip net.IP) string {
+// will be used for IPV6 addresses. Accept net.IP of string as input.
```
masco-enovance: nit: s/of/or

analyzer/flow_client.go
```diff
@@ -110,7 +110,7 @@ func (c *FlowClientWebSocketConn) Close() error {
 // Connect to the WebSocket flow server
 func (c *FlowClientWebSocketConn) Connect() error {
 	authOptions := NewAnalyzerAuthenticationOpts()
-	authAddr := common.NormalizeAddrForURL(c.url.Hostname())
+	authAddr := common.NormalizeIPForUrl(c.url.Hostname())
```
safchain: Hostname can return a domain name and even if we resolve the address earlier in the call flow I would prefer to keep addr to support domain name.

common/types.go
nplanel: done

analyzer/flow_client.go
nplanel: done

nplanel:
run skydive-scale-tests

PR #632, opened 2017-11-29
test: removed code duplication in test.functionals.run target

hunchback:
TestBridgeOVS fails if br-test1 already exists as can be reproduced by running:

$ cd $GOPATH/src/github.com/skydive-projects/skydive
$ sudo ovs-vsctl add-br br-test1 || true
$  sudo -E tests/functionals -test.v -test.timeout 20m -test.run TestBridgeOVS$ -standalone -graph.backend elasticsearch -storage.backend elasticsearch
*FAIL*

After the fix running the same sequence you get:
*PASS*

### DIFF
--- a/api/config.go
+++ b/api/config.go
@@ -10,6 +10,7 @@ import (
 	"github.com/skydive-project/skydive/common"
 	"github.com/skydive-project/skydive/config"
 	shttp "github.com/skydive-project/skydive/http"
+	"github.com/skydive-project/skydive/logging"
 	"github.com/spf13/viper"
 )
 
@@ -25,7 +26,7 @@ func (c *configAPI) configGet(w http.ResponseWriter, r *auth.AuthenticatedReques
 	value := common.NormalizeValue(c.cfg.Get(key))
 	w.WriteHeader(http.StatusOK)
 	if err := json.NewEncoder(w).Encode(value); err != nil {
-		panic(err)
+		logging.GetLogger().Warningf("Error while writing response: %s", err)
 	}
 }
 
--- a/api/packet_injector.go
+++ b/api/packet_injector.go
@@ -30,7 +30,9 @@ import (
 	"strings"
 
 	"github.com/abbot/go-http-auth"
+
 	shttp "github.com/skydive-project/skydive/http"
+	"github.com/skydive-project/skydive/logging"
 	"github.com/skydive-project/skydive/packet_injector"
 	"github.com/skydive-project/skydive/topology"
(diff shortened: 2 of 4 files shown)


safchain:
A cleanup script is used for that

https://github.com/skydive-project/skydive/blob/master/scripts/ci/cleanup.sh

To clean up between two runs

lebauce:
The test creates and removes the bridge at the end.

If, for some reason, the test was "killed" before cleaning up, the "cleanup.sh" script that @safchain mentionned is in charge of cleaning up the environment.

If we wanted to make all the tests idempotent, that would involve quite a lot of work.

The second commit looks fine to me, so I suggest you rebase and skip the first one

safchain:
ok to test

hunchback:
@lebauce , @safchain

> The second commit looks fine to me, so I suggest you rebase and skip the first one

done 

> If we wanted to make all the tests idempotent, that would involve quite a lot of work.

Having each test being idempotent in itself will speedup a typical case of chasing a specific bug which requires re-run of the same test again and again (not always running to the end, thus potentially skipping cleanup). The cleanup.sh code takes 15 seconds to run, so it may be a little less convenient. In this respect even only the *important* tests idempotent  still has value.

safchain:
rerun tests

Makefile
```diff
@@ -81,11 +83,7 @@ test.functionals.compile: govendor genlocalfiles
 	${GOPATH}/bin/govendor test -tags "${TAGS} test" ${GOFLAGS} ${VERBOSE_FLAGS} -timeout ${TIMEOUT} -c -o tests/functionals ./tests/
 
 test.functionals.run:
-ifneq ($(VERBOSE_FLAGS),)
-	cd tests && sudo -E ./functionals -test.v -test.timeout ${TIMEOUT} ${ARGS}
-else
-	cd tests && sudo -E ./functionals -test.timeout ${TIMEOUT} ${ARGS}
-endif
+	sudo -E tests/functionals ${VERBOSE_TESTS_FLAGS} -test.timeout ${TIMEOUT} ${ARGS}
```
lebauce: cd tests && sudo -E functionals ${VERBOSE_TESTS_FLAGS} -test.timeout ${TIMEOUT} ${ARGS}

hunchback:
I have issued a new pull requests [638](https://github.com/skydive-project/skydive/pull/638) in which I revered back to using `cd tests && sudo -E ./functionals ...` as some of the tests run in `./functionals` depend in being run in the `tests` directory (and access some of tests sub-directories directly)

PR #640, opened 2017-11-30
build: Fix snort build when skydive directory is symlinked

nlewo:
When the root skydive directory is a symlink (to be in the GOPATH),
`make -C contrib/snort` moves to the resolved skydive directory
which can be out of the GOPATH. In this case, snort fails to build.


masco-enovance:
ok to test

nplanel:
it's not a good habit to do cd ...; make as the make environment is lost
What is exactly the error ?
Why skydive should be a symlink ?

nlewo:
@nplanel the error I get:
```
$ pwd
/home/lewo/.gocode/src/github.com/skydive-project/skydive

$ make install
...
...
make -C contrib/snort
make[1]: Entering directory '/home/lewo/repos/skydive/contrib/snort'
govendor build snortSkydive.go
Error: Package "/home/lewo/repos/skydive" not a go package or not in GOPATH.
make[1]: *** [Makefile:6: snortSkydive] Error 2
make[1]: Leaving directory '/home/lewo/repos/skydive/contrib/snort'
make: *** [Makefile:67: contribs] Error 2
```

`skydive` is a symlink because I have a directory (`~/repositories`) in my home with all repositories I use, and another directory for go (`~/.gocode`). So, I then symlink some repository from `~/repositories` to `~/.gocode/src/...`.

lebauce:
@nlewo You need to change the order of the symlink : the 'real' directory in your GOPATH and the symlink in your "repositories" folder

nlewo:
@lebauce this is a little bit tricky... and backup directories would then have to contain a `~/.gocode` folder. 
But it's not an important issue, so I close this PR.

nplanel:
[golang issue 15507](https://github.com/golang/go/issues/15507)
Rob Pike : "Symlinks are too problematic. It seems unwise to allow them."

PR #647, opened 2017-12-01
ui: add highlight circle instead of changing color for highlighted nodes

safchain:


safchain:
run cdd-overview

safchain:
run cdd-overview

masco-enovance:
after moving out from flows the highlighted nodes are not coming back to normal. not reproducing for the first time just go and come out on the flows for two-three times after that the nodes remain highlighted.

masco-enovance:
here is my screenshot https://pasteboard.co/GXGUvte.png

safchain:
rerun selenium-tests

PR #669, opened 2017-12-12
ui: add ssh support via Dede terminal

safchain:


safchain:

PR #678, opened 2017-12-16
K8s:  adding full metadata and some code improvements

hunchback:
to test look at a k8s cluster with pods and networkplocies

### DIFF
--- a/statics/js/components/graph-layout.js
+++ b/statics/js/components/graph-layout.js
@@ -1503,15 +1503,13 @@ TopologyGraphLayout.prototype = {
   },
 
   nodeImg: function(d) {
-    if (nodeImgMap[d.metadata.Type]) {
-      return nodeImgMap[d.metadata.Type];
-    }
-
-    return intfImg;
+    var t = d.metadata.Type || "default";
+    return nodeImgMap[t];
   },
 
   managerImg: function(d) {
-    return managerImgMap[d.metadata.Orchestrator || d.metadata.Manager];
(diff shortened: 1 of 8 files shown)


statics/js/components/graph-layout.js
```diff
@@ -1503,15 +1503,12 @@ TopologyGraphLayout.prototype = {
   },
 
   nodeImg: function(d) {
-    if (nodeImgMap[d.metadata.Type]) {
-      return nodeImgMap[d.metadata.Type];
-    }
-
-    return intfImg;
+    return getNodeImage(d.metadata.Type);
```
safchain: It is like this because it allows users to customize the WebUI with their own files thanks to the config file.

statics/js/components/layout.js
```diff
@@ -1,38 +1,39 @@
...
-  "switch": switchImg,
-  "ovsbridge": switchImg,
-  "netns": nsImg,
-  "veth": vethImg,
-  "bond": portImg,
-  "container": dockerImg,
-  "pod": podImg,
-  "networkpolicy": networkPolicyImg
+var getNodeImage = function(type) {
+  var nodeImgMap = {
```
safchain: As said before it is now impossible for a user to customize the UI without rewritten the whole function. The goal here was to allow user to add their own image to the map thanks to the config file.

hunchback:
@safchain - rather than exposing a function to covert label to image I reverted back to exposing a map, only that I setup the map 'automatically' via `setupFixedImages()` function call. This does prohibit adding additional images to the map later. (to enable extending the map via config file)

PR #681, opened 2017-12-20
K8s: add nodes via k8s probe if do not already exist

hunchback:
@lebauce - Added creation of a node item directly from k8s probe, given that this competes with the agent generated node, it will only add/update if an agent created node does not exist. To test you can fireup a k8s and view that even without an agent installed within the host (on which the pods are running) you still get a node element (double-clicking with expose the pods).

### DIFF
--- a/topology/probes/k8s/node.go
+++ b/topology/probes/k8s/node.go
@@ -1,0 +1,99 @@
+/*
+ * Copyright (C) 2017 Red Hat, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package k8s
+
+import (
+	"sync"
+
+	"github.com/skydive-project/skydive/topology/graph"
+
+	"k8s.io/api/core/v1"
+)
+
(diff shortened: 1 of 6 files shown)


topology/probes/k8s/node.go
```diff
@@ -0,0 +1,99 @@
...
+	defer c.graph.Unlock()
+
+	graphMeta := c.getMetadata(newNode)
+	hostIndexer := graph.NewMetadataIndexer(c.graph, graph.Metadata{"Type": "host"}, "Name")
+	hostNodes := hostIndexer.Get(newNode.GetName())
+	if len(hostNodes) == 0 {
+		c.graph.NewNode(graph.Identifier(newNode.GetUID()), graphMeta)
+	} else {
+		if val, ok := graphMeta["Manager"]; ok && val == "k8s" {
+			c.graph.SetMetadata(hostNodes[0], graphMeta)
```
lebauce: SetMetadata overrides all the existing metadata. So if the user added custom metadata through the configuration files, they will be overriden. You could either use AddMetadata, or get the host metadata, add the fields you want and use SetMetadata

lebauce:
Just a really quick review as I'm (supposed to be) on vacation :-)

I'm ok with creating for discovering Kubernetes nodes. We need to be careful though as it's a bit special type of nodes (I'm thinking about HostGraphDeleted messages when a 'host' disappears).

In the case where the K8s are running a Skydive agent but are discovered by K8S first, we need to make sure that the Metadata set by the 'k8s' probe are not overriden (same thing if the 'host' node is seen first)

topology/probes/k8s/node.go
hunchback: ok in next update I will use `AddMetadata()` to only update:
- K8s: which retains the raw metadata provided by k8s objects
- Manager: "k8s": so that we get he icon indicating this is tracked by k8s

hunchback:
@lebauce - the HostGraphDeleted message is a challenge and will delete pods, containers and of-course host (with our k8s metadata) for now when the agent initializes such a message we will loose any accumulated data/metadata - this is not ideal - but that said anything re-created after the delete will appear. To solve this we would need either to periodically re-create k8s elements from cached data, or track which agent (such as host-agent, analyzer, etc.) added any specific graph element, and thus restrict HostGraphDeleted to the elements added by the agent issuing the request ... but in any case this is orthogonal to the change and can be delayed for a future PR.

As to not overriding fields that the agent added, as explained I no longer blindly overwrite the node metadata in case it already exists.

hunchback:
In any case will *close* this PR and use a different PR to handle the changes.

PR #682, opened 2017-12-21
K8s: add k8s container and node (host) discovery

hunchback:
This is a 2nd phase for the PR adding this capability for node discovery via the k8s probe itself (without relaying on the host agent).

To test just connect to a k8s landscape scape

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -253,7 +253,9 @@ ui:
   # extra_assets: /usr/share/skydive/assets
 
 k8s:
-  # The KUBECONFIG yml file, if k8sctl is executed inside the cluster, you may use the empty config file - kube-config-empty.yml
+  # EXPERIMENTAL: k8s probe is still under development and should not be used on production systems
+  # The kube config file, if executed inside the cluster, you may use the empty
+  # file, kube-config-empty.yml
   # config_file: /etc/skydive/kube-cfg.yml
   subprobes:
   - networkpolicy
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -255,3 +255,8 @@ ui:
 k8s:
   # The KUBECONFIG yml file, if k8sctl is executed inside the cluster, you may use the empty config file - kube-config-empty.yml
   # config_file: /etc/skydive/kube-cfg.yml
+  subprobes:
+  - networkpolicy
+  - pod
+  - container
+  - node
--- a/topology/graph/graph.go
+++ b/topology/graph/graph.go
@@ -272,8 +272,7 @@ func GenID() Identifier {
 // GenIDNameBased helper generate a node Identifier
 func GenIDNameBased(namespace, name string) Identifier {
 	ns, _ := uuid.ParseHex(namespace)
-	n := []byte(name)
-	u, _ := uuid.NewV5(ns, n)
+	u, _ := uuid.NewV5(ns, []byte(name))
 
 	return Identifier(u.String())
 }
--- a/topology/graph/graph.go
+++ b/topology/graph/graph.go
@@ -701,6 +701,11 @@ func (g *Graph) DelMetadata(i interface{}, k string) bool {
 	return true
 }
 
+// SetField set metadata value based on dot key ("a.b.c.d" = "ok")
+func (m *Metadata) SetField(k string, v interface{}) bool {
+	return common.SetField(*m, k, v)
+}
+
 func (g *Graph) addMetadata(i interface{}, k string, v interface{}, t time.Time) bool {
 	var e *graphElement
 	ge := graphEvent{element: i}
@@ -718,7 +723,7 @@ func (g *Graph) addMetadata(i interface{}, k string, v interface{}, t time.Time)
 		return false
 	}
 
-	if !common.SetField(e.metadata, k, v) {
+	if !e.metadata.SetField(k, v) {
 		return false
 	}
 
@@ -740,7 +745,7 @@ func (g *Graph) AddMetadata(i interface{}, k string, v interface{}) bool {
 
 // AddMetadata in the current transaction
 func (t *MetadataTransaction) AddMetadata(k string, v interface{}) {
-	common.SetField(t.Metadata, k, v)
+	t.Metadata.SetField(k, v)
 }
 
 // Commit the current transaction to the graph
--- a/topology/graph/graph.go
+++ b/topology/graph/graph.go
@@ -269,6 +269,15 @@ func GenID() Identifier {
 	return Identifier(u.String())
 }
 
+// GenIDNameBased helper generate a node Identifier
+func GenIDNameBased(namespace, name string) Identifier {
+	ns, _ := uuid.ParseHex(namespace)
+	n := []byte(name)
+	u, _ := uuid.NewV5(ns, n)
+
+	return Identifier(u.String())
+}
+
 func (m *Metadata) String() string {
 	j, _ := json.Marshal(m)
 	return string(j)
--- a/topology/probes/k8s/container.go
+++ b/topology/probes/k8s/container.go
@@ -40,7 +40,7 @@ type containerCache struct {
 	graph.DefaultGraphListener
 	*kubeCache
 	graph            *graph.Graph
-	hostIndexer      *graph.MetadataIndexer
+	podIndexer       *graph.MetadataIndexer
 	containerIndexer *graph.MetadataIndexer
 }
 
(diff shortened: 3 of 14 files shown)


lebauce:
When the 'docker' probe is used, nodes of type 'container' are already created. They are added as child of the namespace node.

In case where the node do not run the Skydive agent, we have no "namespace" node. I'm not sure what to do but we should have a consistent way of representing things, whether we run the Skydive agent or not.

hunchback:
@lebauce - the docker will add container to the namespace node, but if the k8s detects a container is under a pod it will place the container under the pod.

hunchback:
@lebauce / @safchain  - please view this as yet another intermediate step in k8s probe implementation, which currently provides (compared to the current state):
- ability to view k8s landscape (including hosts and containers) even w/o an agent probe
- merge to metadata in the case of hosts, containers when agent has created these already

Known limitations:
- missing periodic refresh of all the k8s probe created graph elements (configurable with a default of 30s)
- we must mark a host or container created by k8s in a color indicating its' not an accessible element (e.g. such as ability to inject/capture traffic etc.) and block a-priori any attempt to run any skydive-agent related operations on these nodes.
- we must map a hosts, containers to the same TID regardless of if it was created by the agent or the analyzer/k8s probe
- we must have a clear separation of  data/metadata added by the agent and data/metadata added by the analyzer, we could either have separate graph-nodes or have a clear separation within the graph-node element ... 
- we must support views with only k8s; only agent; merged view in the UI
- the current HOST DELETE message sent to the analyzer will also remove the k8s probe data (for now its' lost until some event occurs), we must change so that HOST DELETE only deletes the data/metadata added by the requested client (such that if issued by the agent, only data/metadata added by the agent will be deleted by this request, and if k8s added data to the host view, these graph-elements will not be deleted)

safchain:
@hunchback Thanks, It's fine for me to have intermediate steps. 

For the limitations, we really need to discuss about what to do with nodes created in two places and if we want to support this kind of deployment as it as deep design impact. 

For the host delete stuff, playing with the host binding (host field) could help. 

Agree that we should mark nodes reported only from the controller/config side. Skydive was designed to not rely on controller/config and to validate what is really done (namespaces, etc), so need to distinguish both ways of getting the things.

That said I'll review it asap

etc/skydive.yml.default
```diff
@@ -253,5 +253,11 @@ ui:
   # extra_assets: /usr/share/skydive/assets
 
 k8s:
-  # The KUBECONFIG yml file, if k8sctl is executed inside the cluster, you may use the empty config file - kube-config-empty.yml
+  # The kube config file, if executed inside the cluster, you may use the empty
+  # file, kube-config-empty.yml
   # config_file: /etc/skydive/kube-cfg.yml
+  subprobes:
```
safchain: Maybe we should put something in order to say that is experimental.

topology/probes/k8s/container.go
```diff
@@ -0,0 +1,171 @@
...
+func (c *containerCache) newMetadata(pod *v1.Pod, container *v1.Container) graph.Metadata {
+	m := newMetadata("container", container.Name, container)
+	common.SetField(m, DockerNameField, container.Name)
+	common.SetField(m, DockerPodNamespaceField, pod.GetNamespace())
+	common.SetField(m, DockerPodNameField, pod.GetName())
+	return m
+}
+
+func makeContainerUID(pod *v1.Pod, containerName string) string {
+	return string(pod.GetUID()) + "-" + containerName
```
safchain: We should use a uuidv5 with pod UID and containerName instead of concatenating so that we will keep the UUID of graph element consistent in term of format.

topology/probes/k8s/container.go
```diff
@@ -0,0 +1,171 @@
...
+	c.graph.Lock()
+	defer c.graph.Unlock()
+
+	containerNode := c.graph.GetNode(graph.Identifier(uid))
+	if containerNode == nil {
+		logging.GetLogger().Infof("Creating container{%s}", container.Name)
+		containerNode = c.graph.NewNode(graph.Identifier(uid), c.newMetadata(pod, container))
+	} else {
+		logging.GetLogger().Infof("container{%s} already exists", container.Name)
+		addMetadata(c.graph, containerNode, container)
```
safchain: I'm sure we should copy all the container metadata directly since we are going to move more or less to something schema based, but I think that is acceptable for now.

topology/probes/k8s/graph.go
```diff
@@ -0,0 +1,42 @@
...
+		"Type":    typ,
+		"Probe":   "k8s",
+		"Manager": "k8s",
+		"Name":    name,
+		"K8s":     extra,
+	}
+}
+
+func addMetadata(g *graph.Graph, n *graph.Node, extra interface{}) {
+	g.AddMetadata(n, "Manager", "k8s")
```
safchain: It would be better to use a transaction otherwise it will generate 2 updates and 2 graph revisions

topology/probes/k8s/graph.go
```diff
@@ -0,0 +1,42 @@
...
+	g.AddMetadata(n, "K8s", extra)
```
safchain: Same here, not that we should put everything directly, but acceptable for now I guess

topology/probes/k8s/k8s.go
```diff
@@ -23,49 +23,87 @@
...
 	podCache           *podCache
 	networkPolicyCache *networkPolicyCache
+	nodeCache          *nodeCache
+	containerCache     *containerCache
 }
 
-func (k8s *probe) Start() {
-	k8s.networkPolicyCache.Start()
-	k8s.podCache.Start()
+type starter interface {
```
safchain: there is already a probe interface probe/bundle.go

topology/probes/k8s/k8s.go
```diff
@@ -23,49 +23,87 @@
...
+	Start()
+	Stop()
 }
 
-func (k8s *probe) Stop() {
-	k8s.networkPolicyCache.Stop()
-	k8s.podCache.Stop()
+func (probe *Probe) getSubProbes() []starter {
```
safchain: same here you could use the proble/bundle.go

etc/skydive.yml.default
hunchback: @safchain - ok have added a comment in the above configuration file

topology/probes/k8s/container.go
hunchback: ok, have added a new function graph.NewIDNameBased(namespace, name string) which is now used to create a well-formed UID.

topology/probes/k8s/graph.go
hunchback: ok, using transactions

topology/probes/k8s/graph.go
hunchback: yes, using transactions now

topology/probes/k8s/k8s.go
hunchback: Replaced map of subProbes with probe.ProbeBundle

topology/probes/k8s/k8s.go
hunchback: replaced with probe.Probe

hunchback:
@lebauce - I have added a set of commits addressing all the above comments, for easing review of newly added code and did not squash commits together, once reviewed I will do so (kindly indicate which you would like to merge together).

lebauce:
It would be nice to enable all or some subprobes by default:

```
diff --git a/config/config.go b/config/config.go
index c1af3b1a..221d0bdc 100644
--- a/config/config.go
+++ b/config/config.go
@@ -100,6 +100,9 @@ func init() {
 
        cfg.SetDefault("host_id", host)
 
+       cfg.SetDefault("k8s.config_file", "/etc/skydive/kube-cfg.yml")
+       cfg.SetDefault("k8s.subprobes", []string{"networkpolicy", "pod", "container", "node"})
+
        cfg.SetDefault("logging.backends", []string{"stderr"})
        cfg.SetDefault("logging.color", true)
        cfg.SetDefault("logging.encoder", "")
@@ -116,8 +119,6 @@ func init() {
        cfg.SetDefault("sflow.port_min", 6345)
        cfg.SetDefault("sflow.port_max", 6355)
 
-       cfg.SetDefault("k8s.config_file", "/etc/skydive/kube-cfg.yml")
-
        cfg.SetDefault("storage.elasticsearch.host", "127.0.0.1:9200")
        cfg.SetDefault("storage.elasticsearch.maxconns", 10)
        cfg.SetDefault("storage.elasticsearch.retry", 60)
```

hunchback:
```
With the elasticsearch backend, I get "Bulk Insertion" errors. That's because elasticsearch does not accept dots in the keys of maps. In my case, it happens for the key "alpha.kubernetes.io/provided-node-ip" in Node.metadata.annotations.

In the Docker probe, we do https://github.com/skydive-project/skydive/blob/master/topology/probes/docker.go#L116-L122
```

I don't really like the approach used for Docker:
(1) it alters the structure `m["a.b.c/d"] ="e"` is now m["a"]["b"]["c/d"] = "e"`
(2) it is non-generic and requires handling each potentially offending field specifically

I have altered `common.NormalizeValue()`:
(1) it will convert `m["a.b.c/d"] = "e"` to `m["a_b_c/d"] = "e"`
(2) it traverses the handed object and make corrections recursively

hunchback:
> If the agent is started agent on the node that runs kubelet, there will be 2 nodes of type "host". We discussed with @safchain and thought it may indeed be simpler to create dedicated nodes of type "node" (or "k8s-node", or something else) and attach the pods to it. And we could simply create links between the hosts and the nodes when both of them has showed up. This way, we don't have problem with merging metadata and we make a clean distinction between the logical view and the physical view. What do you think ?

@lebauce - as you proposed creating a dedicated graph.Node representing the k8s node element (of Type: "node"), such that `pods --owned--> node --owned--> host`

etc/skydive.yml.default
```diff
@@ -253,5 +253,12 @@ ui:
   # extra_assets: /usr/share/skydive/assets
 
 k8s:
-  # The KUBECONFIG yml file, if k8sctl is executed inside the cluster, you may use the empty config file - kube-config-empty.yml
+  # EXPERIMENTAL: k8s probe still under development and should be used in productive systems
```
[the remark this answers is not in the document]
hunchback: ok, handled

topology/graph/graph.go
```diff
@@ -269,6 +269,15 @@ func GenID() Identifier {
 	return Identifier(u.String())
 }
 
+// GenIDNameBased helper generate a node Identifier
+func GenIDNameBased(namespace, name string) Identifier {
+	ns, _ := uuid.ParseHex(namespace)
+	n := []byte(name)
```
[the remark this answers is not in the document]
hunchback: ok

topology/probes/k8s/container.go
```diff
@@ -0,0 +1,170 @@
...
+	m := graph.Metadata{"Type": "container"}
+	addNotNullFilter(m, DockerPodNamespaceField)
+	addNotNullFilter(m, DockerPodNameField)
+
+	return graph.NewMetadataIndexer(g, m, DockerPodNamespaceField, DockerPodNameField)
+}
+
+func (c *containerCache) newMetadata(pod *v1.Pod, container *v1.Container) graph.Metadata {
+	m := newMetadata("container", container.Name, container)
+	common.SetField(m, DockerNameField, container.Name)
```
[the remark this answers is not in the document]
hunchback: ok - added `SetField(k, v)`

topology/probes/k8s/graph.go
```diff
@@ -0,0 +1,44 @@
...
+		"Probe":   "k8s",
+		"Manager": "k8s",
+		"Name":    name,
+		"K8s":     extra,
+	}
+}
+
+func addMetadata(g *graph.Graph, n *graph.Node, extra interface{}) {
+	tr := g.StartMetadataTransaction(n)
+	defer tr.Commit()
```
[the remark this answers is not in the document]
hunchback: ok

topology/probes/k8s/pod.go
```diff
@@ -41,100 +40,119 @@ type podCache struct {
...
-func (p *podCache) getMetadata(pod *api.Pod) graph.Metadata {
-	return graph.Metadata{
-		"Type":       "pod",
-		"Manager":    "k8s",
-		"Name":       pod.GetName(),
-		"UID":        pod.GetUID(),
-		"ObjectMeta": pod.ObjectMeta,
-		"Spec":       pod.Spec,
-	}
+func newPodIndex(g *graph.Graph, by string) *graph.MetadataIndexer {
```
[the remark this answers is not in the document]
hunchback: ok, renamed to `newPodIndexer`

hunchback:
> It would be nice to enable all or some subprobes by default:

added

hunchback:
I have addressed all above comments in the following commits:

```
Applying: build: fix dependency on 'dlv' for 'make debug'
Applying: k8s: normalized keys in 'K8s' to ES friendly form
Applying: k8s: create a dedicate k8s:node grpah.Node
Applying: k8s: fixed configuration *EXPERIMENTAL* message
Applying: k8s: minor fix in graph.GenIDNameBased()
Applying: k8s: added graph.Metadata.SetField(k, v)
Applying: k8s: removed defer in k8s.addMetadata()
Applying: k8s: replaced k8s.newPodIndex() -> k8s.newPodIndexer()
Applying: k8s: added default configuration values
```

Once reviewed I can squash these commits with existing ones

lebauce:
@hunchback Regarding NormalizeValue, I agree it's not optimal to alter the structure. What you suggest is what we did when we first faced this bug. It caused other problems, in gremlin parsing for instance (g.V().HasKey('Labels.My.Label.with.a.dot') should be written g.V().HasKey('Labels["My.Label.with.a.dot"]). So I suggest you replace dots with underscores in the "k8s" probe only, and not in common/types.go

Otherwise it looks good to me. With the NormalizeValue change, I'll happily merge it.

Maybe you can squash the following commits into one commit:

k8s: minor cleanup of configuration
k8s: add a comment that k8s probe is experimental
k8s: added default configuration values
k8s: fixed configuration *EXPERIMENTAL* message

hunchback:
> @hunchback Regarding NormalizeValue, I agree it's not optimal to alter the structure. What you suggest is what we did when we first faced this bug. It caused other problems, in gremlin parsing for instance (g.V().HasKey('Labels.My.Label.with.a.dot') should be written g.V().HasKey('Labels["My.Label.with.a.dot"]). So I suggest you replace dots with underscores in the "k8s" probe only, and not in common/types.go

@lebauce - I have changed `common.NormalizeValue()` to split keys by using `SetField()` - so all including k8s probe will use the dot-notation not the under_score_delimited, in a separate PR will can then remove the specific Docker code (now redundant).

> k8s: minor cleanup of configuration
> k8s: add a comment that k8s probe is experimental
> k8s: added default configuration values
> k8s: fixed configuration EXPERIMENTAL message

Have been merged into: 
- k8s: config cleanup and default values for probe

vendor/vendor.json
```diff
@@ -1935,6 +1941,10 @@
 			"path": "gopkg.in/yaml.v2",
 			"revision": "bef53efd0c76e49e6de55ead051f886bea7e9420"
 		},
+		{
```
lebauce: nit: This can be removed

hunchback: Hi @lebauce when removing it I get the below, but regardless, I would be happy to address (if needed) in a separate PR, as I need the k8s in the community version to support our internal integration efforts ....

```
make[1]: Entering directory '/home/vagrant/projects/gopath/src/github.com/skydive-project/skydive/contrib/snort'
govendor build snortSkydive.go
../../vendor/github.com/spf13/viper/util.go:29:2: cannot find package "gopkg.in/yaml.v2" in any of:
        /home/vagrant/projects/gopath/src/github.com/skydive-project/skydive/vendor/gopkg.in/yaml.v2 (vendor tree)
        /usr/local/go/src/gopkg.in/yaml.v2 (from $GOROOT)
        /home/vagrant/projects/gopath/src/gopkg.in/yaml.v2 (from $GOPATH)
Error: exit status 1
Makefile:6: recipe for target 'snortSkydive' failed
```

PR #684, opened 2017-12-22
ui: highlight only capture node on flow mouse over

safchain:
It allows to detect whether a packet has reached
A or B. Otherwise they are always highlighted
even if the packet didn't not reach one endpoint.

Instead this add a link on the right side of the
ANddeTID, BNodeTID to highlights endpoints.

### DIFF
--- a/statics/js/components/flow-table.js
+++ b/statics/js/components/flow-table.js
@@ -349,6 +349,7 @@ Vue.component('flow-table', {
       autoRefresh: false,
       showDetail: {},
       highlightMode: 'TrackingID',
+      currRelatedNodeHighlighted: '',
       filters: {},
       fields: [
         {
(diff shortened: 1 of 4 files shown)


statics/js/components/flow-table.js
```diff
@@ -615,16 +616,37 @@ Vue.component('flow-table', {
     flowDetailLinks: function(flow) {
       var self = this;
 
+      var links = {};
+
+      if (flow.ANodeTID !== "*") {
+        links.ANodeTID = {
+          "class": "indicator glyphicon glyphicon-link raw-packet-link",
+          "onClick": function() {},
+          "onMouseOver": function(){ self.highlightRelatedTID(flow.ANodeTID); },
```
masco-enovance: this node is already highlighted (https://github.com/skydive-project/skydive/blob/master/statics/js/components/flow-table.js#L301) by this line. after mouse over and out, it is losing the highlight.
can we remove the highlight of the detailed row of flow?

safchain:
rerun selenium-tests

statics/js/components/flow-table.js
safchain: Done now the already highlighted node stay highlighted even on mouse out

PR #686, opened 2017-12-24
elastic: make sure the system doesn't get stuck due to overloaded ES DB

vperelmax:
We add in the case of topology DB updates currently for ES only
channel that will make bulk updates to storage and clear the update buffer
and clear the buffer in case and we reach some watermark. Reaching it will mean
that ES doesn't stand the frequency of updates.


topology/graph/elasticsearch.go
```diff
@@ -509,15 +507,62 @@ func (b *ElasticSearchBackend) WithContext(graph *Graph, context GraphContext) (
...
+		for {
+			select {
+			case <-b.quit:
+				return
+			case <-dlTimer.C:
+				b.runStorageCmds(cmdBuffer)
+				cmdBuffer = cmdBuffer[:0]
+			case cmd := <-b.ch:
+				cmdBuffer = append(cmdBuffer, cmd)
+				if len(cmdBuffer) >= b.bulkInsertMax {
```
safchain: I didn't check the whole patch yet, one remark however. We can skip topology messages otherwise we will be in inconsistent state. I think we can increase the buffer but if the datastore is not able to handle the among of data even with a bigger buffer that a other issue that can be solved by increasing the delay between updates. Updates are usually due to metric, so netlink metric update delay can be increased. We are currently doing benchmarks and clearly the JSON serialization is the issue, we are going to move to protobuf. Maybe ES can be tweaked as well.

vperelmax: hi Sylvain,
I talked with eran about this issue. The main problem is that we can increase buffer sizes and timeouts, which probably for many cases will be the solution.
However, what happens if still ES DB is too slow and we don't stand in the frequency of updates?
Currently in such case the user will se in the WebUI just nothing, things become non-responding and the system stops in such state. Of cause, making DB inconsistent is also not the best solution, but we think we should discuss how as gracefully as possible to resolve such case.
Actually I did the change for topology only, but the problem might occur with flows storage as well (despit the code that is written in flow_server).
But, since we did see such issue related with topology storage, I think let's start discussing it there.
There are a few possible options, tell me if any of them sounds better than current situation:
1) If we reach some water mark - through everything from the buffer for a while and then invoke "syncAll" - to get the whole info from all the agents and sync the DB
2) If we reach water mark - notify about it all the listeners (user should receive such notification via the WebUI) and change mode to work with memory only as a backend
3) Notify users about the issue with DB and stop analyzer (or restart it)
4) Just empty the buffer - print error logs about number of lost messages and continue from this state (as I have implemented in this patch)

Anyway, to me each one of the listed solutions seems better choice than just halting the system without notifying the user and showing frozen UI...

Please, tell me what do you think about the issue.
Thanks!

safchain: I do agree we need at least notify the user that something is going wrong. Anything that will generate messages link re-sync will not help. I think that the analyzer could publish through the state API the current buffer fill rate in order to have the agents consuming this information and to adapt their own update rate ?

vperelmax:
Close the issue since we fix it with buffer overflow management for flows in the meantime.

PR #689, opened 2017-12-29
ui: allow user to provide MAC address

masco-enovance:


statics/js/components/inject-form.js
```diff
@@ -194,6 +212,12 @@ Vue.component('inject-form', {
       }
     },
 
+    getMAC: function(node) {
+      md = node.metadata;
```
safchain: var is missing, that said return node.metadata.MAC || "" should work

statics/js/components/inject-form.js
```diff
@@ -127,15 +137,21 @@ Vue.component('inject-form', {
...
     error: function() {
       if (!this.node1 || !this.node2) {
-          return "Source and destination interfaces must be selected";
-      } else if (this.srcIP == "" && this.dstIP == "") {
-          return "Source and Destination IPs need to be given by user";
-      } else if (this.srcIP == "") {
-          return "Source IP need to be given by user";
-      } else if (this.dstIP == "") {
-          return "Destination IP need to be given by user";
+        return "Source and destination interfaces must be selected";
```
safchain: I think all the "else if" are useless as we "return"

masco-enovance: it is used when we select an interface which doesn't have IP/MAC and the values are not given by a user too.

safchain:
rerun unit-tests

PR #696, opened 2018-01-04
ui: setting fixed position for nodes

vperelmax:
For testing add the following rows to the .yml fabric section:

- TOR1[Name=tor1] -> [color=red] TOR1_PORT1[Name=port1, MTU=1500, X=400, Y=428]

Once X and Y are set as metadata of the node, this patch will pin the node
in the selected location

### DIFF
--- a/statics/js/components/graph-layout.js
+++ b/statics/js/components/graph-layout.js
@@ -1369,6 +1369,27 @@ TopologyGraphLayout.prototype = {
     this.simulation.nodes(nodes);
     this.simulation.force("link").links(links);
     this.simulation.alpha(1).restart();
+    this.reArrangeGraphNodes();
+  },
+
+  setFixedPosition: function(node) {
+    if(node.metadata.X && node.metadata.Y) {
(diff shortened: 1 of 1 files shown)


statics/js/components/graph-layout.js
```diff
@@ -1369,6 +1369,27 @@ TopologyGraphLayout.prototype = {
...
     this.simulation.force("link").links(links);
     this.simulation.alpha(1).restart();
+    this.reArrangeGraphNodes();
+  },
+
+  setFixedPosition: function(node) {
+    if(node.metadata.X && node.metadata.Y) {
+        var theNode = this.nodes[node.id];
+        if(theNode == null) return;
+        theNode.x = parseInt(node.metadata.X);
```
safchain: I would prefer to not put "only UI specific" metadata, I would rather add something in the ui config section in order to add these kind of rules, something like

<uuid>:
  x: 100
  y: 200

Another solution could be to allow the user the change or customize the placement function by using the external JS loading mechanism.

PR #697, opened 2018-01-04
opencontrail: close col only if it was loaded

eonpatapon:
When collection.LoadCollection returns an error we should not call
Close() as col will refer to a nil pointer.

Rewrited the loop to make it more simple to read.

### DIFF
--- a/topology/probes/opencontrail.go
+++ b/topology/probes/opencontrail.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/skydive-project/skydive/common"
 	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/logging"
 	"github.com/skydive-project/skydive/topology"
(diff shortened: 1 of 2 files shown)


topology/probes/opencontrail.go
```diff
@@ -100,27 +100,30 @@ func (mapper *OpenContrailProbe) retrieveMetadata(metadata graph.Metadata, itf c
...
 	for i := 0; i < try; i++ {
-		col, e := collection.LoadCollection(descriptions.Interface(), []string{fmt.Sprintf("%s:%d", host, port)})
-		err = e
-		if e == nil {
-			itf, e := col.SearchStrictUnique(name)
-			err = e
-			if e == nil {
-				return col, itf, e
-			}
+		if i > 0 {
```
safchain: nit: you could use common.Retry

eonpatapon: common.Retry expect a function that returns only an error, I don't see how I can use it here

safchain: That is not important ;) It seems to me you are implementing the exact same pattern. That could be something like this

```
fnc := func() error {
  col, err = collection.LoadCollection(descriptions.Interface(), []string{fmt.Sprintf("%s:%d", host, port)})
  if err != nil {
    return err
  }

  elem, err = col.SearchStrictUnique(name)
  if err != nil {
    col.Close()
    return err
  }

  return nil
}

common.Retry(fnc, 3, 500*timeMillisecond)
```

safchain:
Some dependencies are missing

eonpatapon:

PR #700, opened 2018-01-08
build: allow client to build on OSX

esnible:
I use OSX.  Skydive's `make install` didn't succeed on OSX.

After these changes I am able to run `skydive client query` queries and `skydive client status` etc.

Three changes:
1. OSX uses BSD-style _sed_ so I added argument to sed invocations in Makefile and removed newline and tab escapes that are not necessary.
2. Switched to newer version of github.com/vishvananda/netns to get a fix for non-Linux
3. Added files with unimplemented or failing versions of methods that had Linux dependencies.

### DIFF
--- a/topology/probes/netlink_unspecified.go
+++ b/topology/probes/netlink_unspecified.go
@@ -1,0 +1,62 @@
+// +build !linux
+
+package probes
+
+import (
+	"errors"
+	"net"
+	"sync"
+
+	"github.com/skydive-project/skydive/topology/graph"
+)
+
+var (
+	ErrNotImplemented = errors.New("not implemented")
+)
+
+// NetNsNetLinkProbe describes a topology probe based on netlink in a network namespace
+type NetNsNetLinkProbe struct {
+	sync.RWMutex
+	Graph  *graph.Graph
+	Root   *graph.Node
+	NsPath string
+}
+
+// NetLinkProbe describes a list NetLink NameSpace probe to enhance the graph
+type NetLinkProbe struct {
+	sync.RWMutex
+	Graph *graph.Graph
+}
+
+// RouteTable describes a list of Routes
+type RoutingTable struct {
+	Id     int
+	Src    net.IP `json:"Src,omitempty"`
+	Routes []Route
+}
+
+// Route describes a route
+type Route struct {
+	Prefix   string    `json:"Prefix,omitempty"`
+	Nexthops []NextHop `json:"Nexthops,omitempty"`
+}
+
+// NextHop describes a next hop
+type NextHop struct {
+	Priority int    `json:"Priority,omitempty"`
+	Ip       net.IP `json:"Ip,omitempty"`
+}
+
+func NewNetLinkProbe(g *graph.Graph) (*NetLinkProbe, error) {
+	return nil, ErrNotImplemented
+}
+
+func (u *NetLinkProbe) Start() {
+}
+
+func (u *NetLinkProbe) Stop() {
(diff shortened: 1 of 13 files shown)


safchain:
ok to test

topology/probes/netlink_unspecified.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/skydive-project/skydive/topology/graph"
+)
+
+var (
+	ErrNotImplemented = errors.New("not implemented")
+)
+
+// NetNsNetLinkProbe describes a topology probe based on netlink in a network namespace
+type NetNsNetLinkProbe struct {
```
safchain: It would be better to have common files like netlink.go and specific files for unix and unspecified like netlink_unix.go and netlink_unspecified.go in order to avoid having structs duplicated. Same thing for the other files.

esnible: This makes sense for the constants in options.go.  For the others, it makes sense to rename the files with _linux.go.  (The structs will mostly still be duplicated, as these structs all contain private members that depend on things that are not present on OSX, or implemented in a different way).

safchain:
rerun selenium-tests

topology/probes/netlink_unspecified.go
safchain: For the selenium test nothing related to your patch, I just restarted it

safchain:
add to whitelist

topology/probes/netlink_unspecified.go
safchain: We think that if the goal is to be able to compile only the client part of Skydive it would be better to isolate the client. Otherwise you will get a binary with an agent and an analyzer. Moreover we would prefer to be able to cross compile the client so that we will be able to test it otherwise it is pretty sure that at some point we will break the MacOs support.

@lebauce has a draft which modifiers skydive that way. Can you please test it ?

esnible: https://github.com/lebauce/skydive is old and didn't build for me.

I am modifying the PR so that the binary does not include agent and analyzer on OSX.  This change reduced the OSX _skydive_ binary from 78MB to 62MB.

This PR does three things.  1. change Makefile _sed_ flags which helps with OSX (and maybe BSD.)  Brings in a newer version of vishvananda/netns which is needed anyway.  3. Makes a bunch of methods return "unimplemented".  (Maybe someone with more networking skill will implement for these OSX or use them to get started for another OS.)

I have not cross-compiled a Go program with "// build" comments but suspect it will work as expected.  You should have no trouble building OSX binaries on Linux with these changes, but without them you won't be able to cross-compile an OSX client binary on Linux.

It is possible in the future someone might break skydive into a client, an agent/analyzer, and a common library.  That is beyond my skill level.

esnible:
rerun selenium-tests

esnible:
rerun selenium-tests

esnible:
rerun selenium-tests

lebauce:
@esnible Thanks a lot for this work. An important point for us is to be able to run CI jobs for this so that we don't break Windows and Mac support as we only use Linux (Free Software for the win :-) )

I submitted a PR (https://github.com/skydive-project/skydive/pull/711) on top of your work that adds a new binary `skydive-cli` that is equivalent to `skydive client` and also allows building both agent and analyzers on other platforms. If it's ok for you, I'll close this PR and we'll use the new one.

esnible:
Yes, let's close this and use #711

PR #710, opened 2018-01-11
opencontrail: correctly return error from getInterfaceFromIntrospect

eonpatapon:


eonpatapon:
run skydive-unit-tests

PR #712, opened 2018-01-12
opencontrail: get the err from Retry function

nlewo:
Otherwise, this can lead to empty data structure without any error.


nlewo:
cc @eonpatapon

eonpatapon:
I already pushed this fix

safchain:
I close it because already merged

PR #713, opened 2018-01-12
k8s: fallback in case k8s.config_file is missing

hunchback:
If k8s.config_file (either defined or default value /etc/skydive/kubeconfig) does not exist then resolve k8s configuration as the `kubectl` does: first check `$KUBECONFIG`, next check `$HOME/.kube/config` finally resolve to empty definition (which works when running from within a k8s cluster).

### DIFF
--- a/topology/probes/k8s/client.go
+++ b/topology/probes/k8s/client.go
@@ -46,7 +46,15 @@ func newKubeClient() (*kubeClient, error) {
 	kubeconfig := config.GetConfig().GetString("k8s.config_file")
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


topology/probes/k8s/client.go
```diff
@@ -45,8 +45,16 @@ func (c *kubeClient) getCacheFor(restClient rest.Interface, objType runtime.Obje
 func newKubeClient() (*kubeClient, error) {
 	kubeconfig := config.GetConfig().GetString("k8s.config_file")
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to load Kubernetes config %s: %s", kubeconfig, err.Error())
+	if err == nil {
```
safchain: err != no ?

hunchback: correct - fixed

safchain:

PR #714, opened 2018-01-12
graph: indexer, do not index if key doesn't exist

safchain:


safchain:

PR #717, opened 2018-01-14
HTTP: user defined "Cookie" and extended logging for client(s)

hunchback:
Several fixes affecting HTTP client (both REST and WS), which can be tested by using:

In skydive.yml

```
agent:
  http:
    debug: true
    cookie: <value>
```
And then one can run the following commands:

```
skydive client status -c skydive.yml
skydive agent -c skydive.yml
```

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -107,6 +107,10 @@ agent:
   http:
     # log the HTTP client request and response (to log level DEBUG)
     # debug: false
+    # define the Cookie HTTP Request Header
+    # cookie: 
+    #   <name1>: <value1>
+    #   <name2>: <value2>
   topology:
     # Probes used to capture topology informations like interfaces,
     # bridges, namespaces, etc...
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -104,6 +104,9 @@ agent:
   # Not required, but can be used to allow virtual hosting
   # X509_servername: domain.com
   #
+  http:
+    # log the HTTP client request and response (to log level DEBUG)
+    # debug: false
   topology:
     # Probes used to capture topology informations like interfaces,
     # bridges, namespaces, etc...
--- a/http/auth.go
+++ b/http/auth.go
@@ -23,6 +23,7 @@
 package http
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"net/http"
@@ -54,10 +55,38 @@ func (c *AuthenticationClient) Authenticated() bool {
 	return c.authenticated
 }
 
-func (c *AuthenticationClient) SetHeaders(headers http.Header) {
-	if c.authenticated && c.AuthToken != "" {
-		headers.Set("Cookie", c.Cookie().String())
+func setCookies(headers *http.Header, c *AuthenticationClient) {
+	var b bytes.Buffer
+	for _, cookie := range cookies(c) {
+		b.WriteString(cookie.String())
+		b.WriteString("; ")
 	}
+	headers.Set("Cookie", b.String())
+}
+
+func cookies(c *AuthenticationClient) []*http.Cookie {
+	var cookies []*http.Cookie
+	cookies = append(cookies, configCookies()...)
+	if c != nil {
+		cookies = append(cookies, c.AuthCookies()...)
+	}
+	return cookies
+}
+
+func configCookies() []*http.Cookie {
+	var cookies []*http.Cookie
+	for name, value := range config.GetConfig().GetStringMapString("agent.http.cookie") {
+		cookies = append(cookies, &http.Cookie{Name: name, Value: value})
+	}
+	return cookies
+}
+
+func (c *AuthenticationClient) AuthCookies() []*http.Cookie {
+	var cookies []*http.Cookie
+	if c.authenticated && c.AuthToken != "" {
+		cookies = append(cookies, c.Cookie())
+	}
+	return cookies
 }
 
 func (c *AuthenticationClient) Cookie() *http.Cookie {
--- a/http/client.go
+++ b/http/client.go
@@ -30,10 +30,12 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/http/httputil"
 	"net/url"
 
 	"github.com/skydive-project/skydive/common"
 	"github.com/skydive-project/skydive/config"
+	"github.com/skydive-project/skydive/logging"
 )
 
 type RestClient struct {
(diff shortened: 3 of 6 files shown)


http/client.go
```diff
@@ -110,6 +110,9 @@ func (c *RestClient) Request(method, path string, body io.Reader, header http.He
 	switch resp.Header.Get("Content-Encoding") {
 	case "gzip":
 		resp.Body, err = gzip.NewReader(resp.Body)
+		resp.Header.Del("Content-Encoding")
```
safchain: Why removing something that was present in the response and used ? it seems to me that it will not affect the response at this point of the call path.

http/client.go
```diff
@@ -95,8 +95,12 @@ func (c *RestClient) Request(method, path string, body io.Reader, header http.He
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Add("Accept-Encoding", "gzip")
 
-	cookie := http.Cookie{Name: "authtok", Value: c.authClient.AuthToken}
-	req.Header.Set("Cookie", cookie.String())
+	cookieHdr := config.GetConfig().GetString("agent.http.cookie")
+	if cookieHdr == "" {
+		cookie := http.Cookie{Name: "authtok", Value: c.authClient.AuthToken}
```
safchain: In case of authentication Skydive uses here a cookie to keep track of the token, we shouldn't override it.

http/tls.go
```diff
@@ -36,6 +37,23 @@ func setTLSHeader(w http.ResponseWriter, r *http.Request) {
...
 
+func getTLSConfig(setupRootCA bool) *tls.Config {
+	certPEM := config.GetConfig().GetString("agent.X509_cert")
+	keyPEM := config.GetConfig().GetString("agent.X509_key")
+	var tlsConfig *tls.Config = nil
+	if certPEM != "" && keyPEM != "" {
+		tlsConfig = common.SetupTLSClientConfig(certPEM, keyPEM)
+		if setupRootCA {
+			analyzerCertPEM := config.GetConfig().GetString("analyzer.X509_cert")
+			if analyzerCertPEM != "" {
```
safchain: You changed the original behavior as previously if this field was not set the client was exiting. Any reason for changing this ?

etc/skydive.yml.default
```diff
@@ -262,3 +267,5 @@ k8s:
   - pod
   - container
   - node
+
```
safchain: useless added blank lines

hunchback: ok, removed

http/client.go
hunchback: ok I found a cleaner why to achieve the same behavior as explain here (https://golang.org/pkg/net/http/):

```
/ Uncompressed reports whether the response was sent compressed but
        // was decompressed by the http package. When true, reading from
        // Body yields the uncompressed content instead of the compressed
        // content actually set from the server, ContentLength is set to -1,
        // and the "Content-Length" and "Content-Encoding" fields are deleted
        // from the responseHeader. To get the original response from
        // the server, set Transport.DisableCompression to true.
        Uncompressed bool
```

I set `resp.Uncompressed = true` now

http/client.go
hunchback: I now combine all Cookie together - so I don't remove `authtok` is it exists

http/tls.go
hunchback: I have removed this condition ans will call setup of CA regardless of the values of "analyzer.X509_cert".

hunchback:
Why did test fail?

```
4:12:57.451 --- FAIL: TestSubscription (5.02s)
14:12:57.451 	wsserver_test.go:142: Server should have received 2 OnMessage event: 0
14:12:57.451 FAIL
14:12:57.451 FAIL	github.com/skydive-project/skydive/http	6.070s
14:12:57.451 ?   	github.com/skydive-project/skydive/logging	[no test files]
14:12:57.451 === RUN   TestBridgeAdded
```

http/auth.go
```diff
@@ -54,10 +55,38 @@ func (c *AuthenticationClient) Authenticated() bool {
...
+	var b bytes.Buffer
+	for _, cookie := range Cookies(c) {
+		b.WriteString(cookie.String())
+		b.WriteString("; ")
+	}
+	headers.Set("Cookie", b.String())
+}
+
+func Cookies(c *AuthenticationClient) []*http.Cookie {
+	cookies := make([]*http.Cookie, 0, 0)
```
safchain: var cookies []*http.Cookie will to the same

http/auth.go
```diff
@@ -54,10 +55,38 @@ func (c *AuthenticationClient) Authenticated() bool {
...
+func SetCookies(headers *http.Header, c *AuthenticationClient) {
+	var b bytes.Buffer
+	for _, cookie := range Cookies(c) {
+		b.WriteString(cookie.String())
+		b.WriteString("; ")
+	}
+	headers.Set("Cookie", b.String())
+}
+
+func Cookies(c *AuthenticationClient) []*http.Cookie {
```
safchain: no need to export

http/auth.go
```diff
@@ -54,10 +55,38 @@ func (c *AuthenticationClient) Authenticated() bool {
...
+	cookies := make([]*http.Cookie, 0, 0)
+	cookies = append(cookies, ConfigCookies()...)
+	if c != nil {
+		cookies = append(cookies, c.AuthCookies()...)
+	}
+	return cookies
+}
+
+func ConfigCookies() []*http.Cookie {
```
safchain: same here no need to export

etc/skydive.yml.default
```diff
@@ -104,6 +104,11 @@ agent:
   # Not required, but can be used to allow virtual hosting
   # X509_servername: domain.com
   #
+  http:
+    # log the HTTP client request and response (to log level DEBUG)
+    # debug: false
+    # define the Cookie HTTP Request Header
+    # cookie: { "<name1>": "<value1>", "<name2>": "<value2>" }
```
safchain: I would prefer to keep the file using yaml syntax only so

```
cookie:
  name1: value1
  name2: value2
```

http/auth.go
hunchback: changed to  `cookies`

http/auth.go
hunchback: ok

http/auth.go
hunchback: changed to `configCookies`

etc/skydive.yml.default
hunchback: ok, have changed the example in the skydive.yml.default (no code change needed)

hunchback:
@safchain - ok new version addressing comments

safchain:

safchain:
rerun scale-tests

safchain:
rerun tests

safchain:

PR #720, opened 2018-01-16
graph: introduce MetadataMatcher to address both raw Metadata and Fil…

safchain:
…ters


safchain:

PR #648, opened 2017-12-01
ui config cleanup and add default highlight/filter config parameter

safchain:

PR #722, opened 2018-01-17
etcd: re-organized configuration section

hunchback:
the change should be neutral, only effecting documentation of etc/skydive/yml.default

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -231,19 +231,21 @@ auth:
   analyzer_password: password
 
 etcd:
+  # server parameters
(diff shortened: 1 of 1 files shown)


etc/skydive.yml.default
```diff
@@ -231,20 +231,22 @@ auth:
   analyzer_password: password
 
 etcd:
+  # server parameters
```
safchain: There is already a PR #648 re-working the config file.  I propose to keep only one PR for the whole config refactoring. Maybe you could comment the PR #648.

hunchback: ok, closing the PR

PR #723, opened 2018-01-17
docker: use common.NormalizeValue() to set "Labels"

hunchback:
Test with both a docker probe and verify "Labels" field is in-tacked


safchain:

PR #728, opened 2018-01-19
http: WebSocket enable/disable of write compression

hunchback:
We discovered that our local analyzer was not able to connect with the remote analyzer (in the CogNETive demo landscape) if websocket compression was enabled, I have made this configurable (leaving current behavior as-is by setting default to 'true'). To test this you can override the default value by setting:
```
agent.http.websocket.enable_write_compression = false
```

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -3,6 +3,15 @@
 # host_id is used to reference the agent by default set to hostname
 # host_id:
 
+http:
+  # define the Cookie HTTP Request Header
+  # cookie: 
+  #   <name1>: <value1>
+  #   <name2>: <value2>
+  # rest:
+  #   log the HTTP client request and response (to log level DEBUG)
+  #   debug: false
+
 # WebSocket delay between two pings.
 # ws_ping_delay: 2
 # WebSocket Ping/Pong timeout in second.
@@ -106,13 +115,6 @@ agent:
   # Not required, but can be used to allow virtual hosting
   # X509_servername: domain.com
   #
-  http:
-    # log the HTTP client request and response (to log level DEBUG)
-    # debug: false
-    # define the Cookie HTTP Request Header
-    # cookie: 
-    #   <name1>: <value1>
-    #   <name2>: <value2>
   topology:
     # Probes used to capture topology informations like interfaces,
     # bridges, namespaces, etc...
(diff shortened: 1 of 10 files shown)


hunchback:
Failed the following test:
```
09:30:49.170 --- FAIL: TestOvsMirror (53.98s)
09:30:49.170 	tests.go:379: Replaying test with time 2018-01-19 10:30:13.802818093 +0100 CET m=+344.238144030 (Unix: 1516354213), startTime 2018-01-19 10:30:13.673019604 +0100 CET m=+344.108345505 (Unix: 1516354213)
09:30:49.170 	tests.go:385: Failed to replay test: Unable to find the expected Mirror interface: No result found, graph: 
09:30:49.170 		Warning: Layouter could only place 24 nodes/27 edges out of 24/28 - giving up at /usr/bin/graph-easy line 144.
```

http/wsclient.go
```diff
@@ -236,7 +236,7 @@ func (c *WSConn) write(msg []byte) error {
 	}
 
 	c.conn.SetWriteDeadline(time.Now().Add(writeWait))
-
+	c.conn.EnableWriteCompression(config.GetConfig().GetBool("agent.http.websocket.enable_write_compression"))
```
safchain: this should be done in the connect function no ?

etc/skydive.yml.default
```diff
@@ -105,12 +107,15 @@ agent:
...
   #
   http:
-    # log the HTTP client request and response (to log level DEBUG)
-    # debug: false
     # define the Cookie HTTP Request Header
     # cookie: 
     #   <name1>: <value1>
     #   <name2>: <value2>
+    # rest:
+    #   log the HTTP client request and response (to log level DEBUG)
```
safchain: Not sure that this is related to the agent, maybe we should move the http section outside of the agent section.

etc/skydive.yml.default
```diff
@@ -105,12 +107,15 @@ agent:
...
+    #   debug: false
+    # websocket:
```
safchain: we already have a bunch of Websocket parameter at the root level maybe we could place this close to them (ws_enable_writre_compression)

etc/skydive.yml.default
hunchback: agree, have moved this into 'http.rest.*'

etc/skydive.yml.default
hunchback: will move it to the ws_ block, but have renamed the entire block from `ws_*` to `http.ws.*`

http/wsclient.go
hunchback: agree, moved it

hunchback:
@safchain - I addressed all issues. Please let me know if it is ok that I renamed `ws_*` to `http.ws.*` ?

PR #731, opened 2018-01-21
K8s: code improvements and de-coupling of (sub)probes

hunchback:
Most of the changes have neutral effect on functionality, the only exceptions are:
- changed configuration k8s.subprobes --> k8s.probes
- the k8s.container probe will not attempt to merge metadata onto nodes created by the docker probe


safchain:
As this set of patches are mostly refactoring thing, I would prefer to provide functional tests first for k8s. I propose to base the tests on a minikube deployment so that it is easy to start in a CI job. That could be something similar to what we do with openstack. What do you think ?

lebauce:
@hunchback This code has already been renamed a few times in a short period. Maybe we should slow down a bit on renaming :-)

hunchback:
@lebauce - the re-structuring was the purpose (renaming being a side effect), some stuff addressed in the PR:
- the de-coupling of networkpolicy from pod (such that each pod is self-contained an independent)
- simplified registration of new pods (by adding them to the bundle but not requiring a specific field within k8s.Probe
- the merging of cache.go and client.go eliminating redundant code
- separating k8s.container from docker, important if we want to have both k8s probe and docker probe live side by side.

Immediate next step
- standardizing k8s Edges (such that each be marked with { Manager: k8s } ... this together with the above will enable by using filter-our all k8s or display only k8s
- add sync of k8s data elements every so often (say one every 30 seconds) to ensure skydive topology does not diverge from actual topology

Future steps (after I complete unit tests for k8s):
- improve HostDelete such that the agent only can delete elements added by the agent (and analyzer can only delete stuff added by the analyzer)
- adding of additional k8s elements (Service, Ingress, Deployment, etc.) - so making the registration of each probe streamlined and totally decoupled is imperative 

@safchain - as to adding tests based on minikube - I already have automation for setup of minikube environment and manipulation of various objects (pods, network-policy, etc.), I guess in the timeline of about a week or two I will provide an initial PR providing some coverage.

@lebauce, @safchain - in any case please review the current PR, Thanks!

PR #733, opened 2018-01-22
Dragonflow changes

snapiri:
Added modifications to the websocket client for better usability and debugging.

### DIFF
--- a/contrib/python/api/setup.py
+++ b/contrib/python/api/setup.py
@@ -1,8 +1,8 @@
 from setuptools import setup
 
 setup(name='skydive-client',
-      version='0.2.0',
-      description='Skydive Python 3 client library',
+      version='0.3.0',
+      description='Skydive Python client library',
       url='http://github.com/skydive-project/skydive',
       author='Sylvain Afchain',
       author_email='safchain@gmail.com',
@@ -15,5 +15,7 @@ setup(name='skydive-client',
       },
       install_requires=[
           'autobahn>=0.17.1',
+          'trollius>=2.1;python_version<"3.0"',
+          'asyncio>=3.4.3;python_version>="3.0"',
       ],
       zip_safe=False)
--- a/contrib/python/api/skydive/websocket/client.py
+++ b/contrib/python/api/skydive/websocket/client.py
@@ -24,6 +24,7 @@ try:
 except ImportError:
     import trollius as asyncio
 import base64
+import functools
 import json
 try:
     import http.client as httplib
@@ -98,6 +99,10 @@ class WSClientDefaultProtocol(WebSocketClientProtocol):
 
     def stop(self):
         self.factory.client.loop.stop()
+
+    def stop_when_complete(self):
+        self.factory.client.loop.call_soon(
+            functools.partial(self.factory.client.loop.stop))
 
 
 class WSClientDebugProtocol(WSClientDefaultProtocol):
--- a/contrib/python/api/skydive/websocket/client.py
+++ b/contrib/python/api/skydive/websocket/client.py
@@ -136,6 +136,7 @@ class WSClient(WebSocketClientProtocol):
                  username="", password="", cookie=None,
                  sync="", filter="", persistent=True,
                  **kwargs):
+        super(WSClient, self).__init__()
         self.host_id = host_id
         self.endpoint = endpoint
         self.username = username
--- a/contrib/python/api/skydive/websocket/client.py
+++ b/contrib/python/api/skydive/websocket/client.py
@@ -94,8 +94,14 @@ class SyncRequestMsg:
 
 class WSClientDefaultProtocol(WebSocketClientProtocol):
 
+    def debug_send(self, func, arg):
+        LOG.debug("Running func %s", func.__name__)
+        func(arg)
+
     def sendWSMessage(self, msg):
-        self.sendMessage(msg.to_json().encode())
+        self.factory.client.loop.call_soon(
+            functools.partial(self.debug_send, self.sendMessage,
+                              msg.to_json().encode()))
 
     def stop(self):
         self.factory.client.loop.stop()
@@ -127,6 +133,10 @@ class WSClientDebugProtocol(WSClientDefaultProtocol):
 
     def onClose(self, wasClean, code, reason):
         LOG.debug("WebSocket connection closed: %s", reason)
+
+    def sendWSMessage(self, msg):
+        LOG.debug("Sending message: %s", msg.to_json())
+        super(WSClientDebugProtocol, self).sendWSMessage(msg)
 
 
 class WSClient(WebSocketClientProtocol):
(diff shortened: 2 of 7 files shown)


lebauce:
ok to test

lebauce:
run python-tests

safchain:
Thanks for submitting this. Our python CI job reported some issues https://ci.skydive.network/job/skydive-python-tests/18/console.

safchain:
run python-tests

snapiri:
run python-tests

snapiri:
@lebauce, @safchain How can I re-run the tests? I believe they failed on things that are not related to my changes...
Thanks!

safchain:
add to whitelist

safchain:
@snapiri to rerun tests you can do what you tried but you weren't in the whitelist. I just added you.

safchain:
run python-tests

safchain:

safchain:
@snapiri so the functional tests are failing, going to have a look to understand why

snapiri:
@safchain I think I understand what happened. the test uses onOpen.
After the connect method is called, the client is connected and the onOpen is called before the start() is called.
The onOpen queues the messages and then calls stop() which kills the loop.
I believe what should be done is to change the test to work with async data, the onOpen should queue a stop message and not call the stop().
I will try to add code  that handles it.

safchain:
run python-tests

snapiri:
run python-tests

safchain:
run python-tests

snapiri:
run python-tests

safchain:
run python-tests

safchain:
run python-tests

safchain:

safchain:
rerun selenium-tests

safchain:
rerun selenium-tests

contrib/python/api/skydive/wsshell.py
```diff
@@ -141,7 +142,11 @@ def main():
                       mode=args.mode,
                       file=file)
     client.connect()
+    client.start()
 
 
 if __name__ == '__main__':
+    logging.basicConfig()
```
safchain: I think this should be in the main function otherwise this is not going to be called by the skydive-ws-client(entry_points)

contrib/python/api/skydive/wsshell.py
```diff
@@ -141,7 +142,11 @@ def main():
...
+    WSClient.LOG.setLevel(logging.DEBUG)
```
safchain: I think you can't access to the LOG variable without importing it

contrib/python/api/skydive/wsshell.py
snapiri: This should be called only once per process, thus in the main. I can add a comment in the code that initialization of the logging module should be called for the logs to work.

contrib/python/api/skydive/wsshell.py
snapiri: We have imported the entire package:
from skydive.websocket.client import WSClient (line 28)
so, everything in it is accessible, there is no need to import it explicitly,

contrib/python/api/skydive/wsshell.py
snapiri: If you believe I should do otherwise, I am open to suggestions :)

contrib/python/api/skydive/websocket/client.py
```diff
@@ -19,19 +19,29 @@
...
 
-import asyncio
+try:
+    import asyncio
+except ImportError:
+    import trollius as asyncio
 import base64
+import functools
 import json
 import http.client as httplib
```
safchain: need to address import issue python2/python3 ?

```
try:
    import http.client
except ImportError:
    import httplib
```

contrib/python/api/skydive/websocket/client.py
```diff
@@ -19,19 +19,29 @@
 # under the License.
 #
 
-import asyncio
+try:
+    import asyncio
+except ImportError:
+    import trollius as asyncio
```
safchain: can you update the requirements.txt

contrib/python/api/skydive/wsshell.py
safchain: mm the class is imported not the whole package ? get the following error running it

```
Traceback (most recent call last):
  File "skydive/wsshell.py", line 150, in <module>
    WSClient.LOG.setLevel(logging.DEBUG)
AttributeError: type object 'WSClient' has no attribute 'LOG'
```

contrib/python/api/skydive/wsshell.py
safchain: my point was that placing the log initialization here we don't get any output with command line client `skydive-ws-client`

ex: skydive-ws-client listen

moving these lines within the main function directly solves the issue and we get outputs.

contrib/python/api/skydive/wsshell.py
snapiri: Sorry, my bad. You are right. Will fix.

contrib/python/api/skydive/wsshell.py
snapiri: OK, will fix.

snapiri: Maybe should also consider another patch to support log levels, so we will not have all the debug prints all the time :)

contrib/python/api/skydive/websocket/client.py
snapiri: Will do.

contrib/python/api/skydive/websocket/client.py
snapiri: Will fix

snapiri:
run python-tests

snapiri:
run python-tests

snapiri:

contrib/python/api/skydive/wsshell.py
```diff
@@ -74,6 +77,7 @@ def onOpen(self):
 
 
 def main():
+    logging.basicConfig(level=logging.INFO)
```
safchain: Can you use DEBUG here otherwise we don't get the messages. This is mostly a debugging tool so make sense to have everything in DEBUG anyway.

contrib/python/api/skydive/websocket/client.py
safchain: Sorry forget to say something, can you update the requirements in the setup.cfg too and can you increase the minor version of the package (0.3.0) so that once your patch merged I'll push the new version to pypi

contrib/python/api/skydive/wsshell.py
safchain: or maybe a new argument...

snapiri: Done on both changes.

snapiri:
run python-tests

contrib/python/api/setup.py
```diff
@@ -15,5 +15,7 @@
       },
       install_requires=[
           'autobahn>=0.17.1',
+          'trollius>=2.1;python_version<"3.0"'
```
safchain: "," is missing at the end of this line

snapiri:
run python-tests

snapiri:
rerun unit-tests

safchain:
@snapiri thanks again !

snapiri:
My pleasure :)

safchain:
just uploaded the new version : https://pypi.python.org/pypi/skydive-client/0.3.0

snapiri:
@safchain Excellent! Thanks a lot!

PR #747, opened 2018-01-29
Use only trollius and not asyncio on both python versions

snapiri:
Using asyncio for python3 and trollius for python2 requires us to have
both packages in our dependency list (requirements.txt and setup.py).
When using automated installers, they may parse both packages as
required, resulting in unusable python2 environment (asyncio is not
compatible with python2).

As trollius is compatible with both versions, it would be easier to use
it on both platforms instead.

### DIFF
--- a/contrib/python/api/requirements.txt
+++ b/contrib/python/api/requirements.txt
@@ -1,4 +1,3 @@
 autobahn>=0.17.1
-trollius>=2.1; python_version < '3.0'
-asyncio>=3.4.3; python_version >= '3.0'
+trollius>=2.1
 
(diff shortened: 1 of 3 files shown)


snapiri:
Just a note - possible alternative to this patch would be to modify the setup.py, using different dependencies for each platform (identify using sys.version_info[0] ).
I still believe that using the same package for both platforms is better, as there is less chance of issues arising on different platforms.

safchain:
Not sure to understand exactly the issue as we added markers for the each version of python. Do you have a way to reproduce the issue ? I would prefer to keep using asyncio as Trollius (AFAIK) is deprecated 
https://groups.google.com/forum/#!topic/python-tulip/1UvQc3pswG4

snapiri:
OK. I will upload a new patch to fix the dependency problem

PR #748, opened 2018-01-29
flow: do not send empty LastUpdateMetric

safchain:


safchain:

PR #750, opened 2018-01-29
capture: Invoke capture on preconfigured nodes

vperelmax:
Invoke capture on preconfigured nodes automatically
from the beginning.

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -47,6 +47,12 @@ analyzer:
   # topology_filter:
   #   Namespaces: "g.V().Has('Type', 'netns').OutE().BothV()"
   #   Layer2: "g.E().Has('RelationType', 'layer2')"
+  # Section defining things to be invoked on startup
+  # startup:
+  # By default no capturing,  set filter to capture from selected nodes
+  # from the beginning automatically
+  #   capture_gremlin: "G.V().has('Name', NE('lo'))"
+  #   capture_bpf: "port 80"
   # Enable/disable ssh to hosts
   # ssh_enabled: false
   # Flow storage engine
--- a/flow/ondemand/client/client.go
+++ b/flow/ondemand/client/client.go
@@ -39,6 +39,7 @@ import (
 	shttp "github.com/skydive-project/skydive/http"
 	"github.com/skydive-project/skydive/logging"
 	"github.com/skydive-project/skydive/topology/graph"
+	"github.com/skydive-project/skydive/config"
 )
 
 // OnDemandProbeClient describes an ondemand probe client based on a websocket
@@ -378,6 +379,21 @@ func (o *OnDemandProbeClient) Stop() {
 	o.checkForRegistration.Stop()
 }
 
+// InvokeCaptureFromConfig invokes capture based on preconfigured selected SubGraph
+func (o *OnDemandProbeClient) InvokeCaptureFromConfig() () {
+	gremlin := config.GetConfig().GetString("analyzer.startup.capture_gremlin")
+	bpf := config.GetConfig().GetString("analyzer.startup.capture_bpf")
+	if gremlin == "" {
+		return
+	}
+	logging.GetLogger().Infof("Invoke capturing from the startup with gremlin: %s and BPF: %s", gremlin, bpf)
+	capture := types.NewCapture(gremlin, bpf)
+	capture.SocketInfo = true
+	capture.Type = "pcap"
+	o.onCaptureAdded(capture)
+	return
+}
+
 // NewOnDemandProbeClient creates a new ondemand probe client based on Capture API, graph and websocket
 func NewOnDemandProbeClient(g *graph.Graph, ch *api.CaptureAPIHandler, agentPool shttp.WSJSONSpeakerPool, subscriberPool shttp.WSJSONSpeakerPool, etcdClient *etcd.EtcdClient) *OnDemandProbeClient {
 	resources := ch.Index()
@@ -402,6 +418,7 @@ func NewOnDemandProbeClient(g *graph.Graph, ch *api.CaptureAPIHandler, agentPool
 
 	elector.AddEventListener(o)
 	agentPool.AddJSONMessageHandler(o, []string{ondemand.Namespace})
+	o.InvokeCaptureFromConfig()
 
 	return o
 }


config/config.go
```diff
@@ -79,6 +79,8 @@ func init() {
 	cfg.SetDefault("analyzer.topology.probes", []string{})
 	cfg.SetDefault("analyzer.ssh_enabled", false)
 	cfg.SetDefault("analyzer.replication.debug", false)
+	cfg.SetDefault("analyzer.startup_capture", "")
+	cfg.SetDefault("analyzer.startup_capture_bpf", "")
```
hunchback: If these value by default are empty then these two lines can be omitted.

etc/skydive.yml.default
```diff
@@ -47,6 +47,10 @@ analyzer:
   # topology_filter:
   #   Namespaces: "g.V().Has('Type', 'netns').OutE().BothV()"
   #   Layer2: "g.E().Has('RelationType', 'layer2')"
+  # By default no capturing,  set filter to capture from selected nodes
+  # from the beginning automatically
+  # startup_capture: "G.V().has('Name', NE('lo'))"
+  # startup_capture_bpf: "port 80"
```
hunchback: I suggest grouping captures (as I expect configuration for startup_capture will grow) and adding a specific flag to activate or enable.

```
analyzer:
  capture:
    gremlin_filter: "G.V().has('Name', NE('lo'))"
    bpf_filter: "port 80"
    automatic: true
```

flow/ondemand/client/client.go
```diff
@@ -378,6 +378,21 @@ func (o *OnDemandProbeClient) Stop() {
 	o.checkForRegistration.Stop()
 }
 
+// Invoke capture upon start of analyzer on selected from config SubGraph
```
hunchback: golink expected commend to start with name of function:

```
// InvokeCaptureFromConfig ...
```

flow/ondemand/client/client.go
```diff
@@ -378,6 +378,21 @@ func (o *OnDemandProbeClient) Stop() {
...
+func (o *OnDemandProbeClient) InvokeCaptureFromConfig() () {
```
hunchback: The extra presences for empty return value are redundant - so should be:

```
func (o *OnDemandProbeClient) InvokeCapturefromConfig() {
```

flow/ondemand/client/client.go
```diff
@@ -378,6 +378,21 @@ func (o *OnDemandProbeClient) Stop() {
...
+	capture_gremlin := config.GetConfig().GetString("analyzer.startup_capture")
```
hunchback: You are already within the context of a function handling capture so no need to prefix variables with `capture_` form example `gremlin` or `gremlin_filter` may be better

flow/ondemand/client/client.go
```diff
@@ -378,6 +378,21 @@ func (o *OnDemandProbeClient) Stop() {
...
+	capture_bpf := config.GetConfig().GetString("analyzer.startup_capture_bpf")
+	if(capture_gremlin == "" && capture_bpf == "") {
```
hunchback: You are missing a space after `if` statement ... generally I would expect `gofmt` to detect and fix stuff like this - you can invoke a check from top-level running `make fmt`

flow/ondemand/client/client.go
```diff
@@ -378,6 +378,21 @@ func (o *OnDemandProbeClient) Stop() {
...
+		return
+	}
+	logging.GetLogger().Infof("Try to invoke capturing from the startup with gremlin: %s and BPF: %s", capture_gremlin, capture_bpf)
```
hunchback: I would rather rephrase: "Starting automatic capturing using gremlin-filter: %s; bpf-filter: %s"

flow/ondemand/client/client.go
```diff
@@ -378,6 +378,21 @@ func (o *OnDemandProbeClient) Stop() {
...
+	capture := types.NewCapture(capture_gremlin, capture_bpf)
+	capture.SocketInfo = true
```
hunchback: Should this not be also configurable via startup parameters?

flow/ondemand/client/client.go
```diff
@@ -378,6 +378,21 @@ func (o *OnDemandProbeClient) Stop() {
...
+	capture.Type = "pcap"
```
hunchback: Again here, would we not want to expose `Type` configuration also?

config/config.go
vperelmax: hi,
yes, ok, I will remove them.
Thanks!

etc/skydive.yml.default
vperelmax: I agree with you, good idea!
Thanks!

flow/ondemand/client/client.go
vperelmax: OK

flow/ondemand/client/client.go
vperelmax: u r right

flow/ondemand/client/client.go
vperelmax: ok :)

flow/ondemand/client/client.go
vperelmax: ok

flow/ondemand/client/client.go
vperelmax: thanks!

flow/ondemand/client/client.go
vperelmax: ok

safchain:
Why not using the API/CLI ? Why adding code to provision API objects ? It seems to me that it should be done by a provisioning tool, puppet, ansible, or modifying the docker image adding the ability to start a post script using the client.

eranra:
Hi Sylvain; Yes, it is possible to use external tools and, by the way, there is also the possibility to configure persistent external etcd to be used by the analyzers. 

The specific use-case we are trying to address in this PR is when skydive analyzer/s are deployed into k8s and there is requirement for ‘forever’ capture. In such use-case we see on one hand cases of k8s restarts of the analyzer/s pod/s (including the internal etcd created by the analyzer) and on the other hand we don’t have the external tooling to recreate the captures and we are required for simplicity and simple deployment processes (so we don't want to enforce installation of tools like puppet and ansible and the configuration to communicate with the Analyzer/s to verify captures) 

I think that It make sense to have as part of the product useful autostart configurable parameters, such that add sustainability and the ability of Skydive to be self-deployed.

safchain:
Hi Eran thanks for the explanation. So let introduce something just a bit more generic as we may want to address capture and alert and... I going to comment the code itself

eranra:
Hi Sylvain,

Yes, I agree it makes sense to have something more generic ... 

my 2 cents thoughts in that area ...  I was thinking on something to connect this area with what Masco was doing for the performance panel, and the save performance to file feature. Maybe, on auto-start we will be able to load a file which is an output of what is saved from the performance panel. When we load that file, we will have some auto-starts, so not only captures, but also filters, parameters for the bandwidth ... etc ... that way we can allow save and load of that file, and a user would be able to change those ... not sure all of that make sense, it is a blend of two things the need to allow saving of parameters, and the need to have several different autostart parameters.

config/config.go
safchain: I would rather do

```
  analyzer:
    startup:
       captures:
           <list of capture>
       alerts
           <list of alerts>

so that we can register more than one captures. I do not suggest to handle alerts now, just thinking

flow/ondemand/client/client.go
safchain: I would not put the provisioning code here, otherwise we will have to put provisioning stuffs (alerts, etc) everywhere in the code. I would rather create a dedicated file close to the analyzer/server.go file. So if we decide to change something we will have only one entry point to change.

safchain:
@lebauce will start working on "Skydive" scripts that could be used for that purpose. Meanwhile I think we could start here with something simple that could be converted later. Either we keep the current approach of this PR, moving the code to a central point or we just add an attribute indicating a script(shell) that will be executed. The script could do API request.

etc/skydive.yml.default
vperelmax: changed it according in a bit different way.

flow/ondemand/client/client.go
vperelmax: rephrased

flow/ondemand/client/client.go
vperelmax: no, we would prefer minimizing number of parameters. Too difficult to manage too many.

flow/ondemand/client/client.go
vperelmax: again, we prefer to keep less parameters.

PR #742, opened 2018-01-25
K8s: fixed issues related to links / ownership links

hunchback:
Make improvements and fixes to the following links:
- pod --ownership--> node
- container --ownership--> pod
- container --association--> docker
- networkPolicy --association--> pod

PR #752, opened 2018-01-31
config: default host_id is "<host>-<pid>"

hunchback:
You can tests the effects of this by setting k8s and running  `skydive allinone -c skydive.yml` the result (compared to before the fix) is that you will see all the k8s objects (which now will not be deleted by the agent issued `HostGraphDelMsgType`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -44,12 +44,16 @@ var (
 	ErrNoAnalyzerSpecified = errors.New("No analyzer specified in the configuration file")
 )
 
-func init() {
+func hostID() string {
 	host, err := os.Hostname()
 	if err != nil {
 		panic(err)
 	}
+	pid := os.Getpid()
+	return fmt.Sprintf("%s-%d", host, pid)
+}
 
+func init() {
 	cfg = viper.New()
 
 	cfg.SetDefault("agent.flow.probes", []string{"gopacket", "pcapsocket"})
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -44,12 +44,16 @@ var (
 	ErrNoAnalyzerSpecified = errors.New("No analyzer specified in the configuration file")
 )
 
-func init() {
+func hostID() string {
 	host, err := os.Hostname()
 	if err != nil {
 		panic(err)
 	}
+	pid := os.Getpid()
```
safchain: It will break the TID as it is based on host id

hunchback: @safchain - ok, so I am considering one of the following (please review):

(1) only for the analyzer set the `<host>:<pid>` while leaving for the agent <host>
(2) have in each graph two field (instead of Host)
     Hostname: the hostname from which data is collected (resolved within `NewGraphFromConfig()` using `os.Hostname()`) this value used for deriving TID, and is equal to "Name" field for "Type" == "host" Nodes.
     HostID: which is set to `<hostname>-<pid>` or `host_id` (if set in config) and is used for implementing HostGraphDelMsgType

Please advise ...

eranra: Or another option is to hard-code a suffix for the analyzer <hostname>-analyzer and leave the agents without any change (or also suffix with agent) ... this should not break the TID :-)

safchain: So host id is means "unique host" not process, and I think we need to keep this as it is. I'm not very sold about changing something that is provided by the user, by appending something. So the question why addressing host-binding issue changing a default value. If really needed the user should provide a proper config. If the issue is k8s resources are deleted by agent stop/start, I think that the k8s resources should not belong to any host, like fabric node. I order to have the same behavior we just need to pass empty string as last parameter of NewNode.

hunchback: @safchain - ok I will go with passing "" to `NewNode()`

hunchback: Addressed as proposed in: #742  - closing this PR

PR #754, opened 2018-01-31
ui: add file support for preference

masco-enovance:
preference can be download to file to share with others
and upload from file.

### DIFF
--- a/statics/js/components/preference.js
+++ b/statics/js/components/preference.js
@@ -11,7 +11,7 @@ var PreferenceComponent = {
       <div class="form-group">\
         <label for="default-favorite">Favorite Gremlin Expressions</label>\
         <a><i class="fa fa-question help-text" aria-hidden="true" title="Filter and highlight gremlin queries, displayed in the left panel"></i></a>\
-        <div v-for="favorite in favorites">\
+        <div v-for="favorite in preferences.favorites">\
           <div class="form-group">\
             <div class="input-group">\
               <label class="input-group-addon">Name: </label>\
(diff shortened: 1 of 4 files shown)


statics/js/components/preference.js
```diff
@@ -73,7 +73,10 @@ var PreferenceComponent = {
         </div>\
       </div>\
       <div class="button-holder row">\
-        <button class="btn btn-lg btn-primary" type="submit">Save</button>\
+        <button class="btn btn-lg btn-primary" type="submit" title="save in local storage">Save</button>\
```
eranra: Can this be automatic behavior of the page, so you don't need to click anything ... just add and remove filters, or change parameters and as part of the change it will be kept in 'localStorage'

statics/js/components/preference.js
```diff
@@ -73,7 +73,10 @@ var PreferenceComponent = {
...
+        <button class="btn btn-lg btn-primary" type="button" @click="saveToFile" title="download the preferences in local file"> Download</button>\
```
eranra: "download preferences to local file"

statics/js/components/preference.js
```diff
@@ -73,7 +73,10 @@ var PreferenceComponent = {
...
+        <input type="file" @change="openfile($event)" id="file_selector" style="display:none;">\
+        <button class="btn btn-lg btn-primary" type="button" @click="uploadFile" title="upload the preferences from local file"> Upload</button>\
```
eranra: same

statics/js/components/preference.js
```diff
@@ -149,7 +150,52 @@ var PreferenceComponent = {
         if (f.name !== "" && f.filter !== "") newList.push(f);
       });
       return newList;
-    }
+    },
+
+    saveToFile: function() {
+      var obj = {};
+      obj["filters"] = JSON.stringify(this.filterEmpty(this.filters));
```
eranra: Is there a way to iterate on the page so we don't need to add specific 'key' for each field, can you tag them in the html and iterate on all tags .... this will allow to add more and more parameters without need to change the save and load functions everytime

safchain: +1
we could simply add a "preferences" field to "this" like 

JSON.stringify(this.preferences)

and preferences would be like this 

```
this.preferences = {
  bandwidthThreshold:
  ...
```

masco-enovance: done

statics/js/components/preference.js
masco-enovance: done

statics/js/components/preference.js
masco-enovance: done

statics/js/components/preference.js
masco-enovance: as we discussed in the call, currently added a warning for the cancel button

lebauce:
@masco-enovance Can you please this PR against master ?

masco-enovance:

statics/js/components/preference.js
```diff
@@ -33,114 +33,97 @@ var PreferenceComponent = {
...
           <label for="bw-rel-alert">Bandwidth Relative Alert</label>\
           <a><i class="fa fa-question help-text" aria-hidden="true" title="Alert threshold in between 0 to 1"></i></a>\
-          <input id="bw-rel-alert" type="number" class="form-control input-sm" v-model="bwRelAlert" min="0" max="1" step="0.1"/>\
+          <input id="bw-rel-alert" type="number" class="form-control input-sm" v-model="preferences.bandwidthRelativeAlert" min="0" max="1" step="0.1"/>\
         </div>\
       </div>\
       <div class="button-holder row">\
-        <button class="btn btn-lg btn-primary" type="submit">Save</button>\
+        <button class="btn btn-lg btn-primary" type="submit" title="save in local storage">Save</button>\
+        <button class="btn btn-lg btn-primary" type="button" @click="saveToFile" title="download preferences to local file"> Download</button>\
```
safchain: I would rather use Export/Import aligning these button to the left for ex and the save/cancel to the right with a space between the two groups. That said I had an issue, I changed the preferences, I downloaded, then I cancel and the "cancelled" preferences were applied.

statics/js/components/preference.js
```diff
@@ -33,114 +33,97 @@ var PreferenceComponent = {
...
         <div class="form-group">\
           <label for="bw-rel-alert">Bandwidth Relative Alert</label>\
           <a><i class="fa fa-question help-text" aria-hidden="true" title="Alert threshold in between 0 to 1"></i></a>\
-          <input id="bw-rel-alert" type="number" class="form-control input-sm" v-model="bwRelAlert" min="0" max="1" step="0.1"/>\
+          <input id="bw-rel-alert" type="number" class="form-control input-sm" v-model="preferences.bandwidthRelativeAlert" min="0" max="1" step="0.1"/>\
         </div>\
       </div>\
       <div class="button-holder row">\
-        <button class="btn btn-lg btn-primary" type="submit">Save</button>\
+        <button class="btn btn-lg btn-primary" type="button" @click="saveToFile" title="download preferences to local file"> Download</button>\
```
lebauce: I think that "Import" and "Export" would be clearer that "Download" / "Upload"

safchain:
I lost many preferences fields if I import this exported file

```
"{\"favorites\":[],\"bandwidthThreshold\":\"relative\"}"
```

masco-enovance:
before writing to the file JSON stringfy happened two times, now working fine.

PR #756, opened 2018-02-02
status: list available probes in status api

masco-enovance:


probe/bundle.go
```diff
@@ -67,6 +67,15 @@ func (p *ProbeBundle) GetProbe(name string) Probe {
 	return nil
 }
 
+//ActiveProbes returns all active probes name
+func (p *ProbeBundle) ActiveProbes() []string {
+	activeProbes := make([]string, 0, len(p.probes))
```
safchain: I think you need to RLock there

PR #758, opened 2018-02-05
flow: treat flows buffer overflow state

vperelmax:
If we reach flow limit, all the flows from the channel are cleared
without saving the data and we send notification about it to the log.

### DIFF
--- a/analyzer/flow_server.go
+++ b/analyzer/flow_server.go
@@ -49,14 +49,20 @@ type FlowServerConn interface {
 
 // FlowServerConn describes a UDP flow server connection
 type FlowServerUDPConn struct {
-	conn *net.UDPConn
+	conn                   *net.UDPConn
+	timeOfLastLostFlowsLog time.Time
+	numOfLostFlows         int
+	maxFlowBufferSize      int
 }
 
 // FlowServerConn describes a WebSocket flow server connection
 type FlowServerWebSocketConn struct {
 	shttp.DefaultWSSpeakerEventHandler
-	server *shttp.Server
-	ch     chan *flow.Flow
+	server                 *shttp.Server
+	ch                     chan *flow.Flow
+	timeOfLastLostFlowsLog time.Time
+	numOfLostFlows         int
+	maxFlowBufferSize      int
 }
 
 // FlowServer describes a flow server with pipeline enhancers mechanism
@@ -81,6 +87,16 @@ func (c *FlowServerWebSocketConn) OnMessage(client shttp.WSSpeaker, m shttp.WSMe
 		return
 	}
 	logging.GetLogger().Debugf("New flow from Websocket connection: %+v", f)
+	if len(c.ch) >= c.maxFlowBufferSize {
+		c.numOfLostFlows++
+		if c.timeOfLastLostFlowsLog.IsZero() ||
+			(time.Now().Sub(c.timeOfLastLostFlowsLog) >= time.Second) {
+			logging.GetLogger().Errorf("Buffer overflow - too many flow updates, removing and not storing flows: %d", c.numOfLostFlows)
+			c.timeOfLastLostFlowsLog = time.Now()
+			c.numOfLostFlows = 0
+		}
+		return
+	}
 	c.ch <- f
 }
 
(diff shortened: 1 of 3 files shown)


safchain:
add to whitelist

analyzer/flow_server.go
```diff
@@ -82,6 +83,10 @@ func (c *FlowServerWebSocketConn) OnMessage(client shttp.WSSpeaker, m shttp.WSMe
 		return
 	}
 	logging.GetLogger().Debugf("New flow from Websocket connection: %+v", f)
+	if len(c.ch) >= c.maxFlowBufferSize {
+		logging.GetLogger().Criticalf("Buffer overflow - too many flow updates, removing and not storing flow: %s", f.UUID)
```
safchain: This will generate a lot of log that will slow down the chan. I think we should limit the number of log, maybe check last time we logged and keeping 1sec between two logs?

vperelmax: OK

analyzer/flow_server.go
```diff
@@ -81,7 +85,19 @@ func (c *FlowServerWebSocketConn) OnMessage(client shttp.WSSpeaker, m shttp.WSMe
 		logging.GetLogger().Errorf("Error while parsing flow: %s", err.Error())
 		return
 	}
-	logging.GetLogger().Debugf("New flow from Websocket connection: %+v", f)
+	//logging.GetLogger().Debugf("New flow from Websocket connection: %+v", f)
+	if len(c.ch) >= c.maxFlowBufferSize {
```
safchain: this should be done for UDP too

analyzer/flow_server.go
```diff
@@ -43,7 +43,7 @@ import (
 	"github.com/skydive-project/skydive/topology/graph"
 )
 
-// FlowServerConn describes a flow server connection
+// FlowServerConn describes a fbufferOverflowInfo     BufferOverflowInfolow server connection
```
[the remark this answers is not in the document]
vperelmax: OK

analyzer/flow_server.go
```diff
@@ -81,7 +85,19 @@ func (c *FlowServerWebSocketConn) OnMessage(client shttp.WSSpeaker, m shttp.WSMe
 		logging.GetLogger().Errorf("Error while parsing flow: %s", err.Error())
 		return
 	}
-	logging.GetLogger().Debugf("New flow from Websocket connection: %+v", f)
+	//logging.GetLogger().Debugf("New flow from Websocket connection: %+v", f)
+	if len(c.ch) >= c.maxFlowBufferSize {
+		c.mutexLostFlows.Lock()
```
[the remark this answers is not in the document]
vperelmax: Removed

analyzer/flow_server.go
```diff
@@ -81,7 +85,19 @@ func (c *FlowServerWebSocketConn) OnMessage(client shttp.WSSpeaker, m shttp.WSMe
...
+		defer c.mutexLostFlows.Unlock()
+		c.numOfLostFlows++
+		if c.timeOfLastLostFlowsLog.IsZero() ||
+			(c.timeOfLastLostFlowsLog.Sub(time.Now()) >= time.Second) {
+			logging.GetLogger().Criticalf("Buffer overflow - too many flow updates, removing and not storing flows: %d", c.numOfLostFlows)
```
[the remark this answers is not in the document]
vperelmax: ok

analyzer/flow_server.go
vperelmax: ok

analyzer/flow_server.go
```diff
@@ -81,7 +87,23 @@ func (c *FlowServerWebSocketConn) OnMessage(client shttp.WSSpeaker, m shttp.WSMe
 		logging.GetLogger().Errorf("Error while parsing flow: %s", err.Error())
 		return
 	}
-	logging.GetLogger().Debugf("New flow from Websocket connection: %+v", f)
+	//logging.GetLogger().Debugf("New flow from Websocket connection: %+v", f)
+	if len(c.ch) >= c.maxFlowBufferSize {
+		c.numOfLostFlows++
+		if c.timeOfLastLostFlowsLog.IsZero() ||
+			(c.timeOfLastLostFlowsLog.Sub(time.Now()) >= time.Second) {
```
safchain: time.Now().Sub(c.timeOfLastLostFlowsLog), otherwise it will never match

analyzer/flow_server.go
```diff
@@ -81,7 +87,23 @@ func (c *FlowServerWebSocketConn) OnMessage(client shttp.WSSpeaker, m shttp.WSMe
...
+			if c.numOfLostFlows < c.maxFlowBufferSize/4 {
```
safchain: Can you elaborate on why purging the chan will help instead of just skipping ? moreover you purge the chan only  when logging thus each second. I think that skipping new flow should be enough letting the "insert" goroutine consuming the chan purging it.

analyzer/flow_server.go
vperelmax: Done

analyzer/flow_server.go
vperelmax: I just skip now, you are right

PR #760, opened 2018-02-06
flow: don't store updates for idle connections

vperelmax:
If a connection wasnt changed since last update - dont store
its update in the backend DB

### DIFF
--- a/analyzer/flow_server.go
+++ b/analyzer/flow_server.go
@@ -157,9 +157,17 @@ func NewFlowServerUDPConn(addr string, port int) (*FlowServerUDPConn, error) {
 func (s *FlowServer) storeFlows(flows []*flow.Flow) {
 	if s.storage != nil && len(flows) > 0 {
 		s.enhancerPipeline.Enhance(s.enhancerPipelineConfig, flows)
-		s.storage.StoreFlows(flows)
-
-		logging.GetLogger().Debugf("%d flows stored", len(flows))
+		var flowsToStore []*flow.Flow
+		// Store only flows that had changes since last update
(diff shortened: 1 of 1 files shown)


analyzer/flow_server.go
```diff
@@ -157,9 +157,17 @@ func NewFlowServerUDPConn(addr string, port int) (*FlowServerUDPConn, error) {
...
 	if s.storage != nil && len(flows) > 0 {
 		s.enhancerPipeline.Enhance(s.enhancerPipelineConfig, flows)
-		s.storage.StoreFlows(flows)
-
-		logging.GetLogger().Debugf("%d flows stored", len(flows))
+		var flowsToStore []*flow.Flow
+		// Store only flows that had changes since last update
+		for _, fl := range flows {
+			if fl.LastUpdateMetric == nil ||
+				fl.LastUpdateMetric.ABPackets > 0 ||
```
safchain: Flow shouldn't be sent by agent if not updated so this shouldn't happen. Did you see any flow here without modification ?

safchain: If it happens this should probably be fixed on the agent side within the flow table

vperelmax: First of all, you are right. Now I look at it and I see that here should be no problem.
Second - if we really save only updated flows I think I have seen some bug related to LastUpdate.
I am closing the feature and will try later to reproduce the issue if it exists. 
Anyway, if there is a bug, it is not in this area, I will later try to re-test it with the client that has shown it to me. Maybe there we also didnt see any wrong behavior.

PR #768, opened 2018-02-08
ui: Add launch in context /topology?filter=<G>;hightlight=<G>

hunchback:
- Open Skydive UI with custom view based on URL query string.
- Support for Multiple query keys,values

- Current support is for filter and highlight (gremlin expression)

Some examples:

http://localhost:8082/topology?filter=G.V().Has("Name", "AGGR1")

http://localhost:8082/topology?filter=G.V().Has("Name", "AGGR1")&highlight=G.V().Has("Name", Regex("AGGR1"))

http://localhost:8082/topology?filter=G.V().Has("Probe", Regex("fabric"))&highlight=G.V().Has("Name", Regex("AGGR1"))

Co-authored-by:    eranra <eranra@ERAN-TP>

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -306,6 +306,7 @@ func NewServer(host string, serviceType common.ServiceType, addr string, port in
 	router.PathPrefix("/statics").HandlerFunc(server.serveStatics)
 	router.PathPrefix(ExtraAssetPrefix).HandlerFunc(server.serveStatics)
 	router.HandleFunc("/login", server.serveLogin)
+	router.PathPrefix("/topology").HandlerFunc(server.serveIndex)
 	router.HandleFunc("/", server.serveIndex)
 
 	return server
--- a/statics/js/components/topology.js
+++ b/statics/js/components/topology.js
@@ -285,6 +285,21 @@ var TopologyComponent = {
 
     this.setGremlinFavoritesFromConfig();
 
+    if (typeof(this.$route.query.highlight) !== "undefined") {
+      this.topologyEmphasize = this.$route.query.highlight;
+      setTimeout(function(){self.emphasizeGremlinExpr();}, 2000)
+    }
+
+    if (typeof(this.$route.query.filter) !== "undefined") {
+      this.topologyFilter = this.$route.query.filter;
+    }
+
+    websocket.addConnectHandler(function() {
+      if (self.topologyFilter !== '') {
+        self.topologyFilterQuery();
+      }
+    });
+
     $.when(this.$getConfigValue('analyzer.ssh_enabled').
       then(function(sshEnabled) {
         self.isSSHEnabled = sshEnabled;
--- a/statics/js/app.js
+++ b/statics/js/app.js
@@ -134,7 +134,7 @@ var routes = [
       }
     }
   },
-  { path: '/topology', component: TopologyComponent },
+  { path: '/topology', component: TopologyComponent, props: (route) => ({ query: route.query }) },
   { path: '/conversation', component: ConversationComponent },
   { path: '/discovery', component: DiscoveryComponent },
   { path: '/preference', component: PreferenceComponent },
@@ -142,6 +142,7 @@ var routes = [
 ];
 
 var router = new VueRouter({
+  mode: 'history',
   linkActiveClass: 'active',
   routes: routes
 });


safchain:
run cdd-overview

http/server.go
```diff
@@ -306,6 +306,7 @@ func NewServer(host string, serviceType common.ServiceType, addr string, port in
 	router.PathPrefix("/statics").HandlerFunc(server.serveStatics)
 	router.PathPrefix(ExtraAssetPrefix).HandlerFunc(server.serveStatics)
 	router.HandleFunc("/login", server.serveLogin)
+	router.PathPrefix("/topology").HandlerFunc(server.serveIndex)
```
safchain: I would prefer to not add an endpoint right now as not used by the backend

statics/js/components/topology.js
```diff
@@ -285,6 +285,21 @@ var TopologyComponent = {
 
     this.setGremlinFavoritesFromConfig();
 
+    if (typeof(this.$route.query.highlight) !== "undefined") {
+        this.topologyEmphasize = this.$route.query.highlight;
```
safchain: nit: indentation,  2 extra spaces

statics/js/components/topology.js
```diff
@@ -285,6 +285,21 @@ var TopologyComponent = {
...
+        setTimeout(function(){self.emphasizeGremlinExpr();}, 2000)
+    }
+
+    if (typeof(this.$route.query.filter) !== "undefined") {
+        this.topologyFilter = this.$route.query.filter;
```
safchain: same here

http/server.go
hunchback: I added this as else working in VueRouter(mode: 'history') does not work as explained:

(https://router.vuejs.org/en/essentials/history-mode.html)[history-mode]

statics/js/components/topology.js
hunchback: ok

statics/js/components/topology.js
hunchback: ok

statics/js/components/topology.js
```diff
@@ -285,6 +285,21 @@ var TopologyComponent = {
 
     this.setGremlinFavoritesFromConfig();
 
+    if (typeof(this.$route.query.highlight) !== "undefined") {
+      this.topologyEmphasize = this.$route.query.highlight;
+      setTimeout(function(){self.emphasizeGremlinExpr();}, 2000)
```
safchain: I don't think the timeout is needed as WebSocket event will trigger the function ? By the way I didn't have it working, It triggers exceptions, please check the console to see if it the same for you

eranra: - Hi Sylvain, when I wrote the original code I didn't had the timeout in place, and was thinking (like you) that there is no need, later I discovered that it just doesn't work with time-out., I checked 1 second and it worked, but decide on 2 seconds just to be on the safe side (also, it looks visually nice to see the nodes, and than the highlight appears).

- With regards to the exception, can you share ... I am running a version that is 99% similar to this one (only change is in the PathPrefix("/topology") that should make a difference) and I don't get any exception ....

safchain: I really think the setTimeout is useless but that's fine to let it I guess, here the exception I get, without or with 

```
topology.js:452 Uncaught TypeError: Cannot read property 'indexOf' of null
    at a.endsWith (topology.js:452)
    at a.n [as endsWith] (vue.min-2.1.9.js:6)
    at a.emphasizeGremlinExpr (topology.js:595)
    at a.n [as emphasizeGremlinExpr] (vue.min-2.1.9.js:6)
    at later (utils.js:8)
```

safchain: The issue is not due to your patch, can you fix it with this change the line https://github.com/skydive-project/skydive/blob/master/statics/js/components/topology.js#L463

by this one 
```
if (favorite) self.topologyEmphasize = favorite;
```

BTW a ";" is missing here at the end of the setTimeout line

hunchback: But we already protect against `favorite` being undefined here: https://github.com/skydive-project/skydive/blob/master/statics/js/components/topology.js#L455

Plus such a fix (if needed) this outside the context of the current PR. (so can be addressed separately).

safchain: Ok I will do

PR #781, opened 2018-02-14
flow: keep a flow table clock to keep flow aligned

safchain:
We need to keep the flow last field aligned with a table
clock as the original timestamp can be seen by the table
after a long period. In this situation we have a flow
that is present in the flow table but that is
not sent to the analayzer since not seen as updated.


lebauce:
run scale-tests

lebauce:
run scale-tests

lebauce:
run scale-tests

lebauce:
run scale-tests

PR #800, opened 2018-02-22
Ui extra assets recurse

hunchback:
Added support in uploading from disk extra static assets which reside under a sub-folder such as:

```
x/y/z/my.css
```

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -32,6 +32,7 @@ import (
 	"mime"
 	"net"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


lebauce:
Can you please rebase on master ?

http/server.go
```diff
@@ -258,15 +264,29 @@ func (s *Server) HandleFunc(path string, f auth.AuthenticatedHandlerFunc) {
...
-	files, err := ioutil.ReadDir(folder)
+func (s *Server) loadExtraAssets(folder, prefix string) {
+	files := []string{}
+
+	err := filepath.Walk(folder, func(path string, info os.FileInfo, err error) error {
+		if info.IsDir() {
+			return nil
+		}
+
+		if filepath.Ext(path) == ".go" {
```
lebauce: It seems that was required in the previous version of the patch. Maybe it can now be removed

hunchback: ok, I have removed this

PR #808, opened 2018-02-24
ui: add automatic expand via launching context /topology?expand

hunchback:
The expected result of launching with "expand" URL parameter is that on when layout is invalid - the following will happen:

- expands all levels
- performs a zoom fit (after some delay)

### DIFF
--- a/statics/js/components/graph-layout.js
+++ b/statics/js/components/graph-layout.js
@@ -9,7 +9,15 @@ var TopologyGraphLayout = function(vm, selector) {
 
   this.queue = new Queue();
   this.queue.await(function() {
-    if (self.invalid) self.update();
+    if (self.invalid) {
+      if (self._autoExpand) {
+        while (self.toggleCollapseByLevel(false));
+        setTimeout(function() {
+          self.zoomFit();
+        }, 5000);
+      }
+      self.update();
+    }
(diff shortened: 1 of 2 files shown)


hunchback:
@safchain - can you please review

statics/js/components/graph-layout.js
```diff
@@ -9,7 +9,15 @@ var TopologyGraphLayout = function(vm, selector) {
 
   this.queue = new Queue();
   this.queue.await(function() {
-    if (self.invalid) self.update();
+    if (self.invalid) {
+      if (self._autoExpand) {
```
safchain: could you declare it in the "constructor", set as false.
You never reset the autoExpand variable meaning that it is going to expand for each events, same thing for the zoomFit. So If a user change the zoom it will be changed. I would prefer to have it just once.

hunchback: the intent was that the view keeps adjusting automatically as new updates arrive ... we could (preferably is a separate PR) add a toggle button ("Automatic Expand") which can be set to on by adding the ?Expand option, but can later be disabled by user (only after disabling one could switch back to manual Expand, Zoom, etc.

safchain: looks OK to me

PR #813, opened 2018-02-25
Fix filename for downloaded capture

### LINKED ISSUE #812: Filename of downloaded pcap is wrong
The filename and "pcap" are concatenated without a "." between. This causes the client to not understand what file type it is and not start the associated tool.
See https://github.com/skydive-project/skydive/blob/master/statics/js/api.js#L18

lindhor:
Changes the name from UUIDpcap to UUID.pcap

closes #812


masco-enovance:
ok to test

PR #814, opened 2018-02-26
tests: split docker tests

lebauce:


tests/docker_test.go
```diff
@@ -0,0 +1,183 @@
+/*
+ * Copyright (C) 2015 Red Hat, Inc.
```
masco-enovance: year?

lebauce: Those tests were in topology_test.go, I kept the original date

lebauce:

tests/docker_test.go
masco-enovance: ok, but the file creation date will be new right? is that fine? if yes no problem for me.

PR #825, opened 2018-02-28
http: do not crash if extra assets folder does not exist

lebauce:


lebauce:
run skydive-selenium-tests

http/server.go
```diff
@@ -268,6 +268,10 @@ func (s *Server) loadExtraAssets(folder, prefix string) {
 	files := []string{}
 
 	err := filepath.Walk(folder, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
```
masco-enovance: No return params for this function. look like we can't return the error here, just log the error and return nil is fine.

lebauce: There is already a line that logs this error line 280

masco-enovance: got it. Thanks!

PR #826, opened 2018-03-01
Ui link latency 

hunchback:
Have added option to toggle to latency on link labels.

How to use? for now this is exposed via launch context or URL parameter: `link_label_type=latency` if not specified then the default is `bandwidth` as was before.

The calculation of latency is by calculating the delta between `Math.abs(ANode.RTT-BNode.RTT)/1000`, with the underlying assumption that clocks are synchronized between nodes.

So as not to stress the system only visible nodes which has Capture enabled participate in latency calculation.

A possible test setup:

```
sudo ./scripts/simple.sh start 10.0.0.1/24 10.0.0.2/24
sudo ifconfig vm1-eth0 10.0.0.1
sudo ifconfig vm2-eth0 10.0.0.2
sudo ip netns exec vm1 tc qdisc add dev eth0 root netem delay 100ms
sudo tc qdisc add dev vm1-eth0 root netem delay 200ms
sudo tc qdisc add dev vm2-eth0 root netem delay 400ms
sudo ip netns exec vm2 tc qdisc add dev eth0 root netem delay 800ms
```

You can now inject traffic running:

```
sudo ip netns vm1 ping 10.0.0.2 -c 1
```

### DIFF
--- a/statics/js/components/graph-layout.js
+++ b/statics/js/components/graph-layout.js
@@ -103,6 +103,14 @@ var LinkLabelLatency = Vue.extend({
       return this.$topologyQuery(query)
     },
 
+    flowQueryByNodeTID: function(nodeTID, limit) {
+      return this.flowQuery(nodeTID, undefined, limit);
+    },
+
+    flowQueryByNodeTIDandTrackingID: function(nodeTID, trackingID) {
+      return this.flowQuery(nodeTID, trackingID, 1);
+    },
+
     updateData: function(link) {
       var self = this;
 
@@ -116,11 +124,11 @@ var LinkLabelLatency = Vue.extend({
         return;
       }
 
-      this.flowQuery(a.TID, undefined, 10)
+      this.flowQueryByNodeTID(a.TID, 10)
         .then(function(flows) {
           for (let i in flows) {
             const aFlow = flows[i];
-            self.flowQuery(b.TID, aFlow.TrackingID, 1)
+            self.flowQueryByNodeTIDandTrackingID(b.TID, aFlow.TrackingID)
               .then(function(flows) {
                 if (flows.length !== 0) {
                   const bFlow = flows[0];
--- a/statics/js/components/graph-layout.js
+++ b/statics/js/components/graph-layout.js
@@ -122,8 +122,8 @@ var LinkLabelLatency = Vue.extend({
             const aFlow = flows[i];
             self.flowQuery(b.TID, aFlow.TrackingID, 1)
               .then(function(flows) {
-                for (let i in flows) {
-                  const bFlow = flows[i];
+                if (flows.length !== 0) {
+                  const bFlow = flows[0];
                   self.updateLatency(link, aFlow, bFlow);
                 }
               })
--- a/statics/js/components/graph-layout.js
+++ b/statics/js/components/graph-layout.js
@@ -1,5 +1,172 @@
+const maxClockSkewMillis = 5 * 60 * 1000; // 5 minutes
+
+var LinkLabelBandwidth = Vue.extend({
+  mixins: [apiMixin],
+
+  methods: {
+
+    bandwidthFromMetrics: function(metrics) {
+      if (!metrics) {
+        return 0;
+      }
+      if (!metrics.Last) {
+        return 0;
+      }
+      if (!metrics.Start) {
+        return 0;
+      }
+
+      const totalByte = (metrics.RxBytes || 0) + (metrics.TxBytes || 0);
+      const deltaMillis = metrics.Last - metrics.Start;
+      const elapsedMillis = Date.now() - new Date(metrics.Last);
+
+      if (deltaMillis === 0) {
+        return 0;
+      }
+      if (elapsedMillis > maxClockSkewMillis) {
+        return 0;
+      }
+
+      return Math.floor(8 * totalByte * 1000 / deltaMillis); // bits-per-second 
+    },
+
+    setup: function(topology) {
+      this.topology = topology;
+    },
+
+    updateData: function(link) {
+      if (!link.target.metadata.LastUpdateMetric) {
+        return;
+      }
+
+      const defaultBandwidthBaseline = 1024 * 1024 * 1024; // 1 gbps
+
+      link.bandwidthBaseline = (this.topology.bandwidth.bandwidthThreshold === 'relative') ?
+        link.target.metadata.Speed || defaultBandwidthBaseline : 1;
+
+      link.bandwidthAbsolute = this.bandwidthFromMetrics(link.target.metadata.LastUpdateMetric);
+      link.bandwidth = link.bandwidthAbsolute / link.bandwidthBaseline;
+    },
+ 
+    hasData: function(link) {
+      if (!link.target.metadata.LastUpdateMetric) {
+        return false;
+      }
+
+      if (!link.bandwidth) {
+        return false;
+      }
+      return link.bandwidth > this.topology.bandwidth.active;
+    },
+
+    getText: function(link) {
+      return bandwidthToString(link.bandwidthAbsolute);
+    },
+
+    isActive: function(link) {
+      return (link.bandwidth > this.topology.bandwidth.active) && (link.bandwidth < this.topology.bandwidth.warning);
+    },
+
+    isWarning: function(link) {
+      return (link.bandwidth >= this.topology.bandwidth.warning) && (link.bandwidth < this.topology.bandwidth.alert);
+    },
+
+    isAlert: function(link) {
+      return link.bandwidth >= this.topology.bandwidth.alert;
+    },
+  },
+});
+
+var LinkLabelLatency = Vue.extend({
+  mixins: [apiMixin],
+
+  methods: {
+    setup: function(topology) {
+      this.active = 0;
+      this.warning = 10
+      this.alert = 100;
+    },
+
+    updateLatency: function(link, a, b) {
+      link.latencyTimestamp = Math.max(a.Last, b.Last);
+      link.latency = Math.abs(a.RTT - b.RTT) / 1000000;
+    },
+
+    flowQuery: function(nodeTID, trackingID, limit) {
+      let query = `G.Flows()`;
+      query += `.Has("NodeTID", "${nodeTID}")`;
+      if (typeof trackingID !== 'undefined') {
+        query += `.Has("TrackingID", "${trackingID}")`;
+      }
+      query += `.Has("RTT", NE(0))`;
+      query += `.Sort().Limit(${limit})`;
+      return this.$topologyQuery(query)
+    },
+
+    updateData: function(link) {
+      var self = this;
+
+      const a = link.source.metadata;
+      const b = link.target.metadata;
+
+      if (!a.Capture) {
+        return;
+      }
+      if (!b.Capture) {
+        return;
+      }
+
+      this.flowQuery(a.TID, undefined, 10)
+        .then(function(flows) {
+          for (let i in flows) {
+            const aFlow = flows[i];
+            self.flowQuery(b.TID, aFlow.TrackingID, 1)
+              .then(function(flows) {
+                for (let i in flows) {
+                  const bFlow = flows[i];
+                  self.updateLatency(link, aFlow, bFlow);
+                }
+              })
+              .fail(function(error) {
+                console.log(error);
+              });
+          }
+        })
+        .fail(function(error) {
+          console.log(error);
+        });
+    },
+
+    hasData: function(link) {
+      if (!link.latencyTimestamp) {
+        return false;
+      }
+      const elapsedMillis = Date.now() - new Date(link.latencyTimestamp);
+      return elapsedMillis <= maxClockSkewMillis;
+    },
+
+    getText: function(link) {
+      return `${link.latency} ms`;
+    },
+
+    isActive: function(link) {
+      return (link.latency >= this.active) && (link.latency < this.warning);
+    },
+
+    isWarning: function(link) {
+      return (link.latency >= this.warning) && (link.latency < this.alert);
+    },
+
+    isAlert: function(link) {
+      return (link.latency >= this.alert);
+    },
+  },
+});
+
 var TopologyGraphLayout = function(vm, selector) {
   var self = this;
+
+  this.linkLabelType = "bandwidth";
 
   this.vm = vm;
 
@@ -69,11 +236,27 @@ var TopologyGraphLayout = function(vm, selector) {
 
   this.loadBandwidthConfig()
     .then(function() {
-      self.bandwidth.intervalID = setInterval(self.updateBandwidth.bind(self), self.bandwidth.updatePeriod);
+      self.bandwidth.intervalID = setInterval(self.updatelinkLatency.bind(self), self.bandwidth.updatePeriod);
     });
 };
 
 TopologyGraphLayout.prototype = {
+
+  linkLabelFactory: function(link) {
+    let type, driver;
+    switch (this.linkLabelType) {
+      case "latency":
+        type = "latency";
+        driver = new LinkLabelLatency();
+        break;
+      default:
+        type = "bandwidth";
+        driver = new LinkLabelBandwidth();
+        break;
+    }
+    driver.setup(this);
+    return driver;
+  },
 
   notifyHandlers: function(ev, v1) {
     var self = this;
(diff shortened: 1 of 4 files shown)


flow/flow.go
```diff
@@ -463,9 +463,11 @@ func (f *Flow) updateMetricsWithLinkLayer(packet *gopacket.Packet, length int64)
 	}
 
 	if f.XXX_state.link1stPacket == 0 {
-		f.XXX_state.link1stPacket = (*packet).Metadata().Timestamp.UnixNano()
-	} else {
-		if (f.RTT == 0) && (f.Link.A == ethernetPacket.DstMAC.String()) {
+		if f.Link.A == ethernetPacket.SrcMAC.String() {
```
safchain: I think this commit is not needed as this is the first packet according to the line 465 and we always set Link.A as Src.

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+      this.alert = 100;
+      this.link = link;
+      this.updateData();                    // prepare data for next cycle
+    },
+
+    flowKey: function(a) {
+      return `TrackingID:${a.TrackingID},ANodeTID:${a.ANodeTID},BNodeTID:${a.BNodeTID}`;
+    },
+
+    linklabel: function(a, b) {
```
safchain: doesn't seem to be called

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+    hasData: function() {
+      if (!this.link.latencyTimestamp) {
+        return false;
+      }
+      elapsedMillis = Date.now() - new Date(this.link.latencyTimestamp);
+      return elapsedMillis <= maxClockSkewMillis;
+    },
+
+    getText: function() {
+      return `${this.link.latency} ms`;
```
safchain: should be divided by 1000000 as you add 'ms'

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+      if (!a.Capture) {
+        return;
+      }
+      if (!b.Capture) {
+        return;
+      }
+
+      const limit = 10;
+      const anyNode = 'Within("'+a.TID+'", "'+b.TID+'")';
+      const query = 'G.Flows().Has("NodeTID", '+anyNode+').Limit('+limit+')';
```
safchain: not sure this will be accurate to compare 'start' and not RTT

safchain:
need to be rebased

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+    flowKey: function(a) {
+      return `TrackingID:${a.TrackingID},ANodeTID:${a.ANodeTID},BNodeTID:${a.BNodeTID}`;
+    },
+
+    linklabel: function(a, b) {
+      return `${a.Name}-->${b.Name}`;
+    },
+
+    flowCompare: function(a, b) {
+      keya = this.flowKey(a);
```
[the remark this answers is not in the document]
hunchback: const

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+      let prev, curr;
+      for (i in flows) {
+        prev = curr;
+        curr = flows[i];
+
+        if (!prev) {
+          continue;
+        }
+
+        if (self.flowKey(curr) != self.flowKey(prev)) {
```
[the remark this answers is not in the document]
hunchback: changed to `this.flowKey(..)`

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+          continue;
+        }
+
+        latency = Math.abs(curr.Start - prev.Start);
```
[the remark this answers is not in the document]
hunchback: moved to 114

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+        return;
+      }
+      if (!b.Capture) {
+        return;
+      }
+
+      const limit = 10;
+      const anyNode = 'Within("'+a.TID+'", "'+b.TID+'")';
+      const query = 'G.Flows().Has("NodeTID", '+anyNode+').Limit('+limit+')';
+      console.log(`${prefix}: ${query}`)
```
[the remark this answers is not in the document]
hunchback: removed

statics/js/components/graph-layout.js
hunchback: removed

statics/js/components/graph-layout.js
hunchback: ok have changed to `RTT` and set timestamp using `Last`

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+
+    flowCompare: function(a, b) {
+      keya = this.flowKey(a);
+      keyb = this.flowKey(b);
+      return keya.localeCompare(keyb);
+    },
+
+    updateLatency: function(flows) {
+      let prev, curr;
+      for (i in flows) {
```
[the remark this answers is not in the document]
hunchback: ok

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+      console.log(`${prefix}: ${query}`)
+
+      this.$topologyQuery(query)
+        .then(function(flows) {
+          if (flows.length === 0) {
+            return;
+          }
+          flows = flows.sort(self.flowCompare);
+          if (self.updateLatency(flows)) {
+            console.log(`${prefix}: ${self.getText()}`);
```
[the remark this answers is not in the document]
hunchback: removed

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,187 @@
...
+        .fail(function(error) {
+          console.log(error);
+        });
+    },
+
+    hasData: function() {
+      if (!this.link.latencyTimestamp) {
+        return false;
+      }
+      elapsedMillis = Date.now() - new Date(this.link.latencyTimestamp);
```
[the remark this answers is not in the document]
hunchback: const

statics/js/components/graph-layout.js
hunchback: used `Math.abs(curr.RTT-prev.RTT)/1000`

hunchback:
@lebauce - I have addressed all outstanding issues

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,184 @@
...
+      const b = link.target.metadata;
+
+      if (!a.Capture) {
+        return;
+      }
+      if (!b.Capture) {
+        return;
+      }
+
+      const limit = 10;
```
nplanel: why 10, we need only 1 flow

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,184 @@
...
+        if (!prev) {
+          continue;
+        }
+
+        if (this.flowKey(curr) != this.flowKey(prev)) {
+          continue;
+        }
+
+        link.latencyTimestamp = Math.max(curr.Last, prev.Last);
+        link.latency = Math.abs(curr.RTT - prev.RTT) / 1000;
```
nplanel: should be "Math.abs(curr.RTT - prev.RTT) / 1000000" 
as RTT is nanoseconds and you display latency in milliseconds

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,184 @@
...
+    },
+
+    flowKey: function(a) {
+      return `TrackingID:${a.TrackingID},ANodeTID:${a.ANodeTID},BNodeTID:${a.BNodeTID}`;
+    },
+
+    flowCompare: function(a, b) {
+      const keya = this.flowKey(a);
+      const keyb = this.flowKey(b);
+      return keya.localeCompare(keyb);
```
nplanel: keya == keyb

statics/js/components/graph-layout.js
```diff
@@ -75,6 +253,22 @@ var TopologyGraphLayout = function(vm, selector) {
 
 TopologyGraphLayout.prototype = {
 
+  linkLabelFactory: function(link) {
```
safchain: can you rename the "updateBandwidth" function to something else as it handles two things now latency + bandwidth

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,184 @@
...
+      const limit = 10;
+      const anyNode = 'Within("'+a.TID+'", "'+b.TID+'")';
+      const query = 'G.Flows().Has("NodeTID", '+anyNode+').Has("RTT", NE(0)).Limit('+limit+')';
+
+      this.$topologyQuery(query)
+        .then(function(flows) {
+          if (flows.length === 0) {
+            return;
+          }
+          flows = flows.sort(self.flowCompare);
```
nplanel: you should do 2 gremlin request 
1st: flow = G.Flows().Has("NodeTID", '+anyNode+').Has("RTT", NE(0)).Limit(1)
2nd : flowNodes = G.Flows().Has("TrackingID", flow.TrackingID)

fill a hash with (foreach flowNodes)
RTTNodes[flowNodes.NoteTID] = flowNodes.RTT

pass the link, flow and the hash to your function updateLatency()

Your current implementation will not work in production, as it's currently bet that a same flow will be captured on nodeA and nodeB within your limit (10)

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,184 @@
...
+
+        if (!prev) {
+          continue;
+        }
+
+        if (this.flowKey(curr) != this.flowKey(prev)) {
+          continue;
+        }
+
+        link.latencyTimestamp = Math.max(curr.Last, prev.Last);
```
nplanel: you should use Date.Now() as comparing packet Last (capture) clock with Now (local browser clock) is not correct

statics/js/components/graph-layout.js
hunchback: changing to `updateLinkLabel`

statics/js/components/graph-layout.js
hunchback: fixed

statics/js/components/graph-layout.js
hunchback: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
returns 0, <0 or >0 as needed by for the compare function.

The `==` operator returns `true` or `false`

statics/js/components/graph-layout.js
hunchback: But 1st step may return a flow which only passed via A (but not B) or B (but not A) - as a result we will not be able (in 2nd step) to get RTT data for both A and B.

statics/js/components/graph-layout.js
hunchback: If the flow we based our calculations ran 1 hour ago or 1 day ago (as would be apparent from `curr.Last` and `prev.Last`) do we still want to use it? No! So to mitigate this I *do* want to check the time recorded during capture and if flow is within a set window (using 5 minutes skew time as used for bandwidth calculation) - then use it.

The 5 minute window was intentionally taken so as to also mitigate time skew between the agent and the browser host.

statics/js/components/graph-layout.js
safchain: @hunchback  I would rather do a selection of the flows seen since a period (1 minute), so than we will not miss flow according to what Nicolas said and it will solve what you reported.

hunchback: As there is possible clock skew I have added Context("the last 5 minutes") to the query

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,186 @@
...
+
+  methods: {
+    setup: function(topology) {
+      this.active = 0;
+      this.warning = 10
+      this.alert = 100;
+    },
+
+    flowKey: function(a) {
+      return `TrackingID:${a.TrackingID},ANodeTID:${a.ANodeTID},BNodeTID:${a.BNodeTID}`;
```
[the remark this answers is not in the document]
hunchback: correct have changed key to `a.TrackingID`

statics/js/components/graph-layout.js
```diff
@@ -132,7 +132,9 @@ var LinkLabelLatency = Vue.extend({
       }
 
       const limit = 10;
+      const context = Date.now() - maxClockSkewMillis;
       const query = `G.Flows()` +
+        `.Context(${context})` +
```
safchain: It won't fix the issue as using limit can remove the flow that we expect. My suggestion was to use Flow.Start since really few minutes removing limit. Like Flows().Has('Start', GT())

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,188 @@
...
+      this.warning = 10
+      this.alert = 100;
+    },
+
+    flowKey: function(a) {
+      return ${a.TrackingID};
+    },
+
+    flowCompare: function(a, b) {
+      const keya = this.flowKey(a);
```
safchain: no need to keep the flowKey function anymore

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,188 @@
...
+      if (!a.Capture) {
+        return;
+      }
+      if (!b.Capture) {
+        return;
+      }
+
+      const limit = 10;
+      const context = Date.now() - maxClockSkewMillis;
+      const query = `G.Flows()` +
```
safchain: It won't fix the issue as using limit can remove the flow that we expect. My suggestion was to use Flow.Start since really few minutes removing limit. Like Flows().Has('Start', GT())

hunchback: I still want some limitation as there may be cases in which I had a burst of activity in the last minute(s) ... so I would want two restriction elements in place:
- limit flow count - can be 10, 100, 1000 but we need a set limit to protect the UI
- restrict to relevant time slice - you proposed `.Host('Start' ...)` I am using `Context(..)` is this reasonable (even if not equivalent)?

statics/js/components/graph-layout.js
```diff
@@ -131,21 +116,20 @@ var LinkLabelLatency = Vue.extend({
...
-
-      this.$topologyQuery(query)
+      this.flowQuery(a.TID, undefined, 10)
         .then(function(flows) {
-          if (flows.length === 0) {
-            return;
-          }
-          flows = flows.sort(self.flowCompare);
-          self.updateLatency(link, flows);
+          for (let i in flows) {
```
safchain: @hunchback I would rather do for the first request Flows().Has('NodeTID', A).Sort().Limit(10), so not using context. It will return the most recent flows.

hunchback: @safchain - ok I have replaced Context() with Sort()

hunchback:
I have combined all changed into: https://github.com/skydive-project/skydive/pull/826/commits/ad8c833611d68a2052fde4c7066cc832fe11616b

hunchback:
@safchain failed due to:

```
16:16:29.957 --- FAIL: TestPacketInjectionCapture (470.74s)
16:16:29.957 	selenium_test.go:175: By default 7 rows should be return, but got: 8
16:16:29.957 FAIL
16:16:29.972 make[1]: *** [Makefile:210: test.functionals.run] Error 1
16:16:29.972 make[1]: Leaving directory '/var/lib/jenkins/workspace/skydive-selenium-tests/src/github.com/skydive-project/skydive'
16:16:29.972 make: *** [Makefile:223: test.functionals] Error 2
16:16:30.412 Build step 'Execute shell' marked build as failure
16:16:30.602 Archiving artifacts
16:16:31.264 Setting status of ad8c833611d68a2052fde4c7066cc832fe11616b to FAILURE with url https://ci.skydive.network/job/skydive-selenium-tests/2145/ and message: 'Build finished. '
16:16:31.265 Using context: skydive-selenium-tests
16:16:31.616 Finished: FAILURE
```
Is this related to my change?

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,173 @@
...
+        return;
+      }
+
+      this.flowQuery(a.TID, undefined, 10)
+        .then(function(flows) {
+          for (let i in flows) {
+            const aFlow = flows[i];
+            self.flowQuery(b.TID, aFlow.TrackingID, 1)
+              .then(function(flows) {
+                for (let i in flows) {
```
nplanel: you don't need to iterate on all the flows, just test if len(flows) == 0; return

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,173 @@
...
+                  const bFlow = flows[i];
+                  self.updateLatency(link, aFlow, bFlow);
```
nplanel: once we found a same flow (Tracking) with that pass in 2 captures points, we don't need to iterate on other flows (line 121 : for (let i in flows)

statics/js/components/graph-layout.js
```diff
@@ -1,6 +1,173 @@
...
+      }
+      if (!b.Capture) {
+        return;
+      }
+
+      this.flowQuery(a.TID, undefined, 10)
+        .then(function(flows) {
+          for (let i in flows) {
+            const aFlow = flows[i];
+            self.flowQuery(b.TID, aFlow.TrackingID, 1)
```
nplanel: for readability would be better to have 2 distinct query, like flowQueryNodeTID() and flowQueryTrackingID()

statics/js/components/graph-layout.js
hunchback: ok

statics/js/components/graph-layout.js
hunchback: ok, will fix

statics/js/components/graph-layout.js
hunchback: ok, will fix

PR #831, opened 2018-03-05
ui: fix unhighlight issue of nodes

masco-enovance:
when moving quickly on highlight link
node is not unhighlighted


masco-enovance:
run skydive-selenium-tests

PR #836, opened 2018-03-06
flow: search for flow using only NodeTID instead of NodeTID/ANodeTID/…

safchain:
…BNodeTID

Only returns flow where they have been captured

### DIFF
--- a/statics/js/components/topology.js
+++ b/statics/js/components/topology.js
@@ -358,7 +358,7 @@ var TopologyComponent = {
 
     currentNodeFlowsQuery: function() {
       if (this.currentNode && this.currentNode.isCaptureAllowed())
-        return "G.V('" + this.currentNode.id + "').Flows().Dedup().Sort()";
+        return "G.Flows().Has('NodeTID', '" + this.currentNode.metadata.TID + "').Sort()";
       return "";
     },
 
--- a/topology/graph/elasticsearch.go
+++ b/topology/graph/elasticsearch.go
@@ -272,7 +272,7 @@ func (b *ElasticSearchBackend) NodeDeleted(n *Node) bool {
 func (b *ElasticSearchBackend) GetNode(i Identifier, t *common.TimeSlice) []*Node {
 	return b.SearchNodes(&TimedSearchQuery{
 		SearchQuery: filters.SearchQuery{
-			Filter: filters.NewFilterForIds([]string{string(i)}, "ID"),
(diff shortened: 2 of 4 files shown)


topology/graph/elasticsearch.go
```diff
@@ -320,7 +320,7 @@ func (b *ElasticSearchBackend) EdgeDeleted(e *Edge) bool {
 func (b *ElasticSearchBackend) GetEdge(i Identifier, t *common.TimeSlice) []*Edge {
 	return b.SearchEdges(&TimedSearchQuery{
 		SearchQuery: filters.SearchQuery{
-			Filter: filters.NewFilterForIds([]string{string(i)}, "ID"),
+			Filter: filters.NewOrTermStringFilter([]string{string(i)}, "ID"),
```
lebauce: NewTermStringFilter("ID", string(i)

topology/graph/elasticsearch.go
```diff
@@ -272,7 +272,7 @@ func (b *ElasticSearchBackend) NodeDeleted(n *Node) bool {
 func (b *ElasticSearchBackend) GetNode(i Identifier, t *common.TimeSlice) []*Node {
 	return b.SearchNodes(&TimedSearchQuery{
 		SearchQuery: filters.SearchQuery{
-			Filter: filters.NewFilterForIds([]string{string(i)}, "ID"),
+			Filter: filters.NewOrTermStringFilter([]string{string(i)}, "ID"),
```
lebauce: NewTermStringFilter("ID", string(i)

statics/js/components/topology.js
```diff
@@ -358,7 +358,7 @@ var TopologyComponent = {
 
     currentNodeFlowsQuery: function() {
       if (this.currentNode && this.currentNode.isCaptureAllowed())
-        return "G.V('" + this.currentNode.id + "').Flows().Dedup().Sort()";
+        return "G.Flows().Has('NodeTID', '" + this.currentNode.metadata.TID + "').Sort()";
```
lebauce: Is it still required as you changed the "V().Flows()" behaviour ?

safchain: I think it still better to do it, as internally it will do the same, except that it will have to lookup for the node while here we already have it.

topology/graph/elasticsearch.go
safchain: good catch

lebauce:

PR #843, opened 2018-03-08
client: move topology deprecated note to the top level help

safchain:


safchain:

PR #851, opened 2018-03-12
tests: use packet injector instead of ping

lebauce:


lebauce:

PR #855, opened 2018-03-14
ui: get relative time from user for history

masco-enovance:


statics/js/components/topology.js
```diff
@@ -521,10 +554,24 @@ var TopologyComponent = {
...
+      var h = relTimeStr.match(/(\d+)\s*h/);
+      var m = relTimeStr.match(/(\d+)\s*m/);
+      var s = relTimeStr.match(/(\d+)\s*s/);
+      var totalTime = 0;
+      if (d) totalTime += parseInt(d[1]) * 86400;
+      if (h) totalTime += parseInt(h[1]) * 3600;
+      if (m) totalTime += parseInt(m[1]) * 60;
+      if (s) totalTime += parseInt(s[1]);
+      totalTime = totalTime * 1000;
+      currentTime = new Date().getTime();
```
safchain: var missing

statics/js/components/topology.js
```diff
@@ -521,10 +554,24 @@ var TopologyComponent = {
...
+      absTime = new Date(currentTime - totalTime);
```
safchain: same here

statics/js/components/topology.js
masco-enovance: done

statics/js/components/topology.js
masco-enovance: done

PR #857, opened 2018-03-14
ui: discover latency in very noisy environments

hunchback:
We are attempting to calculate the latency between A to B by finding a flow
which has been seen in both points. Due to current limitions in Gremlin we
are not able to do so in a single query forcing us to:
- 1st find a flow which intersects with A (using a small queue)
- 2nd check if the discovered flow intersects with B
If the system is wieghted down with many unique flows running from A to C we
may fill the queue in phase 1 with flows from which we can't deduce latency
(between A to  B). Such as a case in which we had HTTP traffic (A -> B) and
many many flows of DNS traffic (A -> C).

The new implementation uses a very large queue for 1st phase but then reduces
it down to a limitted set of unique flows: comparing ANodeTID, BNodeTID,
Application. As 2nd step is expenssive we only pass on a limitted list of 10
unique flows. So from the example above we would still see the HTTP traffic (A
-> B) becuase the DNS traffic (A -> C) would only use one spot in the list of
flows passed to the 2nd phase.

### DIFF
--- a/statics/js/components/graph-layout.js
+++ b/statics/js/components/graph-layout.js
@@ -93,22 +93,62 @@ var LinkLabelLatency = Vue.extend({
     },
 
     flowQuery: function(nodeTID, trackingID, limit) {
-      let query = `G.Flows()`;
-      query += `.Has("NodeTID", "${nodeTID}")`;
+      let has = `"NodeTID", ${nodeTID}`;
       if (typeof trackingID !== 'undefined') {
-        query += `.Has("TrackingID", "${trackingID}")`;
+        has += `"TrackingID", ${trackingID}`;
       }
-      query += `.Has("RTT", NE(0))`;
-      query += `.Sort().Limit(${limit})`;
+      has += `"RTT", NE(0)`;
+      let query = `G.Flows().Has(${has}).Sort().Limit(${limit})`;
       return this.$topologyQuery(query)
     },
 
     flowQueryByNodeTID: function(nodeTID, limit) {
-      return this.flowQuery(nodeTID, undefined, limit);
+      return this.flowQuery(`"${nodeTID}"`, undefined, limit);
     },
 
-    flowQueryByNodeTIDandTrackingID: function(nodeTID, trackingID) {
-      return this.flowQuery(nodeTID, trackingID, 1);
+    flowQueryByNodeTIDandTrackingID: function(nodeTID, flows) {
+      let anyTrackingID = 'Within(';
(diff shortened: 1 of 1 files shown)


safchain:
run cdd-overview

safchain:
ok to test

safchain:
rerun tests

safchain:
run cdd-overview

safchain:
rerun tests

safchain:
rerun tests

hunchback:
@safchain - note I made changes such that the 2nd step executes only a single query ... so:
1st step - one query
2nd step - one query (optional)

statics/js/components/graph-layout.js
```diff
@@ -94,21 +94,62 @@ var LinkLabelLatency = Vue.extend({
 
     flowQuery: function(nodeTID, trackingID, limit) {
       let query = `G.Flows()`;
-      query += `.Has("NodeTID", "${nodeTID}")`;
+      query += `.Has("NodeTID", ${nodeTID})`;
       if (typeof trackingID !== 'undefined') {
-        query += `.Has("TrackingID", "${trackingID}")`;
+        query += `.Has("TrackingID", ${trackingID})`;
```
safchain: Would be better to use one Has in term of performances

statics/js/components/graph-layout.js
```diff
@@ -94,21 +94,62 @@ var LinkLabelLatency = Vue.extend({
...
 
     flowQueryByNodeTID: function(nodeTID, limit) {
-      return this.flowQuery(nodeTID, undefined, limit);
+      return this.flowQuery(`"${nodeTID}"`, undefined, limit);
     },
 
-    flowQueryByNodeTIDandTrackingID: function(nodeTID, trackingID) {
-      return this.flowQuery(nodeTID, trackingID, 1);
+    flowQueryByNodeTIDandTrackingID: function(nodeTID, flows) {
+      anyTrackingID = 'Within(';
```
safchain: "let" missing

safchain:
rerun tests

statics/js/components/graph-layout.js
hunchback: ok, using one 'Has(...)'

statics/js/components/graph-layout.js
hunchback: added

safchain:
rerun tests

safchain:
run cdd-overview

PR #858, opened 2018-03-15
socketinfo: support UDP connections when not using eBPF

lebauce:


topology/probes/socketinfo/connection.go
```diff
@@ -105,9 +107,12 @@ type ConnectionCache struct {
 }
 
 // HashTuple computes a hash value for the srcAddr:srcPort dstAddr:dstPort tuple
```
nplanel: please change the comment

PR #862, opened 2018-03-15
Rolling es indices

sivanbhaim:
Roll Elasticseach indices when an entries counter limit or age limit is reached. Indices are deleted only after the number of indices limit is reached.
The index for topology and flows is split to allow separate behaviors.
Topology is dumped into the new index after a roll is executed.


safchain:
ok to test

safchain:
add to whitelist

hunchback:
skydive-go-fmt failed - you can fix this by running:
```
gofmt -s -w filename.go
```

hunchback:
skydive-functional-tests-backend-elasticsearch - please check why this is failing - it appears related to your fix:

```
18:49:48.754 --- FAIL: TestFlowMetricsStep (66.27s)
18:49:48.754 	tests.go:249: Removing existing captures
18:49:48.754 	tests.go:257: Creating captures
18:49:48.754 	tests.go:269: Executing setup commands
18:49:48.754 	tests.go:279: Checking captures are correctly set up
18:49:48.754 	tests.go:325: Executing settle function
18:49:48.754 	tests.go:343: Executing setup function
18:49:48.754 	tests.go:453: Running checks
18:49:48.754 	tests.go:477: Test failed: Could not find metrics (<nil>) for flows 9 flows:
18:49:48.754 		[
18:49:48.754 			{
18:49:48.754 				"UUID": "bbf6554b6220d3f5dcd50fe692f3789b236a03f5",
18:49:48.754 				"LayersPath": "Ethernet/IPv6/IPv6HopByHop/ICMPv6",
18:49:48.754 				"Application": "ICMPv6",
```

config/config.go
```diff
@@ -83,6 +83,9 @@ func init() {
 	cfg.SetDefault("analyzer.storage.bulk_insert", 100)
 	cfg.SetDefault("analyzer.storage.bulk_insert_deadline", 5)
 	cfg.SetDefault("analyzer.storage.max_flow_buffer_size", 100000)
+	cfg.SetDefault("analyzer.storage.index_entries_limit", -1)
+	cfg.SetDefault("analyzer.storage.index_age_limit", -1)
+	cfg.SetDefault("analyzer.storage.indices_to_keep", -1)
 	cfg.SetDefault("analyzer.topology.probes", []string{})
```
hunchback: nit: better to order these alphabetically:
- index_age_limit
- index_entries_limit
- indices_to_keep

config/config.go
```diff
@@ -148,6 +151,9 @@ func init() {
 	cfg.SetDefault("storage.elasticsearch.retry", 60)
 	cfg.SetDefault("storage.elasticsearch.bulk_maxdocs", 100)
 	cfg.SetDefault("storage.elasticsearch.bulk_maxdelay", 5)
+	cfg.SetDefault("storage.elasticsearch.index_entries_limit", -1)
+	cfg.SetDefault("storage.elasticsearch.index_age_limit", -1)
+	cfg.SetDefault("storage.elasticsearch.indices_to_keep", -1)
 	cfg.SetDefault("storage.orientdb.addr", "http://localhost:2480")
```
hunchback: nit: better to order these alphabetically:
- index_age_limit
- index_entries_limit
- indices_to_keep

contrib/snort/snortSkydive.go
```diff
@@ -123,7 +123,7 @@ func (sfe *SnortFlowEnhancer) insertElasticSearch(msg *snortMessage, f *flow.Flo
 		"Message":        msg.Message,
 		"Classification": msg.Classification,
 	}
-	if err := sfe.client.BulkIndex("snortMessage", "", snortMessage); err != nil {
+	if err, _ := sfe.client.BulkIndex("snortMessage", "", snortMessage); err != nil {
```
hunchback: nit: I think `err` works just as `err, _` but its' shorter

etc/skydive.yml.default
```diff
@@ -65,6 +65,17 @@ analyzer:
...
+    # If a limit is specified, when the index reaches it, it is rolled.
+    # index_entries_limit specifies the maximum number of entries allowed in an index.
+    # index_age_limit specifies the maximum age (in seconds) allowed for an index.
+    # For both limits, a value of -1 specifies that there is no limitation.
+    # index_entries_limit: -1
+    # index_age_limit: -1
+
+    # The number of indices to keep before deleting.
+    # A value of -1 specifies no limit (i.e. indices will never be deleted)
+    # indices_to_keep: -1
```
hunchback: as above - this section should probably be removed

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -170,16 +171,28 @@ func (c *ElasticSearchStorage) StoreFlows(flows []*flow.Flow) error {
...
-		if err := c.client.BulkIndex("flow", f.UUID, f); err != nil {
+		err, shouldRoll := c.client.BulkIndex("flow", f.UUID, f)
+		if err != nil {
 			logging.GetLogger().Errorf("Error while indexing: %s", err.Error())
 			continue
 		}
+		if shouldRoll {
+			if err := c.client.RollIndex(); err != nil {
+				logging.GetLogger().Errorf("Error while rolling index: %s", err.Error())
+			}
```
hunchback: I guess to are missing a `continue` here

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -170,16 +171,28 @@ func (c *ElasticSearchStorage) StoreFlows(flows []*flow.Flow) error {
...
 	for _, f := range flows {
-		if err := c.client.BulkIndex("flow", f.UUID, f); err != nil {
+		err, shouldRoll := c.client.BulkIndex("flow", f.UUID, f)
+		if err != nil {
 			logging.GetLogger().Errorf("Error while indexing: %s", err.Error())
 			continue
 		}
+		if shouldRoll {
+			if err := c.client.RollIndex(); err != nil {
+				logging.GetLogger().Errorf("Error while rolling index: %s", err.Error())
```
hunchback: continue

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -170,16 +171,28 @@ func (c *ElasticSearchStorage) StoreFlows(flows []*flow.Flow) error {
...
+			err, shouldRoll := c.client.BulkIndexChild("metric", f.UUID, "", f.LastUpdateMetric)
+			if  err != nil {
 				logging.GetLogger().Errorf("Error while indexing: %s", err.Error())
 				continue
 			}
+			if shouldRoll {
+				if err := c.client.RollIndex(); err != nil {
+					logging.GetLogger().Errorf("Error while rolling index: %s", err.Error())
+				}
+			}
```
hunchback: this entire block is a code duplication - 187..195 is a duplication of 175..182

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -194,10 +207,17 @@ func (c *ElasticSearchStorage) StoreFlows(flows []*flow.Flow) error {
...
+			err, shouldRoll := c.client.BulkIndexChild("rawpacket", f.UUID, "", rawpacket)
+			if err != nil {
 				logging.GetLogger().Errorf("Error while indexing: %s", err.Error())
 				continue
 			}
+			if shouldRoll {
+				if err := c.client.RollIndex(); err != nil {
+					logging.GetLogger().Errorf("Error while rolling index: %s", err.Error())
+				}
+			}
```
hunchback: and again 211..219 is a code duplication

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -432,10 +452,14 @@ func (c *ElasticSearchStorage) SearchFlows(fsq filters.SearchQuery) (*flow.FlowS
 
 // Start the Database client
 func (c *ElasticSearchStorage) Start() {
-	go c.client.Start([]map[string][]byte{
+	entriesLimit := config.GetInt("analyzer.storage.index_entries_limit")
+	ageLimit := config.GetInt("analyzer.storage.index_age_limit")
+	indicesLimit := config.GetInt("analyzer.storage.indices_to_keep")
+	go c.client.Start("flows", []map[string][]byte{
```
hunchback: again - better is storage module is not using configuration from the analyzer section

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -432,10 +452,14 @@ func (c *ElasticSearchStorage) SearchFlows(fsq filters.SearchQuery) (*flow.FlowS
...
 		{"metric": []byte(metricMapping)},
 		{"rawpacket": []byte(rawPacketMapping)},
 		{"flow": []byte(flowMapping)}},
+		entriesLimit, ageLimit, indicesLimit,
```
hunchback: you probably what it create a struct which holds all limits - something like:

```
type StorageLimits struct {
        EntriesLimit int
        AgeLimit int
        IndicesLimit
}
```

storage/elasticsearch/client.go
```diff
@@ -39,26 +39,45 @@ import (
...
 	UpdateWithPartialDoc(obj string, id string, data interface{}) error
 	BulkUpdateWithPartialDoc(obj string, id string, data interface{}) error
 	Get(obj string, id string) (elastigo.BaseResponse, error)
 	Delete(obj string, id string) (elastigo.BaseResponse, error)
 	BulkDelete(obj string, id string)
-	Search(obj string, query string) (elastigo.SearchResult, error)
-	Start(mappings []map[string][]byte)
+	Search(obj string, query string, index string) (elastigo.SearchResult, error)
+	Start(name string, mappings []map[string][]byte, entriesLimit int, ageLimit int, indicesLimit int)
+	GetIndexAlias() string
```
hunchback: Can this method not return an error?

```
GetIndexAlias() (string, error)
```

storage/elasticsearch/client.go
```diff
@@ -39,26 +39,45 @@ import (
...
+	GetIndexAllAlias() string
```
hunchback:
```
GetIndexAllAlias() (string, error)
```

storage/elasticsearch/client.go
```diff
@@ -39,26 +39,45 @@ import (
...
+type ElasticIndex struct {
+	entriesCounter 	int
+	mappings 	[]map[string][]byte
+	name		string
+	path		string
+	timeCreated	time.Time
+	entriesLimit	int
+	ageLimit	int
+	indicesLimit	int
+	lock		sync.Mutex
```
hunchback: nicer to embed the lock inside the struct:

```
type ElasticIndex struct {
        ...
        sync.Mutex
}
```

storage/elasticsearch/client.go
```diff
@@ -68,6 +87,8 @@ type ElasticSearchClient struct {
 	started    atomic.Value
 	quit       chan bool
 	wg         sync.WaitGroup
+	index      *ElasticIndex
+
```
hunchback: please remove space

storage/elasticsearch/client.go
```diff
@@ -87,7 +108,35 @@ func (c *ElasticSearchClient) request(method string, path string, query string,
 	return req.Do(&response)
 }
 
+func (e *ElasticIndex) increaseEntries() {
+	e.lock.Lock()
```
hunchback: with embedding the lock code would be reduced to:

```
e.Lock()
```

storage/elasticsearch/client.go
```diff
@@ -87,7 +108,35 @@ func (c *ElasticSearchClient) request(method string, path string, query string,
...
+	e.entriesCounter += 1
+	e.lock.Unlock()
+}
+
+func getIndexPath(name string) string {
+	t := time.Now()
+	return fmt.Sprintf("skydive_%s_v%d_%d-%02d-%02d_%02d-%02d-%02d",
+		name, indexVersion, t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), t.Second())
```
hunchback: I think you can use:

```
return fmt.Sprintf("skydive_%s_v%d_%s", name, indexVersion, t.Format(time.RFC3339)))
```

storage/elasticsearch/client.go
```diff
@@ -87,7 +108,35 @@ func (c *ElasticSearchClient) request(method string, path string, query string,
...
+	return fmt.Sprintf("%s_%s", indexPrefix, c.index.name)
+}
+
+func (c *ElasticSearchClient) GetIndexAllAlias() string {
+	return fmt.Sprintf("%s_%s", indexPrefix, indexAllAlias)
+}
+
+func (c *ElasticSearchClient) countEntries() int {
+	curEntriesCount, _ := c.connection.Count(c.index.path, "", nil, "")
+	logging.GetLogger().Infof("#####       %s real entries in %s is %d", c.index.name, c.index.path, curEntriesCount.Count)
```
hunchback: Please remove the `#####     ` prefix and I guess this should be a `Debugf` rather

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
+	return nil, c.shouldRollIndex()
 }
 
 // BulkIndexChild index a while object with the indexer
-func (c *ElasticSearchClient) BulkIndexChild(obj string, parent string, id string, data interface{}) error {
-	return c.indexer.Index("skydive", obj, id, parent, "", nil, data)
+func (c *ElasticSearchClient) BulkIndexChild(obj string, parent string, id string, data interface{}) (error, bool) {
+	c.index.lock.Lock()
+	if err := c.indexer.Index(c.GetIndexAlias(), obj, id, parent, "", nil, data); err != nil {
+		c.index.lock.Unlock()
```
hunchback: Rather add a defer c.index.lock.Unlock() right after .Lock()

storage/elasticsearch/client.go
```diff
@@ -118,31 +169,60 @@ func (c *ElasticSearchClient) createAlias() error {
...
 	if err := c.createAlias(); err != nil {
+		logging.GetLogger().Errorf("Failed to create alias")
 		return err
 	}
 
 	c.indexer.Start()
 	c.started.Store(true)
 
-	logging.GetLogger().Infof("ElasticSearchStorage started")
+	logging.GetLogger().Infof("ElasticSearchStorage started with skydive index %s", c.index.name)
```
hunchback: Used c.GetIndexAlias() instead of hardcoding skydive

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
+		logging.GetLogger().Debugf("%s entries limit not set", c.index.name)
+		return false
+	}
+	logging.GetLogger().Debugf("%s entries counter is %d", c.index.name, c.index.entriesCounter)
+	if c.index.entriesCounter < c.index.entriesLimit {
+		return false
+	}
+	c.indexer.Flush()
+	time.Sleep(3 * time.Millisecond)
+
```
hunchback: Why do we this sleep - please add a comment into the code

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
+	return true
+}
+
+func (c *ElasticSearchClient) shouldRollIndexByAge() bool {
+	if c.index.ageLimit == -1 {
+		logging.GetLogger().Debugf("%s age limit not set", c.index.name)
+		return false
+	}
+	age := int(time.Now().Sub(c.index.timeCreated).Seconds())
+	logging.GetLogger().Debugf("%s age is %d", c.index.name, age)
```
hunchback: Add to the comment age time unit - seconds

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
+		logging.GetLogger().Debugf("Deleting index of %s: %s", c.index.name, esIndex.Name)
+		if _, err := c.connection.DeleteIndex(esIndex.Name); err != nil {
+			logging.GetLogger().Errorf("Error deleting index %s: %s", esIndex.Name, err.Error())
+		}
+	}
+}
+
+func (c *ElasticSearchClient) RollIndex() error {
+	c.indexer.Flush()
+	time.Sleep(3 * time.Millisecond)
```
hunchback: If flush always requires a delay then add it to Flush or add a utility function which does both Flush and Sleep.

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
+	logging.GetLogger().Infof("Rolling indices for %s", c.index.name)
+	c.index.lock.Lock()
+
+	if err := c.createIndex(""); err != nil {
+		c.index.lock.Unlock()
```
hunchback: Use defer rather

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
+		return err
+	}
+	if err := c.createAlias(); err != nil {
+		c.index.lock.Unlock()
```
hunchback: Defer

storage/elasticsearch/client.go
```diff
@@ -418,9 +612,12 @@ func NewElasticSearchClient(addr string, port string, maxConns int, retrySeconds
 		connection: c,
 		indexer:    indexer,
 		quit:       make(chan bool),
+		index: 	    nil,
 	}
 
 	client.started.Store(false)
+
+
```
hunchback: Don't need two spaces ... not even sure we need one

storage/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,202 @@
...
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+)
+
+const testMapping = `
+{
```
hunchback: Nit: i would rather bring the { one line up

storage/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,202 @@
...
+					"index":      "not_analyzed",
+					"doc_values": false
+				}
+			}
+		}
+	]
+}`
+
+func (c *ElasticSearchClient) indexEntry(id int) (error, bool) {
+	return c.Index("test_type", fmt.Sprintf("id%d", id), "{\"key\": \"val\"}")
```
hunchback: If you use back-ticks the no need to escape quotes

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
+	return nil, c.shouldRollIndex()
 }
 
 // BulkIndex returns the bulk index from the indexer
-func (c *ElasticSearchClient) BulkIndex(obj string, id string, data interface{}) error {
-	return c.indexer.Index("skydive", obj, id, "", "", nil, data)
+func (c *ElasticSearchClient) BulkIndex(obj string, id string, data interface{}) (error, bool) {
+	c.index.lock.Lock()
+	if err := c.indexer.Index(c.GetIndexAlias(), obj, id, "", "", nil, data); err != nil {
+		c.index.lock.Unlock()
```
hunchback: Use defer

storage/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (c *ElasticSearchClient) cleanupIndices(name string) error {
+	if _, err := c.connection.DeleteIndex(fmt.Sprintf("%s_%s*", indexPrefix, name)); err != nil {
+		return errors.New(fmt.Sprintf("Failed to clear test indices: %s", err.Error()))
```
hunchback: fmt.Errorf() can replace errors.New(fmt.Sprintf())

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
+		return err, false
+	}
+	c.index.lock.Unlock()
```
hunchback: Use defer

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
 
 // IndexChild index a child object
-func (c *ElasticSearchClient) IndexChild(obj string, parent string, id string, data interface{}) error {
-	_, err := c.connection.IndexWithParameters("skydive", obj, id, parent, 0, "", "", "", 0, "", "", false, nil, data)
-	return err
+func (c *ElasticSearchClient) IndexChild(obj string, parent string, id string, data interface{}) (error, bool) {
+	c.index.lock.Lock()
+	_, err := c.connection.IndexWithParameters(c.GetIndexAlias(), obj, id, parent, 0, "", "", "", 0, "", "", false, nil, data)
+	if err != nil {
+		c.index.lock.Unlock()
```
hunchback: Use defer

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
+		return err, false
+	}
+	c.index.lock.Unlock()
```
hunchback: Use defer

storage/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,202 @@
...
+		t.Fatalf("Index should have rolled after %d entries", entriesLimit)
+	}
+
+	if err := client.cleanupIndices(name); err != nil {
+		t.Fatalf(err.Error())
+	}
+}
+
+// test rolling elasticsearch indices based on age limit
+func TestElasticsearchShouldRollByAge(t *testing.T) {
```
hunchback: Most of the code here is a duplicate from tge test above - please write utility function also for your test code

storage/elasticsearch/client.go
```diff
@@ -282,70 +362,184 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
...
 // BulkIndexChild index a while object with the indexer
-func (c *ElasticSearchClient) BulkIndexChild(obj string, parent string, id string, data interface{}) error {
-	return c.indexer.Index("skydive", obj, id, parent, "", nil, data)
+func (c *ElasticSearchClient) BulkIndexChild(obj string, parent string, id string, data interface{}) (error, bool) {
+	c.index.lock.Lock()
+	if err := c.indexer.Index(c.GetIndexAlias(), obj, id, parent, "", nil, data); err != nil {
+		c.index.lock.Unlock()
+		return err, false
+	}
+	c.index.lock.Unlock()
```
hunchback: Use defer

topology/graph/elasticsearch.go
```diff
@@ -446,7 +485,7 @@ func (b *ElasticSearchBackend) GetEdges(t GraphContext, m GraphElementMatcher) [
 		}
 		filter = f
 	}
-
+
```
hunchback: Remove tab

topology/graph/elasticsearch.go
```diff
@@ -250,17 +255,65 @@ func (b *ElasticSearchBackend) NodeDeleted(n *Node) bool {
...
+	}
+
+	return nodes
+}
+
+func (b *ElasticSearchBackend) rollAndDumpTopology() error {
+	nodes := b.GetNodes(GraphContext{nil, false}, nil)
+	edges := b.GetEdges(GraphContext{nil, false}, nil)
+
+	if err := b.client.RollIndex(); err != nil {
```
safchain: First thanks for addressing this feature. 

I'm not sure about the approach of creating a new index, then recreating nodes/edges because there is no guaranty to not have "partial update" in the client chan for nodes/edges that won't be present yet. I tend to think that it would be better to create "archived indexes" moving "archived records" so that we don't have to block anything neither having to take care about consistency. It could even be a new skydive tool or command line option ?

eranra: The entire point with this PR is to allow us to delete indexes, so that we can limit the amount of storage taken by elastic search over time. I am not an expert in elastic, so don't know what "archived indexes" will provide, in any even we need to delete some indexes over time. Since there is strong reliance on history in nodes, I do not see how  "archived records" helps. The idea was that each new index is a "snapshot" standalone point , one that doesn't require historical info at all. If you can elaborate on a different approach to solve it will be good.

safchain: my point was to move from the current index records marked as "archived" with the "archivedAt" field set to another index. The indexes where we put the "archived record" can be deleted according to a retain policy. 

My concern with the current approach is to have "partial update" in the ES client chan modifying record not already re-inserted. Maybe I missed something here, but it seems to me that the patch creates a new index and then copy active nodes from the previous index to the new one ? if so one can updates between index creation and nodes insertions ?

hunchback:
@eranra @safchain - so based on comments above I would propose the following:

split task into three distinct efforts:

(1) split flows and topology indices (including configuration)

(2) flows - implement a rolling index:
     - skydive_flows_v11_2017.04.01 <--- delete old indexes from here
    - ...
     - skydive_flows_v11_2018.03.26
     - skydive_flows_v11_2018.03.27
     - skydive_flows_v11_2018.03.28 <--- write via skydive_flows (alias to live index)
   performing CRUD operations on flows:
     - CREATE- add new to skydive_flows
     - READ - read object from *all* skydive_flows_v11_* indecis
     - UPDATE - not used (@safchain - please confirm)
     - DELETE - not used (@safchain - please confirm)

(3) topology - implement a rolling index:
     - skydive_topology_v11 <--- skydive_topology (alias to live index)
     - skydive_topology_archive_v11_2017.04.01  <--- delete old index from here
     - ...
     - skydive_topology_archive_v11_2018.03.26
     - skydive_topology_archive_v11_2018.03.27
     - skydive_topology_archive_v11_2018.03.28 <--- write via skydive_topology_archive
   performing CRUD operations on topology:
     - CREATE- add new to skydive_topolog
     - READ - read object from skydive_topology
     - UPDATE - add new to skydive_topology; move old from skydive_topology to skydive_topology_archive (no need for archivedAt field anymore)
     - DELETE - delete object from skydive_topology

safchain:
Replaced by #940

PR #863, opened 2018-03-15
Map as envvars

lebauce:


eranra:
Hi,

I tried to execute with several options but all **_did not work,_**  I get in the UI a string:

export SKYDIVE_UI_TOPOLOGY_FAVORITES="{\"Infrastructure\":\"G.V()\"}"
export SKYDIVE_UI_TOPOLOGY_FAVORITES={'Infrastructure':'G.V()'}
export SKYDIVE_UI_TOPOLOGY_FAVORITES="{'Infrastructure':'G.V()'}"

(I also had to do export SKYDIVE_ANALYZER_LISTEN="0.0.0.0:8082", but this is not relevant specifically for this)

In the UI I always get a string and not json struct 

image

((Also, I talked with Masco, and maybe the more aligned solution is to consolidate with the code in lines 502-504)) so that we will have the loop written once, and both options localstorage and config will use exactly the same code.

lebauce:
I forgot to mention that we need to wait for https://github.com/spf13/viper/pull/468/files to be merged for it to work (feel free to +1 this PR :-) )

eranra:
Hi @lebauce, I see that you forked Viper and plan to use the fork. Is that a good idea, I see changes in the Viper community code (16 days ago) ... using a fork we will not get updates from the Viper community. I see the value of that, but this also adds risk for bugs.

lebauce:
@eranra I'm confident that the PR will be merged. It seems the maintainers are a bit busy. Besides, we already closely look at the Viper development so we can still rebase periodically as it's only a 2 lines patch

PR #867, opened 2018-03-16
flows : introduce Group return flows that have the same matching field

nplanel:


nplanel:
run skydive-selenium-tests

nplanel:

nplanel:
run tests

nplanel:
run skydive-selenium-tests

flow/set_test.go
```diff
@@ -114,3 +111,116 @@ func TestMergeDedup(t *testing.T) {
...
+			{TrackingID: "bbb", NodeTID: "111", Start: 8, Last: 9, Metric: &FlowMetric{}},
+			{TrackingID: "aaa", NodeTID: "222", Start: 10, Last: 11, Metric: &FlowMetric{}},
+			{TrackingID: "ccc", NodeTID: "333", Start: 12, Last: 13, Metric: &FlowMetric{}},
+		},
+	}
+
+	expected := FlowSet{
+		Flows: []*Flow{
+			{TrackingID: "aaa", NodeTID: "111", Start: 6, Last: 7, Metric: &FlowMetric{}},
+			{TrackingID: "aaa", NodeTID: "222", Start: 10, Last: 11, Metric: &FlowMetric{}},
```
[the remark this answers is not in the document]
nplanel: no as it's DedupBy: "NodeTID"

tests/flow_test.go
```diff
@@ -1674,6 +1674,18 @@ func TestFlowsWithShortestPath(t *testing.T) {
 				return fmt.Errorf("Expected one flow, got %+v", flows)
 			}
 
+			gremlin = g.G.Context(c.time).Flows().Has("LayersPath", "Ethernet/IPv4/ICMPv4").Has("Network", "169.254.37.33").Group("TrackingID")
+			flows, err = c.gh.GetFlows(gremlin)
+			if err != nil {
+				return err
+			}
+			if len(flows) != 7 {
```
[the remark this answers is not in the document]
nplanel: when I dumped the nodeID there was the 5 captures + 2 captures on mirroring interfaces

flow/set.go
```diff
@@ -252,20 +247,47 @@ func (fs *FlowSet) Dedup(field string) error {
...
+	if err := dedupFlows.Flows(fs.Flows); err != nil {
+		return err
+	}
+	fs.Flows = deduped
+
+	return nil
+}
+
+// Group return only the flows that have a matching field in a FlowSet
+func (fs *FlowSet) Group(field string) error {
```
safchain: Looks like more a "Sort" than a "Group". I would rather expect an output like this, even if I don't have a strong opinion here.

```
<GroupBy ID1>: [
 Flow{}
 Flow{}
] ,
<GroupBy ID2>: [
 Flow{}
 Flow{}
],
```

flow/set.go
```diff
@@ -252,20 +247,47 @@ func (fs *FlowSet) Dedup(field string) error {
...
+			sortedFields = append(sortedFields, fieldValue)
 		}
+		group[fieldValue] = append(group[fieldValue], flow)
 	}
-	fs.Flows = deduped
 
+	var groupFlows []*Flow
+	for _, field := range sortedFields {
+		flows := group[field]
+		if len(flows) > 1 {
```
safchain: Why > 1 ? I don't understand the purpose of this, can you elaborate on why you keep only flow having the "groupby" seen at least twice ?

flow/set.go
nplanel: Group is looks like a Sort but we don't report single/unique flow, as we only want flows that have a common/same matching field
Think about group like a union, but union is not the good word in gremlin syntax

Report the key (grouped by) like Sockets, have no opinion too on that, was just implementing as simple I can for the gremlin latency query

flow/set.go
nplanel: see above

flow/set.go
nplanel: futhermore, that why it's only applied after merge

nplanel: we could Group() do output format as you describe (with unqiue flow) and a Group(morethan=1) report that remove unique flow, What do you think ?

safchain: mmm, I think "Group" means group not filtering. I would just do the group stuff and introduce a filter later based on the result of group.

nplanel:

nplanel:

nplanel:

gremlin/traversal/flows.go
```diff
@@ -886,6 +926,9 @@ func (s *FlowGremlinTraversalStep) Exec(last traversal.GraphTraversalStep) (trav
 		if s.rawpacketsNextStep {
 			return &FlowTraversalStep{GraphTraversal: graphTraversal, Storage: s.Storage, flowSearchQuery: flowSearchQuery}, nil
 		}
+		if s.groupNextStep {
```
safchain: not sure to understand how it could work as the flowset is filled line 933 ? We skip for metrics and rawpackets because the datastore request will be made there but in your case I don't see where the request will be made ?
I think it doesn't work in history mode. I don't think we can reduce for group or you will have to implement to group thing in ES and OrientDB

gremlin/traversal/group.go
```diff
@@ -0,0 +1,145 @@
...
+
+// Exec Group step
+func (g *GroupGremlinTraversalStep) Exec(last traversal.GraphTraversalStep) (traversal.GraphTraversalStep, error) {
+	switch last.(type) {
+	case *FlowTraversalStep:
+		fs := last.(*FlowTraversalStep)
+		return fs.Group(g.StepContext), nil
+	case *GroupTraversalStep:
+		group := last.(*GroupTraversalStep)
+		return group.MoreThan(g.StepContext), nil
```
safchain: MoreThan should be an extension otherwise we don't know that the user called MoreThan, here we only know that thre previous step was group. If later we add more step after group we won't know what step to process. Aggregates is a good example

gremlin/traversal/group.go
```diff
@@ -0,0 +1,145 @@
...
+		return e.MoreThanToken, true
+	}
+	return traversal.IDENT, false
+}
+
+// ParseStep parse metrics step
+func (e *GroupTraversalExtension) ParseStep(t traversal.Token, p traversal.GremlinTraversalContext) (traversal.GremlinTraversalStep, error) {
+	switch t {
+	case e.MoreThanToken:
+		return &GroupGremlinTraversalStep{GremlinTraversalContext: p}, nil
```
safchain: here you should return a MoreThanGremlinTraversalStep, see my comment below

nplanel:

nplanel:

gremlin/traversal/group.go
```diff
@@ -0,0 +1,168 @@
...
+
+// GroupGremlinTraversalStep group flows step
+type GroupGremlinTraversalStep struct {
+	traversal.GremlinTraversalContext
+}
+
+// MoreThanGremlinTraversalStep group flows step
+type MoreThanGremlinTraversalStep struct {
+	traversal.GremlinTraversalContext
+	moreThan int
```
safchain: not used

nplanel:

nplanel:

nplanel:

PR #868, opened 2018-03-19
ui: improve bandwidth display

masco-enovance:
use source node metrics to display bnadwidth
if no target node metrics

### DIFF
--- a/statics/js/components/graph-layout.js
+++ b/statics/js/components/graph-layout.js
@@ -35,21 +35,26 @@ var LinkLabelBandwidth = Vue.extend({
     },
 
     updateData: function(link) {
-      if (!link.target.metadata.LastUpdateMetric) {
+      var metadata;
+      if (link.target.metadata.LastUpdateMetric) {
+        metadata = link.target.metadata;
+      } else if (link.source.metadata.LastUpdateMetric) {
+        metadata = link.source.metadata;
+      } else {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


statics/js/components/graph-layout.js
```diff
@@ -35,21 +35,24 @@ var LinkLabelBandwidth = Vue.extend({
...
         return;
       }
 
       const defaultBandwidthBaseline = 1024 * 1024 * 1024; // 1 gbps
 
       link.bandwidthBaseline = (this.topology.bandwidth.bandwidthThreshold === 'relative') ?
-        link.target.metadata.Speed || defaultBandwidthBaseline : 1;
-
-      link.bandwidthAbsolute = this.bandwidthFromMetrics(link.target.metadata.LastUpdateMetric);
+        link.target.metadata.LastUpdateMetric ? link.target.metadata.Speed || defaultBandwidthBaseline : link.source.metadata.Speed || defaultBandwidthBaseline : 1;
```
safchain: @hunchback ok with this ?

hunchback: Add comment - gist of it - extract metadata (from target or source) then simplify code - other than that logic seams ok

masco-enovance: yep, good idea to use an extracted metadata from source or target. will modify the code.

masco-enovance:
ok to test

PR #871, opened 2018-03-20
tests: openflow switched to use gremlin builder

hunchback:
just verify that openflow tests work as before


hunchback:
moved to tests-all-gremlin PR

PR #872, opened 2018-03-20
tests: converted api tests to use gremlin builder

hunchback:
just verify that api tests work as before


hunchback:
moved to tests-all-gremlin PR

PR #874, opened 2018-03-20
tests: convert ebpf tests to use gremlin

hunchback:
verify that flow ebpf tests still pass


hunchback:
moved to tests-all-gremlin

PR #875, opened 2018-03-20
tests: convert lxd to use gremlin builder

hunchback:
verify that lxd tests still pass


hunchback:
moved to tests-all-gremlin PR

PR #876, opened 2018-03-20
tests: convert docker to use gremlin builder

hunchback:
verify that docker tests still pass


hunchback:
moved to tests-all-gremlin PR

PR #877, opened 2018-03-20
tests: convert neutron to use gremlin builder

hunchback:
verify that neutron tests still pass


hunchback:
moved to tests-all-gremlin PR

PR #878, opened 2018-03-20
tests: converted packet injector to use gremlin builder

hunchback:
verify that packet injector tests still pass


lebauce:
I think a single PR for converting the remaining tests to the Gremlin builder is enough

PR #880, opened 2018-03-20
tests: converted scale tests to gremlin builder

hunchback:
verify scale tests still pass


safchain:
Can you push an only one PR for these changes as it address the same thing. It will reduce the CI queue.

hunchback:
moved to tests-all-gremlin PR

PR #899, opened 2018-03-23
client : don't authenticate if username is not set

nplanel:


nplanel:
to avoid wrong client error message, like
2018-03-23T20:47:50.471Z	ERROR	client/query.go:58 glob..func24	fedora27: **Authentication failed**: dial tcp [::1]:8082: getsockopt: connection refused

PR #911, opened 2018-03-28
sockets: fix states starting at 1 not 0

safchain:


nplanel:
from kernel header [include/net/tcp_states.h](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/net/tcp_states.h?id=HEAD)

PR #914, opened 2018-03-29
ui: fix NaN error for node bandwidth

masco-enovance:
if the time is converted to string don't
calculate bandwidth to avoid NaN error

### DIFF
--- a/statics/js/components/topology.js
+++ b/statics/js/components/topology.js
@@ -692,7 +692,7 @@ var TopologyComponent = {
     },
 
     normalizeMetric: function(metric) {
-      if (metric.Start && metric.Last) {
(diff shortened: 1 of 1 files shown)


statics/js/components/topology.js
```diff
@@ -692,7 +692,7 @@ var TopologyComponent = {
     },
 
     normalizeMetric: function(metric) {
-      if (metric.Start && metric.Last) {
+      if (metric.Start && metric.Last && !metric.Bandwidth) {
```
nplanel: the test should be (metric.Last - metric.Start) > 0

masco-enovance: Done.

PR #921, opened 2018-04-03
ovsmirror: fix panic if unregister while subprobe not yet set

safchain:


lebauce:
nit: s/ovsmmirror/ovsmirror

safchain:
rerun tests

PR #923, opened 2018-04-03
flow: Added config vars to enable/disable capturing/injection

vperelmax:
added config variables to enable/disable
traffic capturing and packet injection.
With this configurable limitations Skydive can
be used in two modes: as a sniffering and as
a monitoring tool.

co-author: Sylvain Afchain <safchain@gmail.com>

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -64,6 +64,13 @@ func writeError(w http.ResponseWriter, status int, err error) {
 	w.Write([]byte(err.Error()))
 }
 
+// RegisterNotAllowedAPIHandler registers a new handler for unpermitted use of API
+func (a *Server) RegisterNotAllowedAPIHandler(handler Handler) {
+	a.HTTPServer.Router.PathPrefix("/api/" + handler.Name()).HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+	})
+}
+
 // RegisterAPIHandler registers a new handler for an API
 func (a *Server) RegisterAPIHandler(handler Handler) error {
 	name := handler.Name()
--- a/config/config.go
+++ b/config/config.go
@@ -77,7 +77,10 @@ func init() {
 	cfg.SetDefault("agent.topology.socketinfo.host_update", 10)
 	cfg.SetDefault("agent.X509_servername", "")
 
+	cfg.SetDefault("analyzer.capture_enabled", true)
 	cfg.SetDefault("analyzer.listen", "127.0.0.1:8082")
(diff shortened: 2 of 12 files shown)


safchain:
ok to test

safchain:
add to whitelist

config/config.go
```diff
@@ -84,6 +84,9 @@ func init() {
 	cfg.SetDefault("analyzer.storage.bulk_insert_deadline", 5)
 	cfg.SetDefault("analyzer.storage.max_flow_buffer_size", 100000)
 	cfg.SetDefault("analyzer.topology.probes", []string{})
+	cfg.SetDefault("analyzer.capture_enabled", true)
```
safchain: can you keep the alphabetical order please

etc/skydive.yml.default
```diff
@@ -52,6 +52,12 @@ analyzer:
   # Enable/disable ssh to hosts
   # ssh_enabled: false
 
+  # Enable/disable capturing (enabled by default)
+  # capture_enabled: true
+  # Enable/disable packet injection (enabled by default)
+  # inject_enabled: true
```
safchain: packet_injection_enabled

etc/skydive.yml.default
```diff
@@ -52,6 +52,12 @@ analyzer:
...
+  # Enable packet capturing (enabled by default)
```
safchain: maybe better to move this just after the capture_enabled

vperelmax:
I have fixed according with the comments.
Thank you for the fast review!

safchain:
go-fmt job reports error, please check

api/server/server.go
```diff
@@ -64,6 +64,49 @@ func writeError(w http.ResponseWriter, status int, err error) {
 	w.Write([]byte(err.Error()))
 }
 
+// RegisterNotAllowedAPIHandler registers a new handler for unpermitted use of API
+func (a *Server) RegisterNotAllowedAPIHandler(handler Handler) {
+	name := handler.Name()
```
lebauce:
```
	a.HTTPServer.Router.PathPrefix("/api/" + handler.Name()).HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusMethodNotAllowed)
	})
```

vperelmax: Could you please explain the comment? I am sorry, I don't understand :(
Should it be instead of all the described cases?

lebauce: Sorry, my comment was a bit raw :-) Yes, you can use it instead of all the described cases. We simply set a handler for all methods and the URLs with the given prefix

vperelmax: OK, will do it. 
Thanks!

PR #928, opened 2018-04-04
python: use login endpoint for authentication

safchain:
Always use /login endpoint in order to use
a token instead of basic auth.

### DIFF
--- a/contrib/python/api/skydive/websocket/client.py
+++ b/contrib/python/api/skydive/websocket/client.py
@@ -23,7 +23,6 @@ try:
     import asyncio
 except ImportError:
     import trollius as asyncio
-import base64
 import functools
 import json
 try:
@@ -31,17 +30,18 @@ try:
 except ImportError:
     import httplib
 import logging
-import requests
 import ssl
 import uuid
 try:
     from urllib.parse import urlparse
 except ImportError:
     from urlparse import urlparse
+import warnings
 
 from autobahn.asyncio.websocket import WebSocketClientProtocol
 from autobahn.asyncio.websocket import WebSocketClientFactory
 
+from skydive.auth import Authenticate
 from skydive.encoder import JSONEncoder
 
 
@@ -168,6 +168,17 @@ class WSClient(WebSocketClientProtocol):
         self.insecure = insecure
         self.kwargs = kwargs
 
+        self.url = urlparse(self.endpoint)
+
+        scheme = "http"
+        if self.url.scheme == "wss":
+            scheme = "http"
+
+        self.auth = Authenticate(
+            "%s:%s" % (self.url.hostname, self.url.port),
+            scheme=scheme, username=username, password=password,
+            insecure=insecure)
+
     def connect(self):
         factory = WebSocketClientFactory(self.endpoint)
         factory.protocol = self.protocol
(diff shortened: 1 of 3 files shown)


safchain:
run python-tests

contrib/python/api/skydive/rest/client.py
```diff
@@ -65,11 +71,55 @@ def lookup(self, gremlin, klass=None):
                                                  context=context))
 
         opener = request.build_opener(*handlers)
-        req = request.Request(url,
-                              data.encode(),
-                              {'Content-Type': 'application/json'})
 
+        req = request.Request(url, data=urlencoder.urlencode(data).encode())
         resp = opener.open(req)
+        if resp.getcode() == 200:
```
snapiri: Why not handle set the authenticate to True here? Why leave it up to the code using this code?

contrib/python/api/skydive/rest/client.py
```diff
@@ -65,11 +71,55 @@ def lookup(self, gremlin, klass=None):
...
+                if cookie.name == "authtok":
+                    self.authtok = cookie.value
+
+        return resp
+
+    def request(self, path, data=None):
+        if self.username and not self.authenticated:
+            resp = self.authenticate()
+            if resp.getcode() != 200:
+                self.authenticated = False
```
snapiri: IMHO this should be set by Authenticate

contrib/python/api/skydive/rest/client.py
```diff
@@ -65,11 +71,55 @@ def lookup(self, gremlin, klass=None):
...
+            handlers.append(request.HTTPSHandler(debuglevel=self.debug,
+                                                 context=context))
+
+        opener = request.build_opener(*handlers)
+
+        headers = {'Content-Type': 'application/json'}
+        req = request.Request(url, data=data, headers=headers)
+
+        resp = opener.open(req)
+        if resp.getcode() == 401:
```
snapiri: In case of 401, we do not get a response code, but rather RuntimeException with 401.
Better to wrap with try..except

contrib/python/api/skydive/websocket/client.py
```diff
@@ -31,18 +31,18 @@
...
     from urllib.parse import urlparse
 except ImportError:
     from urlparse import urlparse
+import warnings
 
 from autobahn.asyncio.websocket import WebSocketClientProtocol
 from autobahn.asyncio.websocket import WebSocketClientFactory
 
 from skydive.encoder import JSONEncoder
+from skydive.rest.client import RESTClient
```
snapiri: Though it should work, do you think it is good thing to use the RESTClient here?
Maybe have a new file that both will use?
(though this is just a nit)

safchain: yes going to move this outside in another dedicated file

contrib/python/api/skydive/websocket/client.py
```diff
@@ -181,9 +190,13 @@ def connect(self):
             factory.headers["X-Persistence-Policy"] = "DeleteOnDisconnect"
 
         if self.username:
-            up = "%s:%s" % (self.username, self.password)
-            authorization = base64.b64encode(up.encode()).decode("ascii")
-            factory.headers["Authorization"] = 'Basic %s' % authorization
+            resp = self.restclient.authenticate()
+            if resp.getcode() == 200:
```
snapiri: What if the authentication failed?

contrib/python/api/skydive/websocket/client.py
```diff
@@ -231,17 +250,12 @@ def login(self, host_spec="", username="", password=""):
...
-            },
-            verify=(not self.insecure)
-        )
+        rest = RESTClient(host_spec, scheme=scheme,
+                          username=username, password=password)
+        resp = rest.authenticate()
 
-        if res.status_code == 200:
-            cookie = 'authtok={}'.format(res.cookies['authtok'])
+        if resp.getcode() == 200:
```
snapiri: Same here...

contrib/python/api/skydive/rest/client.py
safchain: I'll wrap in order to reset the authentication statut but will let the exception so that the user can handle it

safchain:
run python-tests

safchain:
run python-tests

safchain:
run python-tests

safchain:
run python-tests

safchain:
run python-tests

safchain:
rerun tests

safchain:
rerun tests

safchain:
rerun python-tests

safchain:

PR #930, opened 2018-04-05
When client is running in a thread, an exception is thrown

### LINKED ISSUE #929: when running the python client in a thread, an exception is thrown
When running the websocket client in a new thread, an exception is thrown that the thread does not have an event loop.
When running under the main thread the asyncio creates the loop automatically.
See: https://docs.python.org/3/library/asyncio-eventloops.html (Under: Event loop policy interface)

In client.py, the following code:
        self.loop = asyncio.get_event_loop()

Should be changed to something like:
        try:
            self.loop = asyncio.get_event_loop()
        except RuntimeError:
            self.loop = asyncio.new_event_loop()
            asyncio.set_event_loop(self.loop)

snapiri:
This is because there is no event_loop in threads by default, and in the
main thread it is created automatically.
This code resolves #929

### DIFF
--- a/contrib/python/api/skydive/websocket/client.py
+++ b/contrib/python/api/skydive/websocket/client.py
@@ -191,7 +191,12 @@ class WSClient(WebSocketClientProtocol):
         if self.cookies:
             factory.headers['Cookie'] = ';'.join(self.cookies)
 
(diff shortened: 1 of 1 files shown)


safchain:
run python-tests

PR #952, opened 2018-04-12
change helm chart default filter to support PR #937

eranra:


contrib/charts/templates/deployment-analyzer.yaml
```diff
@@ -56,7 +56,7 @@ spec:
         - name: SKYDIVE_UI_TOPOLOGY_FAVORITES
           value: '{"Infrastructure":"G.V().Has(\"Manager\", NE(\"k8s\"))","Kubernetes":"G.V().Has(\"Manager\", \"k8s\")"}'
         - name: SKYDIVE_UI_TOPOLOGY_DEFAULT_FILTER
-          value: 'G.V().Has("Manager", NE("k8s"))'
+          value: 'Infrastructure'
```
safchain: Key has to be lower case, viper constraint

eranra: Fixed and upload new commit move to lower case:

BTW: I checked and it worked with Uppercase ... so not sure about Viper 👍

PR #962, opened 2018-04-17
es: support different elasticsearch connections per index

hunchback:
the following change adds the infra changes needed for pointing topology and flow each to a different ElasticSearch database (the extra change need to support this is extending the `skydive.yml` `analyzer.storage` to support flow connection and point to this path during `NewConnConfig` call (for flow). 

This would enable extra isolation between the two in productive environments.

For now default behavior (in which both flow and topology take connection configuration from `storage.elasticsearch` remains).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -161,6 +161,7 @@ func init() {
 	cfg.SetDefault("sflow.port_min", 6345)
 	cfg.SetDefault("sflow.port_max", 6355)
 
+	cfg.SetDefault("storage.elasticsearch.driver", "elasticsearch")
 	cfg.SetDefault("storage.elasticsearch.host", "127.0.0.1:9200")
 	cfg.SetDefault("storage.elasticsearch.maxconns", 10)
 	cfg.SetDefault("storage.elasticsearch.retry", 60)
@@ -169,10 +170,12 @@ func init() {
 	cfg.SetDefault("storage.elasticsearch.index_age_limit", 0)
 	cfg.SetDefault("storage.elasticsearch.index_entries_limit", 0)
 	cfg.SetDefault("storage.elasticsearch.indices_to_keep", 0)
+	cfg.SetDefault("storage.orientdb.driver", "orientdb")
 	cfg.SetDefault("storage.orientdb.addr", "http://localhost:2480")
 	cfg.SetDefault("storage.orientdb.database", "Skydive")
 	cfg.SetDefault("storage.orientdb.username", "root")
 	cfg.SetDefault("storage.orientdb.password", "root")
+	cfg.SetDefault("storage.memory.driver", "memory")
 
 	cfg.SetDefault("ui.theme", "dark")
 	cfg.SetDefault("ui.bandwidth_absolute_active", 1)
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -63,7 +63,7 @@ analyzer:
 
   # Flow storage engine
   flow:
-    # Available: elasticsearch, orientdb
+    # Available: elasticsearch, orientdb, sample
     # backend: elasticsearch
 
     # maximum number of flows aggregated between two data store inserts
@@ -87,7 +87,7 @@ analyzer:
     # indices_to_keep: 0
 
   topology:
-    # graph backend : memory, elasticsearch, orientdb
+    # graph backend : memory, elasticsearch, orientdb, sample
     # backend: memory
 
     # Define static interfaces and links updating Skydive topology
@@ -244,8 +244,14 @@ opencontrail:
   # mpls_udp_port: 51234
 
 storage:
-  # Elasticsearch connection information.
+  # an example for user defined backend 'sample'.
+  # sample:
+  #   driver: elasticsearch, orientdb, memory
+  #   here add driver type specific fields.
+
+  # Elasticsearch backend information.
   elasticsearch:
+    # driver: elasticsearch
     # host: 127.0.0.1:9200
     # maxconns: 10
     # retry: 60
@@ -262,13 +268,18 @@ storage:
     # The number of indices to keep before deleting.
     # A value of 0 specifies no limit (i.e. indices will never be deleted)
     # indices_to_keep: 0
-
-  # OrientDB connection information
+    
+  # OrientDB backend information.
   orientdb:
+    # driver: orientdb
     # addr: http://127.0.0.1:2480
     # database: Skydive
     # username: root
     # password: hello
+
+  # Memory backend information.
+  memory:
+    # driver: memory
 
 logging:
   # level: INFO
--- a/storage/elasticsearch/client.go
+++ b/storage/elasticsearch/client.go
@@ -47,24 +47,36 @@ const indexVersion = 11
 const indexPrefix = "skydive"
 const indexAllAlias = "all"
 
-// IndexConfig describes index limits driving roll policy
-type IndexConfig struct {
+// Config describes configuration for elasticsearch
+type Config struct {
+	ElasticHost  string
+	MaxConns     int
+	RetrySeconds int
+	BulkMaxDocs  int
+	BulkMaxDelay int
 	EntriesLimit int
 	AgeLimit     int
 	IndicesLimit int
 }
 
-// NewIndexConfigFromConfig create new limits from configuration
-func NewIndexConfig(path string) IndexConfig {
-	indexCfg := IndexConfig{}
-	indexCfg.EntriesLimit = config.GetInt(path + ".index_entries_limit")
-	indexCfg.AgeLimit = 0
+func NewConfig(path string) Config {
+	cfg := Config{}
+
+	cfg.ElasticHost = config.GetString(path + ".host")
+	cfg.MaxConns = config.GetInt(path + ".maxconns")
+	cfg.RetrySeconds = config.GetInt(path + ".retry")
+	cfg.BulkMaxDocs = config.GetInt(path + ".bulk_maxdocs")
+	cfg.BulkMaxDelay = config.GetInt(path + ".bulk_maxdelay")
+
+	cfg.EntriesLimit = config.GetInt(path + ".index_entries_limit")
+	cfg.AgeLimit = 0
 	// TODO: read the AgeLimit from the configuration. At this stage we are setting statically to zero since
 	// TODO: the feature of reading the index creation date is not supported.
 	// TODO: the code that need to happen when we will
-	// TODO: be ready:: indexCfg.AgeLimit = config.GetInt(path + ".index_age_limit")
-	indexCfg.IndicesLimit = config.GetInt(path + ".indices_to_keep")
-	return indexCfg
+	// TODO: be ready:: cfg.AgeLimit = config.GetInt(path + ".index_age_limit")
+	cfg.IndicesLimit = config.GetInt(path + ".indices_to_keep")
+
+	return cfg
 }
 
 // ElasticSearchClientInterface describes the mechanism API of ElasticSearch database client
@@ -99,15 +111,6 @@ type ElasticIndex struct {
 	timeCreated    time.Time
 }
 
-// ConnConfig describes the ElasticSearch configuration section
-type ConnConfig struct {
-	ElasticHost  string
-	MaxConns     int
-	RetrySeconds int
-	BulkMaxDocs  int
-	BulkMaxDelay int
-}
-
 // ElasticSearchClient describes a ElasticSearch client connection
 type ElasticSearchClient struct {
 	connection *elastigo.Conn
@@ -117,8 +120,7 @@ type ElasticSearchClient struct {
 	wg         sync.WaitGroup
 	name       string
 	mappings   Mappings
-	indexCfg   IndexConfig
-	connCfg    ConnConfig
+	cfg        Config
 	index      *ElasticIndex
 }
 
@@ -410,18 +412,18 @@ func (c *ElasticSearchClient) FormatFilter(filter *filters.Filter, mapKey string
 }
 
 func (c *ElasticSearchClient) shouldRollIndexByCount() bool {
-	if c.indexCfg.EntriesLimit == 0 {
+	if c.cfg.EntriesLimit == 0 {
 		return false
 	}
 	logging.GetLogger().Debugf("%s entries counter is %d", c.name, c.index.entriesCounter)
-	if c.index.entriesCounter < c.indexCfg.EntriesLimit {
+	if c.index.entriesCounter < c.cfg.EntriesLimit {
 		return false
 	}
 	c.indexer.Flush()
 	time.Sleep(1 * time.Second)
 
 	c.index.entriesCounter = c.countEntries()
-	if c.index.entriesCounter < c.indexCfg.EntriesLimit {
+	if c.index.entriesCounter < c.cfg.EntriesLimit {
 		return false
 	}
 	logging.GetLogger().Debugf("%s enough entries to roll", c.name)
@@ -429,12 +431,12 @@ func (c *ElasticSearchClient) shouldRollIndexByCount() bool {
 }
 
 func (c *ElasticSearchClient) shouldRollIndexByAge() bool {
-	if c.indexCfg.AgeLimit == 0 {
+	if c.cfg.AgeLimit == 0 {
 		return false
 	}
 	age := int(time.Now().Sub(c.index.timeCreated).Seconds())
 	logging.GetLogger().Debugf("%s age is %d", c.name, age)
-	if age < c.indexCfg.AgeLimit {
+	if age < c.cfg.AgeLimit {
 		return false
 	}
 	logging.GetLogger().Debugf("%s old enough to roll", c.name)
@@ -446,7 +448,7 @@ func (c *ElasticSearchClient) shouldRollIndex() bool {
 }
 
 func (c *ElasticSearchClient) delIndices() {
-	if c.indexCfg.IndicesLimit == 0 {
+	if c.cfg.IndicesLimit == 0 {
 		return
 	}
 
@@ -455,7 +457,7 @@ func (c *ElasticSearchClient) delIndices() {
 		return indices[i].Name < indices[j].Name
 	})
 
-	numToDel := len(indices) - c.indexCfg.IndicesLimit
+	numToDel := len(indices) - c.cfg.IndicesLimit
 	if numToDel <= 0 {
 		return
 	}
@@ -475,7 +477,7 @@ func (c *ElasticSearchClient) rollIndex() error {
 
 	c.indexer.Stop()
 	c.stopErrorReader()
-	c.indexer = newBulkIndexer(c.connection, c.connCfg)
+	c.indexer = newBulkIndexer(c.connection, c.cfg)
 	c.indexer.Start()
 	c.startErrorReader()
 
@@ -643,23 +645,23 @@ func (c *ElasticSearchClient) Started() bool {
 	return c.started.Load() == true
 }
 
-func newBulkIndexer(c *elastigo.Conn, connCfg ConnConfig) *elastigo.BulkIndexer {
-	indexer := c.NewBulkIndexer(connCfg.MaxConns)
-	indexer.RetryForSeconds = connCfg.RetrySeconds
-	if connCfg.BulkMaxDocs > 0 {
-		indexer.BulkMaxDocs = connCfg.BulkMaxDocs
+func newBulkIndexer(c *elastigo.Conn, cfg Config) *elastigo.BulkIndexer {
+	indexer := c.NewBulkIndexer(cfg.MaxConns)
+	indexer.RetryForSeconds = cfg.RetrySeconds
+	if cfg.BulkMaxDocs > 0 {
+		indexer.BulkMaxDocs = cfg.BulkMaxDocs
 
 		// set chan to 80% of max doc
-		if connCfg.BulkMaxDocs > 100 {
-			indexer.BulkChannel = make(chan []byte, int(float64(connCfg.BulkMaxDocs)*0.8))
+		if cfg.BulkMaxDocs > 100 {
+			indexer.BulkChannel = make(chan []byte, int(float64(cfg.BulkMaxDocs)*0.8))
 		}
 	}
 
 	// override the default error chan size
 	indexer.ErrorChannel = make(chan *elastigo.ErrorBuffer, 100)
 
-	if connCfg.BulkMaxDelay > 0 {
-		indexer.BufferDelayMax = time.Duration(connCfg.BulkMaxDelay) * time.Second
+	if cfg.BulkMaxDelay > 0 {
+		indexer.BufferDelayMax = time.Duration(cfg.BulkMaxDelay) * time.Second
 	}
 
 	return indexer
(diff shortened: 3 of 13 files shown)


safchain:
ok to test

safchain:
rerun unit-tests

storage/elasticsearch/client.go
```diff
@@ -83,34 +83,43 @@ type ElasticSearchClientInterface interface {
...
-	bulkMaxDocs  int
-	bulkMaxDelay int
+	connection *elastigo.Conn
+	indexer    *elastigo.BulkIndexer
+	started    atomic.Value
+	quit       chan bool
+	wg         sync.WaitGroup
+	name       string
+	mappings   Mappings
+	idxCfg     IndexConfig
```
eranra: Please do not use short names, change idx to index, it is confusing. so for this variable use indexConfig

storage/elasticsearch/client.go
```diff
@@ -83,34 +83,43 @@ type ElasticSearchClientInterface interface {
...
 	entriesCounter int
-	mappings       []map[string][]byte
-	name           string
 	path           string
 	timeCreated    time.Time
-	limits         ElasticLimits
+}
+
+// ConnConfig describes the ElasticSearch configuration section
+type ConnConfig struct {
```
eranra: Change to ConnectionConfig

storage/elasticsearch/client.go
```diff
@@ -636,72 +642,86 @@ func (c *ElasticSearchClient) Stop() {
...
 	}
 
 	return indexer
 }
 
-// NewElasticSearchClient creates a new ElasticSearch client
-func NewElasticSearchClient(url *url.URL, maxConns int, retrySeconds int, bulkMaxDocs int, bulkMaxDelay int) (*ElasticSearchClient, error) {
+func urlFromHost(host string) (*url.URL, error) {
+	urlStr := host
+	if !strings.HasPrefix(urlStr, "http://") && !strings.HasPrefix(urlStr, "https://") {
```
eranra: So host can be either hostname without http/s or full url ...  can you add explanation in default .yml configuration file

storage/elasticsearch/client.go
hunchback: Using shot names is widly accepted in golang (as opposed for example to Java) srarting from reserved words eg func, chan, int; on to module names eg fmt; and variable names. 

Specifically conn as used widely in idiomatic golang code

storage/elasticsearch/client.go
hunchback: And variables are typically shorter than structs, as a variable lives in a block or struct (constrained scope) while a struct lives within the module (much larger scope) ... thus in idiomatic golang (and other languages) variables tend to be shorter

storage/elasticsearch/client.go
hunchback: This issue is outside the scope of this PR as i did not change the format of the skydive.yml file.

Additionally omitting the http prefix is used elsewhere (although not consistently) in other places such as analyzers.

For now i rather this not be addressed

eranra: K, you are correct that this needs attention also in other places, I submitting a new issue to track https://github.com/skydive-project/skydive/issues/964

storage/elasticsearch/client.go
eranra: Yes, I see that in the code in many places (including the lib); so I'm ok with that change ... although very confusing to see conn and think what it stands for, connection is much more understandable.

storage/elasticsearch/client.go
hunchback: ok

storage/elasticsearch/client.go
hunchback: ok will change idxCfg --> indexCfg

hunchback:
I think i would take #968 furthur, such that i can refer to memory, orientdb, elasticsearch (default entries defined under storage), but can also add more options such as elasticsearch2 (an alternative ES instace) and then i could move limits from analyzer.flow (or analyzer.storage) to storage.elasticsearch, storage.elasticsearch2, etc.

hunchback:
@nplanel / @safchain 

The proposed fix has some general code improvements:
 - moving all initialization to the c'tor (and out of `Start()` which is an interface method). 
 - adding of helper methods for reading from configuration (and eliminating duplicated reading of configuration parameters)
I could be merged regardless of the separate discussion on changes on skydive.yml storage configuration changes.

If nonetheless you prefer to also address changes in the configuration in the same go - then I would suggest the following:

```
analyzer:
  storage:
    flow:
      backend: elasticsearch.flow
      max_chan_size: 100000                        # previous max_flow_buffer_size
    topology:
      backend: elasticsearch.default

storage:
  elasticsearch:
    default:
      host: 127.0.0.1:9200
      bulk_max_conns: 10                                 # previously maxconns        
      bulk_retry_seconds: 60                             # previously retry
      bulk_insert_size: 100                                 # previously bulk_maxdocs, bulk_insert
      bulk_insert_seconds: 5                              # previously bulk_maxdelay, bulk_insert_deadline
      index_entries_limit: 0
      index_age_limit: 0
      indices_to_keep: 0
    flow:
      ...
  orientdb:
    ...
```

nplanel:
@eranra @hunchback @safchain 

1/ Would be better to not mix backend and indexes, there are two different things
2/ Let's keep the configuration simple and backward compatible
  - storage.elasticsearch define the default configuration (as today)
  - storage.elasticsearch.{index} define the {index} configuration (if exist, else fallback to default)
3/ analyzer flow_server bulk_insert and bulk_insert_deadline, use the configuration values according to the selected flow backend

```
analyzer:
  storage:
    flow:
      backend: elasticsearch
      max_flow_buffer_size: 100000
    topology:
      backend: elasticsearch

storage:
  elasticsearch:
    host: 127.0.0.1:9200
    maxconns: 10
    retry: 60
    bulk_maxdocs: 100
    bulk_maxdelay: 5
    index_entries_limit: 0
    index_age_limit: 0
    indices_to_keep: 0
    flow:                                                           # optional
      host: 127.0.1.1:9200
      ...
    topology:                                                    # optional
      host: 127.0.2.1:9200
      ...
    snort:                                                          # optional
      host: 127.0.3.1:9200
      ...
  orientdb:
    ...

```

hunchback:
@nplanel  (@safchain, @eranra) - proposed config seem fine, can this be handled in a separate PR?

nplanel:
I think the config need to be handled in the this PR as it's related to ES and indexes.

storage/elasticsearch/elasticsearch_test.go
```diff
@@ -27,121 +27,122 @@ const testMapping = `
...
+	client.Start()
 
 	return client, nil
 }
 
 // test rolling elasticsearch indices based on count limit
 func TestElasticsearchShouldRollByCount(t *testing.T) {
-	limits := ElasticLimits{}
-	limits.EntriesLimit = 5
+	indexCfg := IndexConfig{}
```
nplanel: please don't rename if it's not improving IndexConfig is exactly the same thing as ElasticLimits
let's focus on the patch behaviour, the review will be more readable

hunchback:
Ok

hunchback:
@nplanel - I have just uploaded an updated proposal - to handle default values I and added a notation with '?' such as `storage.elasticsearch?flow.host` which means check if `storage.elasticsearch.flow.,host` is set if not then fallback to `storage.elasticsearch.host`

nplanel:
@hunchback 
why adding "?" ?
why not try to resolve the key like

```
host := ""
if host = cfg.GetString("storage.elasticsearch.flow.host"); host == "" {
    host = cfg.GetString("storage.elasticsearch.host")
}
```

config/config.go
```diff
@@ -89,12 +104,10 @@ func init() {
 	cfg.SetDefault("analyzer.packet_injection_enabled", true)
 	cfg.SetDefault("analyzer.replication.debug", false)
 	cfg.SetDefault("analyzer.ssh_enabled", false)
-	cfg.SetDefault("analyzer.storage.bulk_insert", 100)
-	cfg.SetDefault("analyzer.storage.bulk_insert_deadline", 5)
-	cfg.SetDefault("analyzer.storage.index_age_limit", 0)
-	cfg.SetDefault("analyzer.storage.index_entries_limit", 0)
-	cfg.SetDefault("analyzer.storage.indices_to_keep", 0)
-	cfg.SetDefault("analyzer.storage.max_flow_buffer_size", 100000)
+	cfg.SetDefault("analyzer.storage.flow.bulk_insert", 100)
```
nplanel: please take a look at #968 
graph.backend => analyzer.topology.backend
analyzer.storage => analyzer.flow

analyzer.flow.{bulk_insert,bulk_insert_deadline,max_flow_buffer_size} must be removed from the configuration and deduced/resolved from the used backend (analyzer.flow.backend)

hunchback:
I am using the '?' notation so as to keep the calling code simple.

Additionaly, by adding the '?' notation i was able to extend all existing
config.GetXXX(key string) functions with only minimal code changes (and
minimal added complexity)

hunchback:
I will fix the handling of deprecated keys according to example #968

hunchback:
I am using the '?' notation so as to keep the calling code simple.

Additionally, by adding the '?' notation i was able to extend all existing
config.GetXXX(key string) functions with only minimal code changes (and
minimal added complexity).

Compare: 

```
retry := 0
if val := cfg.Get("storage.elasticsearch.flow.retry"); val != nil {
    retry = val.(int)
} else {
    retry = cfg.GetInt("storage.elasticsearch.retry")
}
```
 To

```
retry := cfg.GetInt("storage.elasticsearch?flow.retry")
```

In our case things are ever more acute as we have a helper function handling the calling of configuration which is also simplified to:

In your case it would be:

```
cfg := es.NewConfig("storage.elasticsearch.flow", "storage.elasticsearch")
```

Simplified to:

```
cfg := es.NewConfig("storage.elasticsearch?flow")
```

nplanel:
@hunchback the only call you need to do is 

```
limits := elasticsearch.NewElasticLimitsFromConfig("storage.elasticsearch", "flow")

or even you can simplify

limits := elasticsearch.NewElasticLimitsFromConfig("flow")
limits := elasticsearch.NewElasticLimitsFromConfig("topology")
...
limits := elasticsearch.NewElasticLimitsFromConfig()
```

hunchback:
@nplanel - I agree there is little difference between:

```
limits := elasticsearch.NewElasticLimitsFromConfig("storage.elasticsearch", "flow")
```

And:

```
limits := elasticsearch.NewElasticLimitsFromConfig("storage.elasticsearch?flow")
```

But this is a negligible aspect - the main issue is `config.GetInt("storage.elasticsearch?flow.retry")` and friends if not handled seamlessly by `config` then we have a 7-8 way code duplication, and this will only get worse we we use the default/fallback trick elsewhere in the code (in storage.orientdb for example).

hunchback:
Tests failed due to:

```
2018-04-24T07:59:03.582Z	DEBUG	helper/helper.go:135 ExecCmds	skydive-slave.rdocloud: Executing command [lxc launch images:fedora/27 test-skydive-lxd-simple]
2018-04-24T07:59:03.628Z	DEBUG	helper/helper.go:138 ExecCmds	skydive-slave.rdocloud: Command returned Connection refused; is LXD running?

```

@nplanel  - can you check was is the matter?

nplanel:

lebauce:
I'm not a fan of the "?" notation and also prefer ```
cfg := es.NewConfig("storage.elasticsearch.flow", "storage.elasticsearch")
```

One way would be to make the GetString, GetInt take a ```...string``` argument and returns the first that is set, but I'm not sure it's worth it.

There are changes in your PR related to the proposal, but it seems the PR is not rebased of it, that makes it a bit hard to understand.

**Comment by hunchback** [comment]
I thought about extending config.GetInt() and friends ... but this has two
shortcomings:

1.  Usage is more comborsum, eg: GetInt("storage.elasticsearch.flow.retry",
"storage.elasticsearch.retry") compared to
GetInt("storage.elasticsearch?flow.retry")

2. Extending interfaces will cascade down to extending each functions'
implementation ...

So i don't see any advantage of this over the '?' notation .. and would
much prefer we stay with my current implementation. Also note this notation
is not exposed to users - only used from within the code.

As to your comment on rebase - please elaborate

**Comment by lebauce** [comment]
It's not clear what `GetInt("storage.elasticsearch?flow.retry")` (at least for me). And it only allows simple use cases such as "look inside the child then in the parent if you didn't find it".

GetInt("storage.elasticsearch.flow.retry", "storage.elasticsearch.retry") seems a bit clearer on what it does and allows the keys to be completely different.
That being said, maybe we should the KISS approach and go with:

```
config := cfg.Get("storage.elasticsearch.flow")
if config != nil {
    ...
} else {
    config = cfg.Get("storage.elasticsearch")
}
```
as it's only used once for now. We'll still be able to generalize that if we use it for something else. What do you think ?

**Comment by hunchback** [comment]
@lebauce - the snipped you provided will fail generally, as we want the ability per each parameter to get the value first from "storage.elasticseach.flow" and fallback to "storage.elasticsearch". For example:

```
storage:
  elasticsearch:
    retry: 10
    flow:
    # no retry value
```
In your snippet you will incorrectly return `0` rather than `10`

In effect extending the key syntax provides a means of extending capabilities while only having to "fix" one function - `realKey()` - so this is the generic way forward.

**Comment by hunchback** [comment]
@lebauce - I am not very happy with the last commit (below), I would much rather drop it ...

https://github.com/skydive-project/skydive/pull/962/commits/dc3f024832f2796a8651daf0f72e01df464c8215

File: config/config_test.go [hunk 6]
```diff
@@ -76,3 +77,60 @@ openstack:
...
+`)
+
+	cfg.ReadConfig(bytes.NewBuffer(yamlv1))
+
+	expected := 1
+	actual := GetInt("parent.child.read_from_parent", "parent.read_from_parent")
+	if expected != actual {
+		t.Fatal(fmt.Sprintf("Config 'read_from_parent': expected %d; actual %d", expected, actual))
+	}
+}
```
nplanel: please test GetInt/GetString("a", "b", "c") should print an error message (see capturer TestRelocation in config/config_test.go

storage/elasticsearch/client.go
```diff
@@ -47,24 +47,42 @@ const indexVersion = 11
...
 	AgeLimit     int
 	IndicesLimit int
 }
 
-// NewElasticLimitsFromConfig create new limits from configuration
-func NewElasticLimitsFromConfig(path string) ElasticLimits {
-	limits := ElasticLimits{}
-	limits.EntriesLimit = config.GetInt(path + ".index_entries_limit")
-	limits.AgeLimit = 0
+func NewConfig(name string) Config {
```
nplanel: NewConfig(name ...string)
if len(name) > 0 ...

to avoid call like NewConfig("")

storage/elasticsearch/client.go
```diff
@@ -47,24 +47,42 @@ const indexVersion = 11
...
+	cfg := Config{}
+
+	second := "storage.elasticsearch"
+	first := second
+	if name != "" {
+		first += "." + name
+	}
+
+	cfg.ElasticHost = config.GetString(first+".host", second+".host")
```
nplanel: as everything is almost constant and static why not writing

se := "storage.elasticsearch"
cfg.ElasticHost = config.GetString(se + ".host", se+"."+name+".host" )

config/config.go
```diff
@@ -350,7 +362,19 @@ func IsTLSenabled() bool {
 	return false
 }
 
-func realKey(key string) string {
+func expandKey(keys ...string) string {
```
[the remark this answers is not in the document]
hunchback: ok

config/config.go
```diff
@@ -350,7 +362,19 @@ func IsTLSenabled() bool {
...
+	var k string
+	for _, k = range keys {
+		if cfg.IsSet(k) {
+			break
+		}
+	}
+	return k
```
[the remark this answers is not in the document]
hunchback: ok

config/config.go
```diff
@@ -350,7 +362,19 @@ func IsTLSenabled() bool {
 	return false
 }
 
-func realKey(key string) string {
+func expandKey(keys ...string) string {
+	var k string
+	for _, k = range keys {
+		if cfg.IsSet(k) {
+			break
```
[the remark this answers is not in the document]
hunchback: ok

storage/elasticsearch/client.go
hunchback: ok

config/config_test.go
hunchback: I just test the case of no keys - only testing GetInt (as GetString, GetBool, etc, etc) are all using the same underlying `readKey()` func.

config/config.go
```diff
@@ -44,6 +44,21 @@ var ErrNoAnalyzerSpecified = errors.New("No analyzer specified in the configurat
 var (
 	cfg           *viper.Viper
 	relocationMap = map[string][]string{
+		"analyzer.storage.flow.backend": {
```
nplanel: let's focus this pr only on connection per index
PR #968 will address new config namespace

config/config.go
hunchback: ok, note that now:

analyzer.flow.{bulk_insert,bulk_insert_deadline,max_flow_buffer_size} ==> analyzer.storage.flow.{bulk_insert,bulk_insert_deadline,max_flow_buffer_size}

topology/graph/elasticsearch.go
```diff
@@ -615,48 +613,25 @@ func (b *ElasticSearchBackend) IsHistorySupported() bool {
...
-	if err != nil {
-		return nil, err
-	}
-
-	return newElasticSearchBackend(client)
-}
-
 // NewElasticSearchBackendFromConfig creates a new graph backend based on configuration file parameters
 func NewElasticSearchBackendFromConfig() (*ElasticSearchBackend, error) {
-	elasticHost := config.GetString("storage.elasticsearch.host")
```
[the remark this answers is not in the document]
hunchback: This is now resolved by `elasticsearch.NewConfig("topology")` together with the rest of the sotrage parameters.

config/config_test.go
hunchback: ok

config/config_test.go
```diff
@@ -76,3 +77,82 @@ openstack:
...
+  read_from_child: 1
+  child:
+    read_from_child: 1
+`)
+
+	cfg.ReadConfig(bytes.NewBuffer(yamlv1))
+
+	stderr := capturer.CaptureStderr(func() {
+		GetInt()
+	})
```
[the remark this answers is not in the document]
hunchback: ok

config/config.go
hunchback: ok, i have take code related to relocation, as it is more complete that #968 - I have created a new PR from branch `storage-relocate-backend` - @safchain - please suggest how/who will handle.

hunchback:
Failed due to:

```
--- FAIL: TestPacketInjectionCapture (136.31s)
	selenium_test.go:177: By default 8 rows should be return, but got: 3
FAIL
```

Which does not appear related to my fix

lebauce:
run skydive-selenium-tests

hunchback:
can this be merged ?

analyzer/flow_server.go
```diff
@@ -114,7 +114,7 @@ func (c *FlowServerWebSocketConn) Serve(ch chan *flow.Flow, quit chan struct{},
 
 // NewFlowServerWebSocketConn returns a new WebSocket flow server
 func NewFlowServerWebSocketConn(server *shttp.Server) (*FlowServerWebSocketConn, error) {
-	flowsMax := config.GetConfig().GetInt("analyzer.storage.max_flow_buffer_size")
+	flowsMax := config.GetConfig().GetInt("analyzer.storage.flow.max_flow_buffer_size")
```
nplanel: this out of scope of this PR

config/config_test.go
```diff
@@ -76,3 +77,101 @@ openstack:
...
+	var yamlv1 = []byte(`
+parent:
+  read_from_child: 1
+  child:
+    read_from_child: 1
+`)
+
+	cfg.ReadConfig(bytes.NewBuffer(yamlv1))
+
+	val := GetInt("a", "b", "c")
```
nplanel: this test should report an warning

storage/elasticsearch/client.go
hunchback: this will break as `name` may not be "" and in that case what you proposed will evaluate to: 
```
GetString("storage.elasticsearch.host", "storage.elasticseach..host")
```

config/config.go
```diff
@@ -350,7 +357,23 @@ func IsTLSenabled() bool {
...
+		fmt.Fprintf(os.Stderr, "Was expecting at least one key")
+		return ""
+	}
+	var k string
+	for _, k = range keys {
+		if cfg.IsSet(k) {
+			return k
+		}
+	}
+	return k
```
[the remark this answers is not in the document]
hunchback: But this will change the current behavior which does not report attempting to read non-defined keys ... for example currently if run `GetInt("a")` no warning will be emitted. 

If you only want `GetInt("a", "b")` to emit a warming but not `GetInt("a")` then this would make  things inconsistent.

If you want both `GetInt("a", "b")` and `GetInt("a")` to emit warnings if keys not set - they this is a change that should be done outside the scope of this PR (as it will require changing at least one non-related test in config_test.go

config/config_test.go
hunchback: see above

config/config.go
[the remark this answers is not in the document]
lebauce: Ok , then let's remove the warning (the current behavior) and return ""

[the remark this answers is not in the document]
hunchback: ok, so I am returning "" (other than that leaving "as-is")

hunchback:
Selenium is failing:

```
5b2ce5b57e9b: Pull complete
Digest: sha256:df5655cafb23744708d85653156b295979da25eb11eb85b4498219ed38212a99
Status: Downloaded newer image for docker.io/elgalu/selenium:latest
...
...
2018-05-02T16:38:23.576Z	INFO	ovs/ovsdb.go:194 (*OvsMonitor).interfaceUpdated	jenkins-slave-5.rdocloud: Interface "vm2-eth0(ac79407d-c37b-4e4f-b0c6-05519a89c807)" updated
panic: test timed out after 10m0s
```

hunchback:
@nplanel - if key is not deprecated and a "" is returned by isDepKeySet()
then firstKetSet() checks if key is set - if so then *key* is returned by
firstKeySet()

lebauce:

lebauce:
run skydive-selenium-tests

etc/skydive.yml.default
```diff
@@ -262,6 +262,11 @@ storage:
     # The number of indices to keep before deleting.
     # A value of 0 specifies no limit (i.e. indices will never be deleted)
     # indices_to_keep: 0
+    
+    # NOTE: you can override with specific values any of the above values
```
safchain: I'm really not a big fan of overriding flow, topology, etc here. What about named backend/connection that you can pick to use in the backend parameter, like

```
storage:
   es_1:
      type: elasticsearch
      maxconns: 200
   es_2:
      type: elasticsearch
      maxconns: 400

analyzer:
   flow:
      backend: es_2
   topology:
      backend: es_1
```

hunchback: This raise two separate issues:

(1) should we use defaults? (such as storage.elasticsearch.flow and fallbacking to storage.elasticsearch) - as this significantly reduces the size of configuration I think this makes a lot of sense.

(2) to we use labels for backend, so as backend can also be memory or orientdb you would rather have to support values such as:

```
analyzer:
  flow:
    # backend: memory
    # backend: orentdb
    # backend: elasticsearch
    # backend: elasticsearch:es_2
```
But I am not sure this extra flexibility is required ... (as compared with defining backend:elasticsearch and having the logic of where to look for configuration coded within the code)

hunchback:
@nplanel / @safchain / @lebauce  - please review https://github.com/skydive-project/skydive/pull/962/commits/f9e3e066d1cd2d44a997f0ba35825b52d3091797

PR #966, opened 2018-04-18
auth: get ride of http basic auth

safchain:
This patch removes the support of the HTTP
part of our basic auth support as it causes
UI issue and authentification path inconsistency.

With this patch only the login endpoint using
the cookie returned as token is available.

Since this commit 81a691913310c2c019eb475bd64a760d8f6a6859
the python library relies on login and cookie.

### DIFF
--- a/http/auth.go
+++ b/http/auth.go
@@ -144,7 +144,6 @@ func NewAuthenticationClient(url *url.URL, authOptions *AuthenticationOpts) *Aut
 }
 
 type AuthenticationBackend interface {
-	AuthType() string
 	Authenticate(username string, password string) (string, error)
 	Wrap(wrapped auth.AuthenticatedHandlerFunc) http.HandlerFunc
 }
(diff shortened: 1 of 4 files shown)


safchain:
run python-tests

nplanel:
not sure that urllib will love that as it miss "WWW-Authenticate" header
urllib python lib need to do a 1st request without auth, and will authenticate via the method return by "WWW-Authenticate" header

safchain:
We are now using anymore basic auth stuff we do provide the right cookie

safchain:
previously we were using real http basic auth that is why it was needed but now just doing request on the login endpoint and then using the coookie as token. Since 81a6919

safchain:
rerun unit-tests

safchain:
rerun selenium-tests

nplanel:
@safchain yeah ok :)

safchain:

PR #967, opened 2018-04-18
ansible : fix service, reboot slaves, pause jenkins_master

nplanel:
o service enable: yes (not true)
 o reboot slaves after install aim to match running kernel with devel headers
 o pause jenkins_master only on deploy_master


nplanel:
run skydive-go-fmt

PR #968, opened 2018-04-18
config: relocate graph.backend and analyzer.storage

safchain:


eranra:
👍 for moving storage --> flow

PR #971, opened 2018-04-19
ui,oflow: group openflow rules by priority and actions

### LINKED ISSUE #970: Group Openflow rules by priority and actions
Long lists of rules can be hard to read. A first step would be to group them by priority and for a given priority by actions. Rules of a given subgroup would only be distinguishable by the filters (it should be read as a disjunction). 

Here is a table taken from an vswitch controlled by openstack/ovn with the current style: 
old_layout

Here is the same with the new slightly more structured style:
new-layout

pierrecregut:
* Make priority of the rule a separated column.
* Rules of same priority share a single priority cell spanning multiple
  rows.
* Rules with the same set of actions share a single cell spanning
  multiple rows (They have different filters).

Closes #970

### DIFF
--- a/statics/js/components/rule-detail.js
+++ b/statics/js/components/rule-detail.js
@@ -38,6 +38,7 @@ var OTHER_PORTS = -4;
 /** Port number representing the entry port - not an Openflow normalized port */
 var SAME_PORT = -5;
 var reFindInport = new RegExp('(^|.*,)in_port=([0-9]*)(,.*|$)');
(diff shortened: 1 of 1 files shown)


pierrecregut:
run tests

masco-enovance:
run tests

masco-enovance:
ok to test

nplanel:
run skydive-selenium-tests

PR #976, opened 2018-04-23
client: do not override analyzer list from config file

lebauce:


cmd/client/client.go
```diff
@@ -42,7 +42,9 @@ var ClientCmd = &cobra.Command{
 		if analyzerAddr != "" {
 			config.Set("analyzers", analyzerAddr)
 		} else {
-			config.SetDefault("analyzers", "localhost:8082")
+			if len(config.GetStringSlice("analyzers")) == 0 {
```
nplanel: why/how SetDefault override the current value of analyzers
Shall SetDefault("analyzers") rather be in config/config.go ?

lebauce: It's probably a bug in Cobra. When doing `config.SetDefault("analyzers", []string{"localhost:8082"})`, it seems to work correctly. I updated the PR

masco-enovance:
run skydive-selenium-tests

lebauce:
run skydive-selenium-tests

PR #982, opened 2018-04-29
k8s: generate defaults when k8s.probes is empty

hunchback:
to test try:

(1)
```
analyzer:
  probes:
  - k8s
k8s:
  probes:
    # emptry
```
In this case we expect probes as listed in LOG to include: container, networkpolicy, namespace, networkpolicy, node, pod

(2)
Use some defaults

```
analyzer:
  probes:
  - k8s

k8s:
  probes: 
    node
```
In this case we expect LOG to indicate that only node probe of k8s is registered

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -136,7 +136,6 @@ func init() {
 	cfg.SetDefault("http.ws.enable_write_compression", true)
 
 	cfg.SetDefault("k8s.config_file", "/etc/skydive/kubeconfig")
-	cfg.SetDefault("k8s.probes", []string{"networkpolicy", "pod", "container", "node", "namespace"})
 
 	cfg.SetDefault("logging.backends", []string{"stderr"})
 	cfg.SetDefault("logging.color", true)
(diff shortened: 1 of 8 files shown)


hunchback:
@nplanel / @safchain / @lebauce - my next PRs depend on this one - kindly review

lebauce:
@hunchback I'm not sure to understand the purpose. If an empty probe list is explicitely specified in the configuration, why would you load a default set of probe. Why not simply commenting the k8s.probe parameter in the configuration ?

hunchback:
@lebauce - note that I have not changed the current behavior only improved implementation such that I no longer need to maintain the list of k8s probes in config/config.go ... but the motivation for defaulting to the list of *all* probes is that I don't want to have to change the skydive.yml on each update of the k8s probe which would add new (sub)probes, and I don't want to relay on having the entire list of probes within skydive.yml for k8s to work ...

hunchback:
I attempted to use:

```
k8s:
  probes:
    - *
```

But this caused YAML parsing error, and using quoting such as `- "*"` is rather ugly and non-intuitive - I have reverted back to an empty list indicating  all (adding a comment to this effect into default YAML file).

safchain:
Need to be rebased properly.

hunchback:
@safchain - rebased

PR #994, opened 2018-05-03
ES : fix configuration

nplanel:
ES specific configuration need to be keep in storage.elasticsearch section

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -75,17 +75,6 @@ analyzer:
     # Max number of flows in write buffer (after which all flows accumulated are dropped)
     # max_flow_buffer_size: 100000
 
-    # If a limit is specified, when the index reaches it, it is rolled.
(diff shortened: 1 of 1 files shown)


nplanel:
run skydive-unit-tests

lebauce:
@nplanel Can you please rebase ?

etc/skydive.yml.default
```diff
@@ -75,17 +75,6 @@ analyzer:
     # Max number of flows in write buffer (after which all flows accumulated are dropped)
     # max_flow_buffer_size: 100000
 
-    # If a limit is specified, when the index reaches it, it is rolled.
```
lebauce: Those parameters still exist, they just need to moved to the right section

nplanel: yes they exist under storage.elastricsearch or storage.newbackend.driver = elasticsearch
but should not exist under analyzer.flow

PR #995, opened 2018-05-03
analyzer.flow : get flow bulk configuration from the backend

nplanel:


etc/skydive.yml.default
```diff
@@ -63,7 +63,8 @@ analyzer:
 
   # Flow storage engine
   flow:
-    # Available: elasticsearch, orientdb, sample
+    # Available: myelasticsearch, myorientdb
```
safchain: I would say : name of the backend definition

etc/skydive.yml.default
```diff
@@ -63,7 +63,8 @@ analyzer:
...
+    # legacy backend/driver : elasticsearch, orientdb
```
safchain: no need to speak about legacy

etc/skydive.yml.default
```diff
@@ -87,7 +88,8 @@ analyzer:
     # indices_to_keep: 0
 
   topology:
-    # graph backend : memory, elasticsearch, orientdb, sample
+    # graph backend : mymemory, myelasticsearch, myorientdb
```
safchain: same here

etc/skydive.yml.default
safchain: and I'll put backend: myelasticsearch

analyzer/flow_server.go
```diff
@@ -232,6 +245,33 @@ func (s *FlowServer) Stop() {
 	}
 }
 
+func (s *FlowServer) setupBulkConfigFromBackend() error {
+	s.bulkInsert = FlowBulkInsertDefault
+	s.bulkInsertDeadline = time.Duration(FlowBulkInsertDeadlineDefault) * time.Second
+
+	backend := config.GetString("analyzer.flow.backend")
+	switch backend {
```
safchain: Doesn't seem to be the last version as this switch should be removed

etc/skydive.yml.default
```diff
@@ -63,8 +63,8 @@ analyzer:
 
   # Flow storage engine
   flow:
-    # Available: elasticsearch, orientdb, sample
-    # backend: elasticsearch
+    # Available: myelasticsearch, myorientdb
```
safchain: please use:

Name of the backend definition, ex: myelasticsearch, mybackend, etc.

otherwise we think that is only those that are available

nplanel: done

PR #1002, opened 2018-05-05
ovs: link orphan ovs interface to bridge

safchain:


safchain:

PR #1010, opened 2018-05-09
fabric: dynamically add fabric node and links

masco:


jonnary:
Hi, masco.  Why we didn't merge this commit?
Maybe we can do this together.

masco-enovance:
Hey Jonnary, thanks for your interest. We have added one generic way to add/edit nodes and edges. with that, we can create fabric nodes also. This is no more needed. I will discard this soon. thanks!
cmd to use the generic way: skydive client node-rule & skydive client edge-rule.

jonnary:
ok. I have test the new patch. Many things need to continue to work about external switch.

lebauce:
Efforts moved to new UI

PR #1014, opened 2018-05-11
Fix selenium img

safchain:


safchain:
rerun selenium-tests

scripts/ci/cleanup.sh
```diff
@@ -31,7 +31,7 @@ function cleanup() {
   cleanup_items ovsdb "ovs-vsctl del-br"
   cleanup_items docker "docker_rm"
   cleanup_items lxd "lxc delete --force"
-  cleanup_items docker-images "docker rmi -f"
+  #cleanup_items docker-images "docker rmi -f"
```
hunchback: Better to remove line

hunchback: Better to remove line

tests/selenium_test.go
```diff
@@ -158,6 +158,9 @@ func TestPacketInjectionCapture(t *testing.T) {
 		return
 	}
 
+	// wait a bit to be sure that the capture is really started because of slow systems.
+	time.Sleep(5 * time.Second)
```
hunchback: Rather retry the Gremlin statement bellow

scripts/ci/cleanup.sh
```diff
@@ -76,7 +76,7 @@ function snapshot() {
   snapshot_items intf $ext "ip -o link show | awk -F': ' '{print \$2}' | cut -d '@' -f 1"
   snapshot_items ovsdb $ext "ovs-vsctl list-br"
   snapshot_items docker $ext "docker ps -a -q"
-  snapshot_items docker-images $ext "docker images -a -q"
+  #snapshot_items docker-images $ext "docker images -a -q"
```
hunchback: Better to delete

tests/selenium_test.go
safchain: I don't think so, as it is to inject and I don't want to inject twice or more

hunchback: I am fine with the above if it works even on very-slow systems ... if this is not the case I would still recommend polling (such as 10 retry / 1s delay) ... is fast on fast systems but will still wait a while to give slow systems a chance to come up.

scripts/ci/cleanup.sh
safchain: will do once I will have something really working

safchain:
rerun selenium-tests

safchain:
rerun unit-tests

safchain:
rerun selenium-tests

safchain:
rerun selenium-tests

safchain:
rerun selenium-tests

PR #1015, opened 2018-05-11
json: wrong json omitempty option in host cpu

masco-enovance:
cpu index is starting from zero but the omitempty option not
allowing to display the first index which is zero


safchain:

masco-enovance:

PR #1017, opened 2018-05-11
CI : fix ansible CI slave deploy

nplanel:


contrib/ansible/roles/skydive_dev/tasks/main.yml
```diff
@@ -51,7 +51,13 @@
   when: ansible_distribution == "Fedora"
 
 - include: protobuf.yml
+  tags:
```
lebauce: Why skipping protobuf.yml and static.yml ? They are required for Jenkins slaves

nplanel: my bad I didn't check that it was removed from script/ci/install*.sh

Issue #1023, opened 2018-05-15
Skydive flow is L2 or L3? (which packets are packed to flow)


eranra:
Up until #991 Skydive didn't use L2 fields to calculate the key and pack packets into flows. #991 added to the key calculation the Link layer. 

This new behavior distinguish flows on L2 basis, so L3 traffic using different L2 path is considered different flow. 

Data saved to storage backend (e.g. elastic) presented in UI etc ... will show correctly info from A--> B but will show zero BA packets, bytes etc (the B-->A data will be part of a different flow).

L3TrackingID exists, but not used for statistics etc ... so we can correlate A-->B and B-->A but not trivial especially in the back-end.

Link to current key calculation::

https://github.com/skydive-project/skydive/blob/master/flow/flow.go#L209

PR #1026, opened 2018-05-15
flow: add option to specify layers used, l2 or l3

safchain:
Fixes: #1023


flow/flow.go
```diff
@@ -94,10 +94,18 @@ type RawPackets struct {
 	RawPackets []*RawPacket
 }
 
+type LayerKeyMode int
+
+const (
+	L2L3KeyMode LayerKeyMode = 0 // uses Layer2 and Layer3 for hash computation, default mode
```
eranra: Hi Sylvain,

Thanks for adding this so quickly :-)

The obvious question, why L2L3KeyMode is default:: 

L2L3KeyMode is not the expected behavior for regular users working on networks. I agree that having the option to always take into account L2 is useful for network experts that understand L2 layer and the fact that L3 traffic can move on different L2. But when you say network flow to people ... most will assume L3 (IP) and most will not want to know about the L2 options that traffic can pass through. Also it makes things more complicated when using  persistence storage, you need to aggregate all L2 flows to get L3 layer info.

Your change to the key that takes into account L2 when L3 doesn't exist makes L3KeyMode very compelling almost to the point that I think that there is no real practical need for the L2L3KeyMode.

One more small idea ... when in L3KeyMode it can be nice to add a small icon, or label in the flow table (UI) that show which key type is used, so user will have clarity on what is a flow L2 based or L3 based.

flow/flow.go
```diff
@@ -94,10 +94,18 @@ type RawPackets struct {
...
+	L3KeyMode   LayerKeyMode = 1 // uses Layer3 only and layer2 if no Layer3
```
eranra: consider chaining the name of the variable to L3PreferredKeyMode just to make it super clear that we still use L2 if L3 doesn't exist

flow/flow.go
safchain: I think that the l2L3 is useful as Skydive is a debugging tool too and in a multipath context you may want to have one flow per path to ensure that the traffic is correctly forwarded. 

While writing this I thought making L3 by default or adding a configuration option to change the default ?

flow/flow.go
safchain: I'll do, thx for the suggestion

cmd/client/capture.go
```diff
@@ -205,6 +207,7 @@ func addCaptureFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVarP(&extraTCPMetric, "extra-tcp-metric", "", false, "Add additional TCP metric to flows, default: false")
 	cmd.Flags().BoolVarP(&ipDefrag, "ip-defrag", "", false, "Defragment IPv4 packets, default: false")
 	cmd.Flags().BoolVarP(&reassembleTCP, "reassamble-tcp", "", false, "Reassemble TCP packets, default: false")
+	cmd.Flags().StringVarP(&layerKeyMode, "layer-key-mode", "", "L3", "Defines the layers uses by flow key calculation, L2 or L3")
```
nplanel: we should keep the same behavior as today (L2)

flow/flow.go
```diff
@@ -107,6 +118,31 @@ type FlowUUIDs struct {
 	L3ID       int64
 }
 
+func (l LayerKeyMode) String() string {
+	if l == 1 {
```
nplanel: l == L2KeyMode

cmd/client/capture.go
nplanel: or you would like to ignore bonding / L2multipath artifacs ?

safchain:
rerun -functional-tests-backend-elasticsearch

safchain:

safchain:

safchain:

safchain:

flow/flow.go
```diff
@@ -94,10 +95,20 @@ type RawPackets struct {
 	RawPackets []*RawPacket
 }
 
+// defines what are the layers used for the flow key calculation
+type LayerKeyMode int
```
nplanel: this type shall be private

PR #1033, opened 2018-05-18
ci : retry upload vagrant boxes

nplanel:


lebauce:
Every command could fail from time to time because of network issues. Why retrying this one and why once ?

nplanel:
basically I seen the job failed twice during the end of the upload
may I need to add --retry to the getter
once because, just lazy reason

lebauce:
@nplanel I've never seen the job fail because of upload (but for many other reasons :-) )

PR #1039, opened 2018-05-22
ui: enable fixed position (X,Y)

vperelmax:
Change in config file:
   #topology:
   # Pin to ceratin location specified nodes
   # Nodes are selected via binary boolean expressions, where
   # '!', '&&', '||' operators are enabled.
   # Atoms of the expressions refer to values of Metadata fields.
      #fixed_nodes:
         #- (Name=tor1): (X=300, Y=300)
         #- (Name=port2): (X=500, Y=500)
As a result in the example: Skydive will pin all the nodes
with "Metadata.Name = tor1" to location [300, 300].
The same about port2.

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -393,6 +393,11 @@ ui:
     # default_highlight: "layer2"
     # Filter Gremlin expression used by default and applied on WebUI load.
     # default_filter: "layer2"
+    # Pre-defined boolean expression to select nodes with certain Metadata and change their view
+    # Pin to ceratin location.
+    # fixed_nodes:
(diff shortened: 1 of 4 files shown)


safchain:
rerun cdd-overview-tests

etc/skydive.yml.default
```diff
@@ -393,6 +393,11 @@ ui:
     # default_highlight: "layer2"
     # Filter Gremlin expression used by default and applied on WebUI load.
     # default_filter: "layer2"
+    # Pre-defined boolean expression to select nodes with certain Metadata and change their view
+    # Pin to ceratin location.
```
safchain: s/ceratin/certain/

PR #1047, opened 2018-05-24
ui:  ensure openflow rule elements are split

pierrecregut:
Some kind of actions are not breakable in smaller
chunks making the table overflow. A typical example
is userdata for controller that can be a large blob.

We replace the use of span with unicode zero length whitespaces to
create breakable points.


pierrecregut:
run tests

safchain:
@pierrecregut so the CI network is unavailable for now :(, I'll rerun the tests once up again

safchain:
run tests

safchain:
rerun unit-tests

PR #1077, opened 2018-06-01
WIP topology: add MQTT probe

safchain:
Just a WIP that listen for memory from collectd

This has to be improved a lot, making the subscribers
generic and the import parsing too.

### DIFF
--- a/agent/probes.go
+++ b/agent/probes.go
@@ -31,6 +31,7 @@ import (
 	"github.com/skydive-project/skydive/topology/graph"
 	"github.com/skydive-project/skydive/topology/probes/docker"
 	"github.com/skydive-project/skydive/topology/probes/lxd"
+	"github.com/skydive-project/skydive/topology/probes/mqtt"
 	"github.com/skydive-project/skydive/topology/probes/netlink"
 	"github.com/skydive-project/skydive/topology/probes/netns"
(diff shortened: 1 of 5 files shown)


yogananth-subramanian:
mqtt TopicName could be made more flexible by using similar configuration variables used by collectd in Skydive config file.
For Example:
     Host "localhost"
     Prefix "collectd"

But with Centos 7.5, "FQDN" is used instead of localhost, event when "Host" is set to "localhost" in collectd.conf

Collectd.conf mqtt plugin config used:
<Plugin "mqtt">
   <Publish "name1">
     Host "localhost"
     Prefix "collectd"
     Retain true
   </Publish>
</Plugin>

collectd version:
collectd-5.8.0-4.el7.x86_64 : Statistics collection daemon for filling RRD files
Repo        : @epel

mosquitto_sub output:
collectd/sk1.rdocloud/memory/memory-used 1529000128.445:5060583424

Centos release:
CentOS Linux release 7.5.1804 (Core)

safchain:
@yogananth-subramanian if the mqtt probe is running on the agent side I think we can ignore the host part of the topic, assuming that the collectd metric are from the localhost.

safchain:
Replaced by the collectd plugin

PR #1080, opened 2018-06-05
K8s edges fix

hunchback:
To check configure all elements for example by:

```
k8s:
  probes:
    # empty
```

Now create the hello-node landscape using:

```
kubectl run hello-minikube --image=k8s.gcr.io/echoserver:1.10 --port=8080
```

Now check that UI is ok (all elements are tied to cluster or indirectly to namespace).

### DIFF
--- a/topology/probes/k8s/namespace.go
+++ b/topology/probes/k8s/namespace.go
@@ -44,25 +44,27 @@ type namespaceProbe struct {
 }
 
 func newObjectIndexer(g *graph.Graph) *graph.MetadataIndexer {
+	ownedByNamespaceFilter := filters.NewOrFilter(
+		filters.NewTermStringFilter("Type", "deployment"),
+		filters.NewTermStringFilter("Type", "daemonset"),
+		filters.NewTermStringFilter("Type", "ingress"),
+		filters.NewTermStringFilter("Type", "job"),
+		filters.NewTermStringFilter("Type", "pod"),
+		filters.NewTermStringFilter("Type", "persistentvolume"),
+		filters.NewTermStringFilter("Type", "persistentvolumeclaim"),
+		filters.NewTermStringFilter("Type", "networkpolicy"),
+		filters.NewTermStringFilter("Type", "replicaset"),
+		filters.NewTermStringFilter("Type", "replicationcontroller"),
+		filters.NewTermStringFilter("Type", "service"),
+		filters.NewTermStringFilter("Type", "statefulset"),
+	)
+
 	filter := filters.NewAndFilter(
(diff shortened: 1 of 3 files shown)


topology/probes/k8s/namespace.go
```diff
@@ -44,25 +44,32 @@ type namespaceProbe struct {
...
+		"networkpolicy",
+		"replicaset",
+		"replicationcontroller",
+		"service",
+		"statefulset",
+	}
+
+	ownedByNamespaceFilter := filters.NewOrFilter()
+	for _, ty := range ownedByNamespace {
+		ownedByNamespaceFilter = filters.NewOrFilter(ownedByNamespaceFilter, filters.NewTermStringFilter("Type", ty))
```
lebauce: The new version creates NewOrFilter(NewOrFilter(NewOrFilter(NewOrFilter(term), term), term), term) which is equivalent but create more objects and is not more readable

hunchback: I now only make one OrFilter but I have left code to generate the "Type is equal to" filter from the type of the object (e.g. "pod")

topology/probes/k8s/namespace.go
```diff
@@ -44,25 +44,32 @@ type namespaceProbe struct {
 }
 
 func newObjectIndexer(g *graph.Graph) *graph.MetadataIndexer {
+	ownedByNamespace := []string{
```
safchain: You could simply do

```
ownedByNamespaceFilter := filters.NewOrFilter(
			filters.NewTermStringFilter("Type", "deployment"),	
			filters.NewTermStringFilter("Type", "daemonset"),	
			filters.NewTermStringFilter("Type", "ingress"),	
			filters.NewTermStringFilter("Type", "job"),	
			filters.NewTermStringFilter("Type", "pod"),	
			filters.NewTermStringFilter("Type", "persistentvolume"),	
			filters.NewTermStringFilter("Type", "persistentvolumeclaim"),	
			filters.NewTermStringFilter("Type", "networkpolicy"),	
			filters.NewTermStringFilter("Type", "replicaset"),	
			filters.NewTermStringFilter("Type", "replicationcontroller"),	
			filters.NewTermStringFilter("Type", "service"),	
			filters.NewTermStringFilter("Type", "statefulset"),	
		)

filter := filters.NewAndFilter(filter := filters.NewAndFilter(
 		filters.NewTermStringFilter("Manager", managerValue),
                ownedByNamespaceFilter)
```

This will keep it readable as you did introducing a variable while avoiding to add a loop + array allocation, etc

A comment would do the job too.

safchain:

safchain:
rerun scale-tests

PR #1097, opened 2018-06-13
K8s hellonode scenario

hunchback:
simply make sure that k8s tests run, when developing the tests I discovered some issues with links:
- cluster --> node
- node --> pod

So for now (until fixing the k8s probe) these were eliminated from the test scenario

### DIFF
--- a/tests/k8s_test.go
+++ b/tests/k8s_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/skydive-project/skydive/common"
 	g "github.com/skydive-project/skydive/gremlin"
 	"github.com/skydive-project/skydive/tests/helper"
+	"github.com/skydive-project/skydive/topology/graph"
 )
 
 func k8sConfigFile(name string) string {
@@ -40,10 +41,10 @@ func k8sConfigFile(name string) string {
 }
 
 const (
-	manager         = "k8s"
-	objName         = "skydive-test"
-	k8sRetry        = 10
-	k8sDelaySeconds = 1
+	manager  = "k8s"
+	objName  = "skydive-test"
+	k8sRetry = 10
+	k8sDelay = time.Second
 )
 
 var (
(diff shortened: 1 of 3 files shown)


tests/k8s_test.go
```diff
@@ -61,54 +62,77 @@ var (
...
+	args := []interface{}{"Namespace", m["Namespace"], "Name", m["Name"]}
+	args = append(args, args1...)
+	return makeHasArgsType(m["Type"], args...)
+}
+
+func queryNodeCreation(c *CheckContext, query g.QueryString) (node *graph.Node, err error) {
+	err = common.Retry(func() error {
+		const expectedNumNodes = 1
+
+		fmt.Printf("Gremlin Query: %s\n", query)
```
safchain: t.Logf instead of Printf passing the `t` variable to the function

tests/k8s_test.go
```diff
@@ -61,54 +62,77 @@ var (
...
+
+		if len(nodes) != expectedNumNodes {
+			return fmt.Errorf("Ran '%s', expected %d node, got %d nodes: %+v", query, expectedNumNodes, len(nodes), nodes)
+		}
+
+		if expectedNumNodes > 0 {
+			node = nodes[0]
+		}
+		return nil
+	}, k8sRetry, k8sDelaySeconds*time.Second)
```
safchain: as k8sDelaySeconds says that it is in second I would put the `8*time.Second` in the constant definition line 47

tests/k8s_test.go
hunchback: fixed

tests/k8s_test.go
hunchback: fixed and renamed `k8sDelaySeconds` --> `k8sDelay`

safchain:
rerun scale-tests

PR #1115, opened 2018-06-20
K8s metadata pod

hunchback:
Adding new fields to Pod:
- IP (optional)
- Status
- Node

The mandatory fields are verified via functional test - so if this passes functional we should be ok

### DIFF
--- a/tests/k8s_test.go
+++ b/tests/k8s_test.go
@@ -112,24 +112,36 @@ func checkEdgeCreation(t *testing.T, c *CheckContext, from, to *graph.Node, relT
 }
 
 /* -- test creation of single resource -- */
-func testNodeCreation(t *testing.T, setupCmds, tearDownCmds []helper.Cmd, typ, name string) {
+func testNodeCreation(t *testing.T, setupCmds, tearDownCmds []helper.Cmd, typ, name string, fields ...string) {
 	test := &Test{
 		mode:         OneShot,
 		retries:      3,
 		setupCmds:    append(tearDownCmds, setupCmds...),
 		tearDownCmds: tearDownCmds,
 		checks: []CheckFunction{func(c *CheckContext) error {
-			_, err := checkNodeCreation(t, c, typ, "Name", name)
-			return err
+			obj, err := checkNodeCreation(t, c, typ, "Name", name)
+			if err != nil {
+				return err
+			}
+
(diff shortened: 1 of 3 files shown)


hunchback:
@safchain / @lebauce - kindly review

topology/probes/k8s/pod.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/skydive-project/skydive/probe"
 	"github.com/skydive-project/skydive/topology/graph"
 
-	api "k8s.io/api/core/v1"
```
nplanel: using api namespace was much better for readability

tests/k8s_test.go
```diff
@@ -112,24 +112,42 @@ func checkEdgeCreation(t *testing.T, c *CheckContext, from, to *graph.Node, relT
...
-			_, err := checkNodeCreation(t, c, typ, "Name", name)
-			return err
+			obj, err := checkNodeCreation(t, c, typ, "Name", name)
+			if err != nil {
+				return err
+			}
+
+			m := obj.Metadata()
+
+			const fieldDoesNotExist = "DoesNotExist"
```
nplanel: how a metadata Node that have been queried will reply "DoesNotExist" ?
This should not append, as the metadata (obj.Metadata()[key]) hash key should not exist

topology/probes/k8s/pod.go
hunchback: but this is not the standard way these packages are accessed from within the k8s code base ... either the default is used such as `v1` or `v1beta1` or if it conflicts with other packages then you append the parent directory such as `corev1` or `metav1beta1`

tests/k8s_test.go
hunchback: ok - removing

topology/probes/k8s/pod.go
nplanel: so when they will switch (and the probe too) to a v2 of the API you want to rename all the code ?

hunchback:
Its' built this way so one can support several versions within a single
code base, as done here

https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go

So when we add v2.Pod support it will be along side v1.Pod (note that each
k8s resource has an independant API versioning)

PR #1116, opened 2018-06-20
netlink: stop probe when unregister

safchain:


topology/probes/netlink/netlink.go
```diff
@@ -1075,6 +1075,8 @@ func (u *NetLinkProbe) Unregister(nsPath string) error {
 			}
 			delete(u.probes, fd)
 
+			probe.stop()
```
nplanel: :+1:

PR #1119, opened 2018-06-20
SELinux: Fedora and Centos/RHEL have different SELinux rules support

nplanel:


lebauce:
run skydive-selenium-tests

PR #1146, opened 2018-07-03
topology: add node/edge apis

masco-enovance:
sample command to create a fabric node:
```
skydive client node create --name="node1" --type="fabric" --query="create:" --metadata="fd=sdf"
skydive client node create --name="node2" --type="fabric" --query="create:" --metadata="fds=sdf"
skydive client edge create --src="g.v().has('Name', 'node1')" --dst="g.v().has('Name', 'node2')" --realtiontype="both"
skydive client edge create --src="g.v().has('Name', 'lo', 'State', 'UP')" --dst="g.v().has('Name', 'node1')" --realtiontype="layer2"
```

### DIFF
--- a/api/server/edgerule.go
+++ b/api/server/edgerule.go
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2018 Red Hat, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package server
+
+import (
+	"github.com/nu7hatch/gouuid"
+
+	"github.com/skydive-project/skydive/api/types"
+	shttp "github.com/skydive-project/skydive/http"
+	"github.com/skydive-project/skydive/topology/graph"
+)
+
+// EdgeRuleResourceHandler describes a edge rule resource handler
+type EdgeRuleResourceHandler struct {
+	ResourceHandler
+}
+
+// EdgeRuleAPI based on BasicAPIHandler
+type EdgeRuleAPI struct {
+	BasicAPIHandler
+	Graph *graph.Graph
+}
+
+// Name returns resource name "edgerule"
+func (erh *EdgeRuleResourceHandler) Name() string {
+	return "edgerule"
+}
+
+// New creates a new EdgeRule
+func (erh *EdgeRuleResourceHandler) New() types.Resource {
+	id, _ := uuid.NewV4()
+
+	return &types.EdgeRule{
+		UUID: id.String(),
+	}
+}
+
+// RegisterEdgeRuleAPI registers an EdgeRule's API to a designated API Server
+func RegisterEdgeRuleAPI(apiServer *Server, g *graph.Graph, authBackend shttp.AuthenticationBackend) (*EdgeRuleAPI, error) {
+	era := &EdgeRuleAPI{
+		BasicAPIHandler: BasicAPIHandler{
+			ResourceHandler: &EdgeRuleResourceHandler{},
+			EtcdKeyAPI:      apiServer.EtcdKeyAPI,
+		},
+		Graph: g,
+	}
+	if err := apiServer.RegisterAPIHandler(era, authBackend); err != nil {
+		return nil, err
+	}
+
+	return era, nil
+}
--- a/api/types/types.go
+++ b/api/types/types.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	shttp "github.com/skydive-project/skydive/http"
+	"github.com/skydive-project/skydive/topology/graph"
 )
 
 // Resource used as interface resources for each API
@@ -189,4 +190,43 @@ type Workflow struct {
 	Description   string          `yaml:"description"`
 	Parameters    []WorkflowParam `yaml:"parameters"`
 	Source        string          `valid:"isValidWorkflow" yaml:"source"`
+}
+
+// NodeRule describes a node rule
+type NodeRule struct {
+	UUID     string
+	Type     string `valid:"nonzero"`
+	Name     string `valid:"nonzero"`
+	Metadata graph.Metadata
+	Action   string `valid:"regexp=^(create|update)$"`
+	Query    string
+}
+
+// ID returns the node rule ID
+func (n *NodeRule) ID() string {
+	return n.UUID
+}
+
+// SetID set ID
+func (n *NodeRule) SetID(id string) {
+	n.UUID = id
+}
+
+// EdgeRule describes a edge rule
+type EdgeRule struct {
+	UUID         string
+	Src          string `valid:"isGremlinExpr"`
+	Dst          string `valid:"isGremlinExpr"`
+	RelationType string `valid:"regexp=^(layer2|ownership|both)$"`
+	Metadata     graph.Metadata
+}
+
+// ID returns the edge rule ID
+func (e *EdgeRule) ID() string {
+	return e.UUID
+}
+
+// SetID set ID
+func (e *EdgeRule) SetID(id string) {
+	e.UUID = id
 }
--- a/cmd/client/edgerule.go
+++ b/cmd/client/edgerule.go
@@ -1,0 +1,188 @@
+/*
+ * Copyright (C) 2018 Red Hat, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package client
+
+import (
+	"os"
+
+	"github.com/skydive-project/skydive/api/client"
+	api "github.com/skydive-project/skydive/api/types"
+	"github.com/skydive-project/skydive/logging"
+	"github.com/skydive-project/skydive/topology/enhancers"
+	"github.com/skydive-project/skydive/topology/graph"
+	"github.com/skydive-project/skydive/validator"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	src          string
+	dst          string
+	relationType string
+)
+
+// EdgeRuleCmd skydive edge rule root command
+var EdgeRuleCmd = &cobra.Command{
+	Use:          "edge-rule",
+	Short:        "edge-rule",
+	Long:         "edge-rule",
+	SilenceUsage: false,
+}
+
+// EdgeRuleCreate skydive edge create command
+var EdgeRuleCreate = &cobra.Command{
+	Use:          "create",
+	Short:        "create",
+	Long:         "create",
+	SilenceUsage: false,
+
+	Run: func(cmd *cobra.Command, args []string) {
+		client, err := client.NewCrudClientFromConfig(&AuthenticationOpts)
+		if err != nil {
+			logging.GetLogger().Error(err.Error())
+			os.Exit(1)
+		}
+
+		m, err := usertopology.DefToMetadata(metadata, graph.Metadata{"RelationType": relationType})
+		if err != nil {
+			logging.GetLogger().Error(err.Error())
+			os.Exit(1)
+		}
+
+		edge := &api.EdgeRule{
+			Src:          src,
+			Dst:          dst,
+			RelationType: relationType,
+			Metadata:     m,
+		}
+
+		if err = validator.Validate(edge); err != nil {
+			logging.GetLogger().Error(err.Error())
+			os.Exit(1)
+		}
+
+		if err = client.Create("edgerule", &edge); err != nil {
+			logging.GetLogger().Error(err.Error())
+			os.Exit(1)
+		}
+
+		printJSON(edge)
+	},
+}
+
+// EdgeRuleGet skydive edge get command
+var EdgeRuleGet = &cobra.Command{
+	Use:          "get",
+	Short:        "get",
+	Long:         "get",
+	SilenceUsage: false,
+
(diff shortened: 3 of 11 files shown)


rbac/policy.csv
```diff
@@ -18,3 +18,7 @@ p, admin, websocket, /ws/flow, allow
 p, admin, websocket, /ws/publisher, allow
 p, admin, websocket, /ws/replication, allow
 p, admin, websocket, /ws/subscriber, allow
+p, admin, topology/node, read, allow
```
lebauce: only the name of the resource is used, so it should be
```
p, admin, node, read, allow
p, admin, node, write, allow
p, admin, edge, read, allow
p, admin, edge, write, allow
```

api/server/server.go
```diff
@@ -187,6 +187,36 @@ func (a *Server) RegisterAPIHandler(handler Handler) error {
 				w.WriteHeader(http.StatusOK)
 			},
 		},
+		{
+			Name: title + "Update",
+			Method: "PUT",
+			Path: shttp.PathPrefix(fmt.Sprintf("/api/%s/", name)),
+			HandlerFunc: func(w http.ResponseWriter, r *auth.AuthenticatedRequest) {
+				resource := handler.New()
```
lebauce:
```
if rbac.Enforce(r.Username, name, "write") == false {
        w.WriteHeader(http.StatusMethodNotAllowed)
        return
}
```

rbac/policy.csv
masco-enovance: i am using the resource name as topology/node and topology/edge

topology/enhancers/topology_manager.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	tm.syncTopology()
+	return nil
+}
+
+func (tm *TopologyManager) handleCreateNode(node *types.Node) error {
+	query := strings.Split(node.Query, ":")
+	if len(query) != 2 {
+		logging.GetLogger().Errorf("Query formate is wrong")
```
lebauce: nit: s/formate/format

topology/enhancers/topology_manager.go
```diff
@@ -0,0 +1,320 @@
...
+		return errors.New("Query formate is wrong")
+	}
+	switch strings.ToLower(query[0]) {
+	case "create":
+		return tm.createNode(node)
+	case "update":
+		return tm.updateMetadata(query[1], node.Metadata)
+	default:
+		logging.GetLogger().Errorf("Query formate is wrong. supported prefixes: create and update")
```
lebauce: ditto

topology/graph/graph.go
```diff
@@ -93,6 +93,7 @@ type MetadataTransaction struct {
 	graph        *Graph
 	graphElement interface{}
 	adds         map[string]interface{}
+	remove       []string
```
lebauce: nit: removes or removals

topology/enhancers/topology_manager.go
```diff
@@ -0,0 +1,320 @@
...
+	case "create", "set":
+		tm.createEdge(edge)
+	case "delete":
+		src := tm.getNodes(edge.Src)
+		dst := tm.getNodes(edge.Dst)
+		if len(src) < 1 || len(dst) < 1 {
+			logging.GetLogger().Errorf("Source or Destination node not found")
+			return
+		}
+		tm.graph.Unlink(src[0], dst[0])
```
lebauce: Unlink removes all the edges so we should probably use:
```
if link := tm.graph.GetFirstLink(src[0], dst[0], graph.Metadata{"RelationType": edge.RelationType); link != nil {
    tm.Graph.DelEdge(link)
}
```

api/types/types.go
```diff
@@ -183,3 +183,41 @@ type Workflow struct {
...
 	Source        string          `valid:"isValidWorkflow" yaml:"source"`
 }
+
+// Node describes a node
+type Node struct {
+	UUID     string
+	Type     string
+	Name     string
+	Metadata string
+	Query    string
```
lebauce: Query has the "<action>:<query>" format, it would be clearer if we have 2 separate values (that also avoid splitting to get both values)

api/types/types.go
```diff
@@ -183,3 +183,41 @@ type Workflow struct {
...
+	n.UUID = id
+}
+
+// Edge describes a edge
+type Edge struct {
+	UUID         string
+	Src          string
+	Dst          string
+	RelationType string
+	Metadata     string
```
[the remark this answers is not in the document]
masco-enovance: from UI we can't do this right?

topology/enhancers/user_metadata.go
```diff
@@ -20,7 +20,7 @@
  *
  */
 
-package metadata
+package user_topology
```
lebauce: Nit: according to the [Golang docs](https://golang.org/doc/effective_go.html#names) :  By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps. So this could be `usertopology` for example.

topology/graph/graph.go
```diff
@@ -904,6 +910,10 @@ func (t *MetadataTransaction) Commit() {
 		}
 	}
 
+	for _, k := range t.removes {
+		common.DelField(e.metadata, k)
```
lebauce: It would be nice if `DelField` could also return a boolean that would be true if an element was really deleted and false otherwise. That way, we would not generate an event if the metadata were not modified.

lebauce:
I thought I told it in my previous review but it seems I did not, sorry. I think it would be clearer if the resources were `noderule` and `edgerule` (instead of `node` and `edge`) as we are not really directly creating nodes and edges but rules that define nodes and edges. What do you think ?

rbac/policy.csv
lebauce:
```
p, admin, noderule, read, allow
p, admin, edgerule, read, allow
```

topology/graph/graph.go
masco-enovance: even an inbuilt delete function will not return any boolean value. we need to implement the logic. so I am thinking of a separate PR. is that fine for you?

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
 	Parameters    []WorkflowParam `yaml:"parameters"`
 	Source        string          `valid:"isValidWorkflow" yaml:"source"`
 }
+
+// Node describes a node
+type Node struct {
```
lebauce: s/Node/NodeRule/

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
...
+	return n.UUID
+}
+
+// SetID set ID
+func (n *Node) SetID(id string) {
+	n.UUID = id
+}
+
+// Edge describes a edge
+type Edge struct {
```
lebauce: s/Edge/EdgeRule/

cmd/client/node.go
```diff
@@ -0,0 +1,184 @@
...
+// TopologyNodeCmd skydive node root command
+var TopologyNodeCmd = &cobra.Command{
+	Use:          "node-rule",
+	Short:        "node-rule",
+	Long:         "node-rule",
+	SilenceUsage: false,
+}
+
+// NodeCreate skydive node create command
+var NodeCreate = &cobra.Command{
```
lebauce: s/NodeCreate/NodeRuleCreate/

cmd/client/node.go
```diff
@@ -0,0 +1,184 @@
...
+var (
+	nodeName string
+	nodeType string
+	metadata string
+	query    string
+	action   string
+)
+
+// TopologyNodeCmd skydive node root command
+var TopologyNodeCmd = &cobra.Command{
```
lebauce: s/TopologyNodeCmd/NodeRuleCmd/

cmd/client/node.go
```diff
@@ -0,0 +1,184 @@
...
+			logging.GetLogger().Error(err.Error())
+			os.Exit(1)
+		}
+
+		printJSON(node)
+	},
+}
+
+// NodeGet skydive node get command
+var NodeGet = &cobra.Command{
```
lebauce: s/NodeGet/NodeRuleGet/

cmd/client/node.go
```diff
@@ -0,0 +1,184 @@
...
+		if err := client.Get("noderule", args[0], &node); err != nil {
+			logging.GetLogger().Error(err.Error())
+			os.Exit(1)
+		}
+		printJSON(&node)
+	},
+}
+
+// NodeList skydive node list command
+var NodeList = &cobra.Command{
```
lebauce: s/NodeList/NodeRuleList/

cmd/client/node.go
```diff
@@ -0,0 +1,184 @@
...
+		if err := client.List("noderule", &nodes); err != nil {
+			logging.GetLogger().Error(err.Error())
+			os.Exit(1)
+		}
+		printJSON(nodes)
+	},
+}
+
+// NodeDelete skydive node delete command
+var NodeDelete = &cobra.Command{
```
lebauce: s/NodeDelete/NodeRuleDelete/

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
...
 	Source        string          `valid:"isValidWorkflow" yaml:"source"`
 }
+
+// Node describes a node
+type Node struct {
+	UUID     string
+	Type     string `valid:"nonzero"`
+	Name     string `valid:"nonzero"`
+	Metadata string
+	Action   string `valid:"regexp=^(create|delete)$"`
```
lebauce: It seems like it should be create|update no ?

cmd/client/edge.go
```diff
@@ -0,0 +1,180 @@
+/*
```
lebauce: Should probably be edgerule.go

topology/enhancers/topology_manager.go
```diff
@@ -0,0 +1,320 @@
...
+		pair = strings.TrimSpace(pair)
+
+		kv := strings.Split(pair, "=")
+		if len(kv)%2 != 0 {
+			return nil, fmt.Errorf("attributes must be defined by pair k=v: %v", def)
+		}
+		key := strings.Trim(kv[0], `"`)
+		value := strings.Trim(kv[1], `"`)
+
+		metadata[key] = value
```
lebauce: common.SetField(metadata, key, value)

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
 	Parameters    []WorkflowParam `yaml:"parameters"`
 	Source        string          `valid:"isValidWorkflow" yaml:"source"`
 }
+
+// Node describes a node
+type Node struct {
+	UUID     string
+	Type     string `valid:"nonzero"`
+	Name     string `valid:"nonzero"`
+	Metadata string
```
lebauce: What about making Metadata a map[string]interface{}, so we just have to call defToMetadata at creation time ?

cmd/client/edge.go
```diff
@@ -0,0 +1,180 @@
...
+		for _, id := range args {
+			if err := client.Delete("edgerule", id); err != nil {
+				logging.GetLogger().Error(err.Error())
+			}
+		}
+	},
+}
+
+func addCreateEdgeFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&src, "src", "", "", "src node")
```
lebauce: We should probably say that "src" and "dst" must be Gremlin expressions

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
...
+
+// SetID set ID
+func (n *Node) SetID(id string) {
+	n.UUID = id
+}
+
+// Edge describes a edge
+type Edge struct {
+	UUID         string
+	Src          string `valid:"nonzero"`
```
lebauce: valid:"isGremlinExpr"

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
...
+	Dst          string `valid:"nonzero"`
```
lebauce: valid:"isGremlinExpr"

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
 	Parameters    []WorkflowParam `yaml:"parameters"`
 	Source        string          `valid:"isValidWorkflow" yaml:"source"`
 }
+
+// Node describes a node
```
lebauce: s/Node/NodeRule/

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
...
+type NodeRule struct {
+	UUID     string
+	Type     string `valid:"nonzero"`
+	Name     string `valid:"nonzero"`
+	Metadata string
+	Action   string `valid:"regexp=^(create|update)$"`
+	Query    string
```
lebauce: `valid:"isGremlinExpr"

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
...
+}
+
+// ID returns the node ID
```
lebauce: s/node/node rule/

api/types/types.go
```diff
@@ -190,3 +190,42 @@ type Workflow struct {
...
+func (n *NodeRule) ID() string {
+	return n.UUID
+}
+
+// SetID set ID
+func (n *NodeRule) SetID(id string) {
+	n.UUID = id
+}
+
+// Edge describes a edge
```
lebauce: // EdgeRule describes an edge rule

cmd/client/edgerule.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	src          string
+	dst          string
+	relationType string
+)
+
+// TopologyEdgeRuleCmd skydive edge root command
```
lebauce: s/TopologyEdgeRuleCmd/EdgeRuleCmd/

cmd/client/noderule.go
```diff
@@ -0,0 +1,184 @@
...
+
+var (
+	nodeName string
+	nodeType string
+	metadata string
+	query    string
+	action   string
+)
+
+// TopologyNodeRuleCmd skydive node root command
```
lebauce: s/TopologyNodeRuleCmd/NodeRuleCmd/

topology/graph/graph.go
lebauce: Now that DelField returns a boolean, we can do :
`updated = common.DelField(e.metadata, k) || updated`

api/server/edgerule.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"github.com/nu7hatch/gouuid"
+
+	"github.com/skydive-project/skydive/api/types"
+	shttp "github.com/skydive-project/skydive/http"
+	"github.com/skydive-project/skydive/topology/graph"
+)
+
+// TopologyEdgeResourceHandler describes a edge resource handler
+type TopologyEdgeResourceHandler struct {
```
lebauce: s/Edge/EdgeRule/

api/server/edgerule.go
```diff
@@ -0,0 +1,72 @@
...
+	ResourceHandler
+}
+
+// TopologyEdgeAPI based on BasicAPIHandler
```
lebauce: s/Edge/EdgeRule/

api/types/types.go
masco-enovance: this is optional field when creating node rule. so if we validate it will fail when it is empty.

cmd/client/noderule.go
```diff
@@ -0,0 +1,192 @@
...
+	SilenceUsage: false,
+
+	Run: func(cmd *cobra.Command, args []string) {
+		client, err := client.NewCrudClientFromConfig(&AuthenticationOpts)
+		if err != nil {
+			logging.GetLogger().Error(err.Error())
+			os.Exit(1)
+		}
+
+		m, err := usertopology.DefToMetadata(metadata, graph.Metadata{"Name": nodeName})
```
lebauce: 1/ DefToMetadata is only used by in the "cmd/client" package so I suggest moving it to it, so you don't have to export it.
2/ Graph.Metadata{} should be empty here, as you are giving the node the name of the rule

PR #1148, opened 2018-07-03
etcd: fix embedded etcd custom port

safchain:
The port used by the client if no
server list was provided was hard coded
and not using the listen port.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -53,6 +53,10 @@ var (
 	}
 )
 
+const (
+	etcdDefaultPort = 12379
+)
+
 func init() {
 	host, err := os.Hostname()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


lebauce:

safchain:

config/config.go
```diff
@@ -98,7 +102,7 @@ func init() {
 	cfg.SetDefault("etcd.data_dir", "/var/lib/skydive/etcd")
 	cfg.SetDefault("etcd.embedded", true)
 	cfg.SetDefault("etcd.name", host)
-	cfg.SetDefault("etcd.listen", "127.0.0.1:12379")
+	cfg.SetDefault("etcd.listen", fmt.Sprintf("127.0.0.1:%d", etcdDefaultPort))
```
masco-enovance: nit: string concatenation won't be sufficient?

safchain: that's string and integer, so will have to convert it anyway

PR #1172, opened 2018-07-11
graph: Add origin field to graph element

masco-enovance:


topology/graph/graph.go
```diff
@@ -1155,10 +1156,15 @@ func (g *Graph) GetNode(i Identifier) *Node {
 }
 
 func newNode(i Identifier, m Metadata, t time.Time, h string) *Node {
+	o := h
+	if nodeType, ok := m["Type"]; ok {
```
safchain: Sorry Masco, I haven't been enough clear, That is not the type of the node but the type of the service, agent/analyzer.

The ES schema should be adapted probably, and some other stuff, serialization, etc.

masco-enovance: Got it Sylvain. I will change according to that.

PR #1150, opened 2018-07-04
k8s: extend netpol for namespaces

hunchback:
Now netpol object is associated with both pods and namespaces:
* networkpolicy --association --> namespace
* networkpolicy --association --> pod

So that association does not conflict with ownership networkpolicy objects are owned by the cluster (not namespace).

The code includes two new test scenarios which verify association links have been created correctly.

NOTE: this PR depends on first merging: https://github.com/skydive-project/skydive/pull/1178
NOTE: this PR depends on first merging: https://github.com/skydive-project/skydive/pull/1182

PR #1178, opened 2018-07-12
K8s tests retry

hunchback:
Eliminated entire-check retry mechanism (3 x 1sec) ; so now only relaying on a per gremlin-query retry (3 x 10sec). As having both retrys set would cause a multiplication of retries possibly having 9 retry attempt - with the new code only 3 retry attempts at max.


hunchback:
@lebauce @safchain - I plan to rebase https://github.com/skydive-project/skydive/pull/1150 on top of this

PR #1189, opened 2018-07-19
docker/ci: use proxy only if http_proxy is set

grooverdan:
I'm assuming the proxy is set as env http_proxy in the ci environment


masco-enovance:
ok to test

PR #1206, opened 2018-07-24
netlink: close fds if we can't poll events

lebauce:


lebauce:
run skydive-compile-tests

lebauce:

PR #1211, opened 2018-07-24
topology: don't alter Layer2Metadata pseudo constant

lebauce:


lebauce:
run skydive-unit-tests

PR #1162, opened 2018-07-06
split agent / analyzer and api auth

safchain:

PR #1229, opened 2018-07-31
agent: print each analyzer connection

nplanel:


agent/agent.go
```diff
@@ -111,7 +111,8 @@ func (a *Agent) GetStatus() interface{} {
 	}
 
 	analyzers := make(map[string]AnalyzerConnStatus)
-	for id, status := range a.analyzerClientPool.GetStatus() {
+	for _, status := range a.analyzerClientPool.GetStatus() {
+		id := fmt.Sprintf("%s:%d", status.Addr, status.Port)
```
safchain: If there is collision on the id this will not fix the issue as the collision will be in GetStatus function

safchain: this should help
https://github.com/skydive-project/skydive/pull/1162/commits/047f77d855d6f1b6f85094fe7abc6db69ea13ab6#diff-50b3a20b4d0d5dfb0e488c26e828dd39R95

safchain:
can we close this one ?

PR #1231, opened 2018-07-31
k8s: fix node to host link handling

lebauce:


agent/host.go
```diff
@@ -54,7 +55,11 @@ type CPUInfo struct {
 // createRootNode creates a graph.Node based on the host properties and aims to have an unique ID
 func createRootNode(g *graph.Graph) (*graph.Node, error) {
 	hostID := config.GetString("host_id")
-	m := graph.Metadata{"Name": hostID, "Type": "host"}
+	hostname, err := os.Hostname()
+	if err != nil {
+		return nil, err
```
hunchback: I rather you set `hostname = ""` and continue adding the root node, would make it more robust

topology/probes/k8s/node.go
```diff
@@ -116,9 +116,20 @@ func (c *nodeProbe) OnDelete(obj interface{}) {
 	}
 }
 
+func (c *nodeProbe) OnNodeAdded(n *graph.Node) {
+	hostname, _ := n.GetFieldString("Hostname")
+	if t, _ := n.GetFieldString("Type"); t == "host" {
```
hunchback: this condition is redundant, your indexer only triggers events when `"Type" == "host"`

agent/host.go
lebauce: We already `panic` - on purpose - if we can't retrieve the hostname in config/config.go
If we can't get the hostname, I'm pretty sure nothing will work :-)

topology/probes/k8s/node.go
lebauce: It was a precaution but a useless one. Removed

agent/host.go
hunchback: fine

PR #1241, opened 2018-08-05
k8s: service metadata

hunchback:
To test run the below and verify using UI:

```
kubectl create -f tests/k8s/statefulset.yaml
```

PR #1251, opened 2018-08-14
K8s service links

hunchback:
create links between `service` to `pod`, can be tested running:

```
kubectl create -f tests/k8s/service-pod.yaml
```

*NOTE*: depends on https://github.com/skydive-project/skydive/pull/1241

### DIFF
--- a/topology/probes/k8s/service.go
+++ b/topology/probes/k8s/service.go
@@ -30,13 +30,18 @@ import (
 	"github.com/skydive-project/skydive/topology/graph"
 
 	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/tools/cache"
 )
 
 type serviceProbe struct {
 	defaultKubeCacheEventHandler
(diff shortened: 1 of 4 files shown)


topology/probes/k8s/service.go
```diff
@@ -78,19 +134,46 @@ func (p *serviceProbe) OnDelete(obj interface{}) {
...
-		if nsNode := p.graph.GetNode(serviceUID(srv)); nsNode != nil {
-			p.graph.DelNode(nsNode)
+		if srvNode := p.graph.GetNode(serviceUID(srv)); srvNode != nil {
+			p.graph.DelNode(srvNode)
 			logging.GetLogger().Debugf("Deleted %s", dumpService(srv))
 		}
 	}
 }
 
+func (p *serviceProbe) onNodeUpdated(podNode *graph.Node) {
```
nplanel: you should remove the link to the pod when a node is Deleted : OnNodeDeleted(n *Node)

hunchback: But deletion is handled anyway: https://github.com/skydive-project/skydive/blob/master/topology/graph/graph.go#L1298

nplanel: yes, I forgot that

PR #1253, opened 2018-08-16
devmode: explicitly install iptables 1.6.1

tomerlange:
We need iptables 1.6.1 as when using iptables 1.6.2 then minikube 0.25.2 fails (it has an issue with parameters passed to iptables-restore).

You can test this fix by creating a new devmode:

```
export PREPARE_BOX=true
vagrant up
vagrant ssh
vagrant@dev $ iptables -v
# expected to get 1.6.1 and not 1.6.2
```

Note that after accepting this PR you need to build a new skydive/skydive-dev image as the current skydive-dev image has iptables 1.6.2 bundled into it.

### DIFF
--- a/contrib/ansible/roles/skydive_dev/tasks/main.yml
+++ b/contrib/ansible/roles/skydive_dev/tasks/main.yml
@@ -9,6 +9,7 @@
     name: "{{ item }}"
     state: present
   with_items:
+     - iptables-1.6.1                   # iptables-1.6.2 not compatible with minikube-0.25.2
      - git
      - wget
      - unzip


lebauce:
add to whitelist

lebauce:
run skydive-create-vagrant-boxes

lebauce:
ok to test

nplanel:
run tests

nplanel:
@hunchback we can't do that, as people that deploy Skydive expect to run on latest iptables package, the bug is on minikube, so minikube need to be fixed. Downgrade iptables workaround is not a solution
One day or another we will need latest/updated minikube version

hunchback:
@nplanel the goal of devmode is to provide a development platform, as such it is opinionated for example using f27 and using specific versions of packages which work well with the test/development tools such as iptables serving the minikube tool, so downgrading to iptables 1.6.1 does not limit any possibilities to deploy skydive with the latest iptables on any target platform of choice (fc27, fc28, ubuntu, etc. each with possibly a different version of iptables)...

Now minikube when run in "none" driver mode has a very tight integration with the OS on which it is running - and 0.25.2 requires iptables 1.6.1. We have made some initial attempts to upgrade to a newer minikube atop fc27 (which is an "old" distro) these attempts have failed after reasonable effort (this is not to say its' impossible only that it is not trivial and will entail other changes), more so I expect *anyway* that regardless switching to fc28 will require tweaking/upgrading minikube installation ... so I am not seeing a significant benefit for getting our minikube installation to work with iptables 1.6.2.

Moreover iptables 1.6.1 *was* previously used by devmode, only that it was silently upgraded to 1.6.2 due to changes in the fc27 RPM repositories ... so why was 1.6.1 ok before and is not ok now? 1.6.1 is still a totally supported option for fc27 - thus switching back to iptables 1.6.1 does not break anything - as such *kindly approve this PR as currently devmode is broken for k8s (helm, istio) development.*

Thanks!

nplanel:
@hunchback First we will switch to FC28 ASAP because we will keep track on the latest version, for obvious reason.
2nd, if FC27 switched to iptables 1.6.2 there a good reason, it's parts from the updates repo

3rd, _"thus switching back to iptables 1.6.1 does not break anything"_  how to you known ?
As workaround for fc27 + buggy minikube, it was once a iptables wrapper script (on one of the PR) to avoid this issue, isn't it ?

Last, we will work with the latest distribution version, this problem is not new (... months)
So the right fix would be to update (fc28 and minikube) to work on up to date version.
Minikube will not fix the dependency rules of what we should use, he has a bug, has been fixed, let's update it ... it's not only devmode, it's about someone using Skydive and Minikube (latest)

PR #1257, opened 2018-08-21
ui: k8s netpol deny uses an X

hunchback:
you can use:

```
kubectl create -f tests/k8s/networkpolicy-ingress-deny.yaml
```

the `deny` will be represented as a cross (X)


hunchback:
@lebauce / @nplanel - please review/merge

PR #1259, opened 2018-08-23
minikube: upgrade to v0.28.2

tomerlange:
To test just ensure k8s test passes.

### DIFF
--- a/scripts/ci/install-minikube.sh
+++ b/scripts/ci/install-minikube.sh
@@ -6,11 +6,11 @@ OS=linux
 ARCH=amd64
 TARGET_DIR=/usr/bin
 
-MINIKUBE_VERSION="v0.25.2"
+MINIKUBE_VERSION="v0.28.2"
 MINIKUBE_URL="https://github.com/kubernetes/minikube/releases/download/$MINIKUBE_VERSION/minikube-$OS-$ARCH"
 
-KUBECTL_VERSION="v1.9.4"
-KUBECTL_URL="https://storage.googleapis.com/kubernetes-release/release/$KUBECTL_VERSION/bin/$OS/$ARCH/kubectl"
+K8S_VERSION="v1.10.0"
+KUBECTL_URL="https://storage.googleapis.com/kubernetes-release/release/$K8S_VERSION/bin/$OS/$ARCH/kubectl"
 
 WITH_CALICO=true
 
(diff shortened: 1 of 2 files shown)


nplanel:
@tomerlange could you revert this patch in your PR 4e8848ae804574fffc717fb1e86b6d40280a5fdd

hunchback:
@nplanel - this change is orthogonal to PR downgrading iptables, the reason we changed to the new version of minikube was to address a race condition (@tomerlange - kindly add references to that issue). In any case as iptables 1.6.1 works well with the new minikube 0.28.2 and as iptable 1.6.1 is a standard part of fc27 I see no reason to change.

In future in the context of switching to fc28 (which officially only uses iptables 1.6.2) we would anyway upgrade to iptables 1.6.2 and then in that context could check the current minikube/kubernetes compatibility to iptables 1.6.2.

nplanel:
@hunchback iptables-1.6.1 is workaround .... so if minikube has been updated, we switch back to the latest distribution package

hunchback:
@tomerlange - kindly check if the new minikube works with iptables-1.6.2 - if so then add a 2nd commit which reverses https://github.com/skydive-project/skydive/commit/4e8848ae804574fffc717fb1e86b6d40280a5fdd

tomerlange:
DONE

lebauce:
@tomerlange @hunchback This PR broke the CI because it switched to localkube, so it was reverted. Please accept the invitation to the Github organization so that the tests are triggered for your PRs, then submit a new PR with the suggested changes. Thanks !

scripts/ci/install-minikube.sh
```diff
@@ -101,8 +101,9 @@ stop() {
 start() {
         check_minikube
 
+	local args="--kubernetes-version $K8S_VERSION --bootstrapper=localkube"
```
lebauce: nit: please fix indentation.

In addition, this should be:
```
local args="--kubernetes-version $K8S_VERSION"
```

scripts/ci/install-minikube.sh
```diff
@@ -101,8 +101,9 @@ stop() {
...
         if [ "$MINIKUBE_DRIVER" == "none" ]; then
-                local args="--vm-driver=none"
+                args="$args --vm-driver=none"
```
lebauce: Here you can do
```
args="$args --vm-driver=none --bootstrapper=localkube"
```

PR #1260, opened 2018-08-23
Istio probe stub

hunchback:
This is an initial skeletal istio probe (sharing code with the k8s probe). Immediately will follow PRs adding CI tests for istio probe, later followed by adding istio resources sub-probes. 

To activate the probe setup `skydive.yml` with:

```
analyzer:
  probes:
  - istio
```

Next run as analyzer:

```
sudo $(which skydive) analyzer -c skydive.yml
```


hunchback:
run tests

topology/graph/graph.go
```diff
@@ -1461,3 +1462,40 @@ func NewBackendByName(name string, etcdClient *etcd.Client) (backend GraphBacken
 	}
 	return backend, nil
 }
+
+// DumpNode dumps major node fields
+func DumpNode(n *Node) string {
```
lebauce: This function should go to k8s, no the `graph` package

topology/graph/graph.go
```diff
@@ -1461,3 +1462,40 @@ func NewBackendByName(name string, etcdClient *etcd.Client) (backend GraphBacken
...
+	return fmt.Sprintf("%s:%s{Namespace: %s, Name: %s}", manager, ty, namespace, name)
+}
+
+// DumpLink dumps the link and parent, child nodes
+func DumpLink(parent, child *Node, m Metadata) string {
+	return fmt.Sprintf("%s -> %s: %s", DumpNode(parent), DumpNode(child), m)
+}
+
+// AddLinkTry if the link does not already exist then add it
+func AddLinkTry(g *Graph, parent, child *Node, m Metadata) *Edge {
```
lebauce: I don't really like this pattern :

topology/probes/k8s/graph.go
```diff
@@ -33,22 +31,18 @@ import (
...
+	detailsField = "K8s"
+	managerValue = "k8s"
 )
 
 func lookupChildren(g *graph.Graph, parent, childFilter *graph.Node, edgeFilter graph.Metadata) []*graph.Node {
 	return g.LookupChildren(parent, childFilter.Metadata(), edgeFilter)
 }
 
-func newMetadata(ty, namespace, name string, details interface{}) graph.Metadata {
+func NewMetadata(manager, ty, namespace, name string, details interface{}) graph.Metadata {
```
lebauce: Why exporting this function ?

topology/probes/k8s/graph.go
```diff
@@ -57,84 +51,86 @@ func newMetadata(ty, namespace, name string, details interface{}) graph.Metadata
...
+}
+
+func AddMetadata(g *graph.Graph, n *graph.Node, details interface{}) {
 	tr := g.StartMetadataTransaction(n)
 	tr.AddMetadata(detailsField, common.NormalizeValue(details))
 	tr.Commit()
 }
 
-func newEdgeMetadata() graph.Metadata {
+func NewEdgeMetadata(manager string) graph.Metadata {
```
lebauce: Why exporting this function ?

topology/graph/graph.go
hunchback: ok

topology/graph/graph.go
hunchback: ok

topology/probes/k8s/graph.go
hunchback: used by `topology/probes/istio`

topology/probes/k8s/graph.go
hunchback: used by `topology/probes/istio`

topology/probes/k8s/probe.go
```diff
@@ -106,14 +88,40 @@ func (p *Probe) Stop() {
 	p.bundle.Stop()
 }
 
-// NewProbe create the Probe for tracking k8s events
-func NewProbe(g *graph.Graph) (*Probe, error) {
+// NewProbeHelper create the Probe for tracking events
+func NewProbeHelper(g *graph.Graph, context string, m *ProbeMap) (*Probe, error) {
```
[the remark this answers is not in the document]
hunchback: used by `topology/probes/istio`

hunchback:
@lebauce - it will be shortly followed by a PR which adds a first real istio object - as it is each of these PRs is still relatively large.

hunchback:
run skydive-scale-test

hunchback:
run skydive-scale-tests

etc/skydive.yml.default
```diff
@@ -417,6 +418,12 @@ k8s:
   - statefulset
   - storageclass
 
+istio:
```
lebauce: This should be in :
```
analyzer:
  topology:
    istio:
```

like for the `fabric` probe.
By the way, we should do the same for the `k8s` probe

lebauce:
@hunchback I prefer to have a single PR with :
- one commit with the simpliest Istio probe that still does something useful
- one commit with the istio install script
- one istio to run the tests
(this is just an example)

There is no way to express dependency between PRs on Github:
- You have to put the dependency as a comment on the PR, which is not optimal
- You have to include the commits of the PRs you depend on
- Tests are not triggered for the `lowest` PR
- This is very hard to follow

The way to go - as long as it's not tens thousands of lines of code - is to have a PR and multiple commits in the PR.

etc/skydive.yml.default
hunchback: Lets' discuss and handle separately for both k8s and istio , once we are aligned on this I'll issue a separate PR.

hunchback:
@lebauce - even after these three steps the istio probe is not actually very useful ... and will not be useful until significant functionality is added (due 1-2 month from today) ... so rather than aiming for minimally useful functionality I rather target:
(1) additional functionality (non necessarily useful)
(2) no depredations

This now enables me to provide PRs which can easily be reviewed, and processed as they are relatively well defined, in our case:
(1) minimal istio probe --> focus review on the k8s refactoring which enabled istio not to have to duplicate code
(2) istio tests --> focus review again of how I refactored k8s/helm scripts to allow adding istio
(3) destinationrole istio probe --> focus review on how I integrated wrappers from kiali enabling access to istio objects

Now while Git does not support dependencies, by adding commits and placing a note/link to base PRs we indicate PR should not be merged (but can get feedback) or if merged we ensure all it  depends on is also brought in.

lebauce:
@hunchback I'm sorry but reviewing a probe that does nothing is not helpful.
Same than installing a software when we don't even test it was properly installed.

When I was saying "useful", I meant : at least does something that talks to Istio. So, in my opinion, what you did in https://github.com/skydive-project/skydive/pull/1265 would be a self contained PR that makes sense and that can be reviewed and tested

PR #1282, opened 2018-09-06
Topology api validator changes 

masco-enovance:
new gremlin or empty validator
new name description fields for rules
name and type changed to node-name and node-type respectively


masco-enovance:
run skydive-ppc64le-tests

PR #1283, opened 2018-09-07
host: Reporting DMI structures

ErwanAliasr1:
This patch is about adding DMI capabilities to skydive.

It offer users an option to get what are the hardware capabilities of a particular node.
This usually includes the vendor/product/serial numbers, bios versions, DIMM detection and many other items.
A simple laptop usually expose more than 600 items.

There was several options to make the implementation:
- ochapman/godmi
    it offer nice structures but is really outdate (4yo)
    we need accurate DMI information for supporting recent systems

- go-smbios
    easy to use but doesn't provides any human readable output
    it only reports the row values that have to be analyzed against the DMI specification of that particular version
    A sample output looks like : &{{7 19 1} [1 128 1 32 0 32 0 64 0 64 0 0 5 4 7] [L1-Cache]}

- paulstuart/dmijson
    it parses dmidecode but goes through a json structure which isn't really needed here

- dselans/dmidecode
    this is the one selected for this patch.
    it parses dmidecode and offer simple structures to parse
    then we recreate a dynamic structure to add into the metadata
    Pros:
    - it uses dmidecode which is up to date regarding the specification, vendors are usually contributing too
    Cons:
    - it parses dmidecode output, which is string based and could break over time

This light patch was enough to report the full DMI tables into the metadata.

Signed-off-by: Erwan Velu <erwan@redhat.com>

### DIFF
--- a/agent/host.go
+++ b/agent/host.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/dselans/dmidecode"
 	"github.com/shirou/gopsutil/cpu"
 	"github.com/shirou/gopsutil/host"
 
@@ -132,6 +133,53 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
 	}
 	if hostInfo.VirtualizationRole != "" {
 		m.SetField("VirtualizationRole", hostInfo.VirtualizationRole)
+	}
+
+	dmi := dmidecode.New()
+	if err := dmi.Run(); err == nil {
+		// If we have a valid DMI table
+		dmi_data := map[string]interface{}{}
+
+		// Let's scan every DMI type
+		for type_number := 0; type_number < 128; type_number++ {
+			// For each DMI type
+			byTypeData, byTypeErr := dmi.SearchByType(type_number)
+
+			// Only consider the populated types
+			if byTypeErr == nil {
+				// For each entry of that type
+				for _, typeData := range byTypeData {
+					// The structure will represent that type
+					dmi_data_array := []map[string]interface{}{}
+
+					// For every entry of that type
+					for dmi_label, dmi_value := range typeData {
+						// Some value are sub fields like the Characteristics in Type 0
+						// Values are splitted by tabulations
+						dmi_value_array := strings.Split(dmi_value, string(9)+string(9))
+
+						if len(dmi_value_array) > 1 {
+							// Extract all sub fields if any
+							// Compute the sub dynamic structure label : value
+							dmi_data_array = append(dmi_data_array, map[string]interface{}{
+								dmi_label: dmi_value_array,
+							})
+						} else {
+							// Compute the dynamic structure like label : value
+							dmi_data_array = append(dmi_data_array, map[string]interface{}{
+								dmi_label: dmi_value,
+							})
+						}
+					}
+					// Add the computed value for the given type
+					//DMIName represent the human readable name of the current type
+					// Like "Bios Information" for type 0
+					dmi_data[typeData["DMIName"]] = dmi_data_array
+				}
+				// Adding all types to the main DMI item
+				m.SetField("DMI", dmi_data)
+			}
+		}
 	}
 
 	return g.NewNode(graph.GenID(), m), nil
--- a/contrib/packaging/rpm/skydive.spec
+++ b/contrib/packaging/rpm/skydive.spec
@@ -85,6 +85,7 @@ Requires:         %{name} = %{version}-%{release}
 Requires(post):   systemd %{selinux_semanage_pkg}
 Requires(preun):  systemd
 Requires(postun): systemd %{selinux_semanage_pkg}
+Requires:         dmidecode
 
 %description agent
 The Skydive agent has to be started on each node where the topology and
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -931,6 +931,12 @@
 			"revision": "5d2041e26a699eaca682e2ea41c8f891e1060444"
 		},
 		{
+			"checksumSHA1": "D2Aijfk0Ccbw5ftQF3RVTTSPQbk=",
+			"path": "github.com/dselans/dmidecode",
+			"revision": "65c3f9d819108e993ede783bdecbc7d28748cbd9",
+			"revisionTime": "2018-08-14T05:30:09Z"
+		},
+		{
 			"checksumSHA1": "g3z4plpw9F/ho3hdJb+X/bN/OgE=",
 			"path": "github.com/emicklei/go-restful",
 			"revision": "68c9750c36bb8cb433f1b88c807b4b30df4acc40",


lebauce:
add to whitelist

agent/host.go
```diff
@@ -28,6 +28,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/dselans/dmidecode"
```
safchain: you need to "govendor fetch github.com/dselans/dmidecode" to get this package added to vendor/vendor.json

ErwanAliasr1: ok

ErwanAliasr1: solved

safchain:
run tests

lebauce:
ok to test

safchain:
run tests

lebauce:
add to whitelist

ErwanAliasr1:
Cool, tests passes

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
 
+	dmi := dmidecode.New()
+	if err := dmi.Run(); err == nil {
+		// If we have a valid DMI table
+		dmi_data := map[string]interface{}{}
+
+		// Let's scan every DMI type
+		for type_number := 0; type_number < 128; type_number++ {
+			// For each DMI type
+			byTypeData, byTypeErr := dmi.SearchByType(type_number)
```
hunchback: rather reuse `err` as we process each error immediately no need to have a per operation error variable.

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
+
+			// Only consider the populated types
+			if byTypeErr == nil {
```
hunchback: rather:

```
if byTypeErr != nil {
        continue
}
```

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
+				// For each entry of that type
+				for _, typeData := range byTypeData {
+					// The structure will represent that type
+					dmi_data_array := []map[string]interface{}{}
```
hunchback: s/dmi_data_array/dmiDataArray/

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
+
+					// For every entry of that type
+					for dmi_label, dmi_value := range typeData {
```
hunchback: s/dmi_label/dmiLabel/
s/dmi_value/dmiValue/

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
+						// Some value are sub fields like the Characteristics in Type 0
+						// Values are splitted by tabulations
+						dmi_value_array := strings.Split(dmi_value, string(9)+string(9))
```
hunchback: s/dmi_value_array/dmiValueArray/

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
+			if byTypeErr == nil {
+				// For each entry of that type
+				for _, typeData := range byTypeData {
+					// The structure will represent that type
+					dmi_data_array := []map[string]interface{}{}
+
+					// For every entry of that type
+					for dmi_label, dmi_value := range typeData {
+						// Some value are sub fields like the Characteristics in Type 0
+						// Values are splitted by tabulations
```
hunchback: s/splitted/split/

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
+							})
+						} else {
+							// Compute the dynamic structure like label : value
+							dmi_data_array = append(dmi_data_array, map[string]interface{}{
+								dmi_label: dmi_value,
+							})
+						}
+					}
+					// Add the computed value for the given type
+					//DMIName represent the human readable name of the current type
```
hunchback: Add a space after //

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
 		m.SetField("VirtualizationRole", hostInfo.VirtualizationRole)
 	}
 
+	dmi := dmidecode.New()
+	if err := dmi.Run(); err == nil {
+		// If we have a valid DMI table
+		dmi_data := map[string]interface{}{}
```
safchain: nit: should always use camel case

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
+					// For every entry of that type
+					for dmi_label, dmi_value := range typeData {
+						// Some value are sub fields like the Characteristics in Type 0
+						// Values are splitted by tabulations
+						dmi_value_array := strings.Split(dmi_value, string(9)+string(9))
+
+						if len(dmi_value_array) > 1 {
+							// Extract all sub fields if any
+							// Compute the sub dynamic structure label : value
+							dmi_data_array = append(dmi_data_array, map[string]interface{}{
```
safchain: I would prefer to not add "Unknown" value it adds overhead for nothing ?

safchain: We should try to convert number to int64 values otherwise we will have to request them as string value

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
 		m.SetField("VirtualizationRole", hostInfo.VirtualizationRole)
 	}
 
+	dmi := dmidecode.New()
+	if err := dmi.Run(); err == nil {
+		// If we have a valid DMI table
+		dmi_data := map[string]interface{}{}
+
+		// Let's scan every DMI type
+		for type_number := 0; type_number < 128; type_number++ {
```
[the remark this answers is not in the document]
hunchback: https://linux.die.net/man/8/dmidecode states that there are vendor specific types from 128 to 255, should these not be handled?

agent/host.go
```diff
@@ -134,5 +135,52 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
 		m.SetField("VirtualizationRole", hostInfo.VirtualizationRole)
 	}
 
+	dmi := dmidecode.New()
+	if err := dmi.Run(); err == nil {
```
safchain: i would prefer "log warn" explain that the DMI info will not be reported in case of error due to dmidecode error

agent/host.go
safchain: any reason to not simply 

```
for _, byTypeData := range dmi.Data {
```
 instead of using SearchByType ?

ErwanAliasr1:
Closing is favor of #1307

PR #1295, opened 2018-09-11
api: add node and edge metadata validator making use of JSON schemas

lebauce:


cmd/client/noderule.go
```diff
@@ -67,7 +67,10 @@ var NodeRuleCreate = &cobra.Command{
 			os.Exit(1)
 		}
 
-		m, err := usertopology.DefToMetadata(metadata, graph.Metadata{})
+		m, err := usertopology.DefToMetadata(metadata, graph.Metadata{
+			"Name": nodeName,
```
masco-enovance: while doing update these two fields are optional and we are not going to create any node. so we need to skip this ```isValidNodeMetadata``` check.

masco-enovance: while doing update these two fields are optional and we are not going to create any node. so we need to skip this ```isValidNodeMetadata``` check.

lebauce:
run skydive-ppc64le-tests

PR #1303, opened 2018-09-12
topologyapi: fix delete edge

masco-enovance:


topology/enhancers/topology_manager.go
```diff
@@ -242,9 +242,8 @@ func (tm *TopologyManager) handleEdgeRuleRequest(action string, resource types.R
 			logging.GetLogger().Errorf("Source or Destination node not found")
 			return
 		}
-		if link := tm.graph.GetFirstLink(src[0], dst[0], graph.Metadata{"RelationType": edge.RelationType}); link != nil {
-			tm.graph.DelEdge(link)
-		}
+
+		tm.graph.Unlink(src[0], dst[0])
```
lebauce: With this change, we could delete edges that were not created by the TopologyManager
Why do you need this change ?

masco-enovance: via edge rule, if we create a link by setting ```RelationType``` as ```both``` it will create two links(layer2 and ownership) but while deleting it is deleting only one link.
but your point also a valid one. let me come up with some different solution.
Thanks for the review!

masco-enovance:
run skydive-scale-tests

PR #1307, opened 2018-09-13
agent/host: Using lshw to report hardware

ErwanAliasr1:
This commit is about using lshw tooling to report the hardware
information of nodes.

The JSON output of lshw is parsed and a little bit adjusted to rename
'children' nodes into user-readable types.

As lshw provides much more cpu info that the actual code,
the existing CPU reporting is only used when lshw is not available.

Signed-off-by: Erwan Velu <erwan@redhat.com>

### DIFF
--- a/agent/host.go
+++ b/agent/host.go
@@ -23,9 +23,11 @@
 package agent
 
 import (
+	"encoding/json"
 	"errors"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/shirou/gopsutil/cpu"
@@ -50,6 +52,62 @@ type CPUInfo struct {
 	Mhz        int64  `json:"Mhz,omitempty"`
 	CacheSize  int64  `json:"CacheSize,omitempty"`
 	Microcode  string `json:"Microcode,omitempty"`
+}
+
+// extractClassAndID to report class and ID from the children node
+func extractClassAndID(children map[string]interface{}) (string, string) {
+	var foundID string
+	var foundClass string
+	for objectName, objectValue := range children {
+		switch objectValue.(type) {
+		case string:
+			if objectName == "class" {
+				foundClass = objectValue.(string)
+			}
+			if objectName == "id" {
+				foundID = objectValue.(string)
+			}
+			// Once we discover the class and ID, let's return them
+			if (len(foundID) > 0) && (len(foundClass) > 0) {
+				return foundClass, foundID
+			}
+		}
+	}
+	return "", ""
+}
+
+// parseLshw to rename children nodes inside the lshw data structure
+func parseLshw(items map[string]interface{}) {
+	// For each object of the LSHW structure
+	for objectName, objectValue := range items {
+		switch objectValue.(type) {
+		case []interface{}:
+			// Only consider object that have children
+			if objectName != "children" {
+				continue
+			}
+			// When we have children
+			for _, child := range objectValue.([]interface{}) {
+				_, validChildType := child.(map[string]interface{})
+				if validChildType {
+					// For each child
+					class, ID := extractClassAndID(child.(map[string]interface{}))
+					// Extract from the child the class and ID to rename it
+					if len(class) > 0 {
+						if items[class] == nil {
+							items[class] = []map[string]interface{}{}
+						}
+						// Let's create a new child with the name of the class and add an entry with the ID and point it to the actual child
+						items[class] = append(items[class].([]map[string]interface{}), map[string]interface{}{ID: child})
+						// Delete the actual entry so we renamed it
+						delete(items, objectName)
+						// Do the same parsing with the current child as he could have some 'children' nodes
+						parseLshw(child.(map[string]interface{}))
+					}
+				}
+			}
+		}
+	}
 }
 
 // createRootNode creates a graph.Node based on the host properties and aims to have an unique ID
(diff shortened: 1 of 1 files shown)


agent/host.go
```diff
@@ -52,6 +54,59 @@ type CPUInfo struct {
...
+
+// Parse the LSHW data structure to rename children nodes
+func parseLshw(items map[string]interface{}) {
+	// For each object of the LSHW structure
+	for objectName, objectValue := range items {
+		switch objectValue.(type) {
+		case []interface{}:
+			// Only consider object that have children
+			if objectName != "children" {
+				return
```
safchain: return not continue ?

agent/host.go
```diff
@@ -52,6 +54,59 @@ type CPUInfo struct {
...
+			}
+			// When we have children
+			for _, child := range objectValue.([]interface{}) {
+				// For each child
+				class, ID := extractClassAndID(child.(map[string]interface{}))
```
safchain: As there is no guaranty for the lshw output I think we should check all the type assertion like

```
childType, ok :=  child.(map[string]interface{})
if ok {
```

agent/host.go
```diff
@@ -52,6 +54,59 @@ type CPUInfo struct {
...
+				// Extract from the child the class and ID to rename it
+				if len(class) > 0 {
+					if items[class] == nil {
+						items[class] = []map[string]interface{}{}
+					}
+					// Let's create a new child with the name of the class and add an entry with the ID and point it to the actual child
+					items[class] = append(items[class].([]map[string]interface{}), map[string]interface{}{ID: child})
```
safchain: same here

agent/host.go
```diff
@@ -52,6 +54,59 @@ type CPUInfo struct {
...
+					// Delete the actual entry so we renamed it
+					delete(items, objectName)
```
safchain: Maybe id and class could be removed too from the final structure, it seems to be useless after transformation ?

agent/host.go
ErwanAliasr1: Woot. Fixed.

agent/host.go
```diff
@@ -52,6 +54,59 @@ type CPUInfo struct {
 	Microcode  string `json:"Microcode,omitempty"`
 }
 
+// Extract class and ID from the children node
+func extractClassAndID(children map[string]interface{}) (string, string) {
+	var foundID string
+	var foundClass string
+	for objectName, objectValue := range children {
+		switch objectValue.(type) {
```
hunchback: More idiomatic writting:

```
switch objectValue := objectValue.(type) {
case string:
        switch objectName {
        "class": 
                  ...
        "id":
                  ....
         }
          ...
default:
          // log error
}
```

agent/host.go
```diff
@@ -52,6 +54,59 @@ type CPUInfo struct {
...
+			// Once we discover the class and ID, let's return them
+			if (len(foundID) > 0) && (len(foundClass) > 0) {
+				return foundClass, foundID
+			}
+		}
+	}
+	return "", ""
+}
+
+// Parse the LSHW data structure to rename children nodes
```
hunchback: comment should start with function name

agent/host.go
```diff
@@ -134,5 +163,39 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
 
+	var lshwMap map[string]interface{}
+	lshw, err := exec.Command("lshw", "-quiet", "-json").Output()
+	if err == nil {
+		err = json.Unmarshal(lshw, &lshwMap)
+		if err == nil {
+			parseLshw(lshwMap)
+			m.SetField("Hardware", lshwMap)
+		}
+	} else {
```
hunchback: maybe we should always populate "CPU" field regardless of if we managed to extract data via lshw

agent/host.go
```diff
@@ -134,5 +163,39 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
 		m.SetField("VirtualizationRole", hostInfo.VirtualizationRole)
 	}
 
+	var lshwMap map[string]interface{}
+	lshw, err := exec.Command("lshw", "-quiet", "-json").Output()
+	if err == nil {
+		err = json.Unmarshal(lshw, &lshwMap)
+		if err == nil {
+			parseLshw(lshwMap)
+			m.SetField("Hardware", lshwMap)
```
hunchback: if you use:

```
m.SetFieldAndNormalize("Hardware", lshwMap)
```

you may be able to eliminate the need for calling parseLshw()

ErwanAliasr1: I'm sorry I don't get why. I need to rename some children node. That's is very lshw specific.

agent/host.go
ErwanAliasr1: If we get more info from lshw, what would be the need of having duplicated info ?

agent/host.go
ErwanAliasr1: I was wondering the same. I though that keeping the ID/CLASS as members of a child might help some to get all needed info at the same "level" of the tree. Make sense ?

agent/host.go
ErwanAliasr1: I created items[class]  in the proper type. So the test seems harmless here, isn't it ?

agent/host.go
ErwanAliasr1: Yup, good catch. Fixed.

agent/host.go
safchain: you're right, test is useless

agent/host.go
safchain: I would prefer to remove them as it will reduce the amount of data in the graph/data store and the size of the messages. If needed we could simply introduce a custom marshal function to add them on fly or even re-introduce them ?

safchain:

agent/host.go
hunchback: what i don't like is that the fields (starting from top-level node -  "CPU") change, as a result is i write a Gremlin Query (such as to retrieve nodes with X86 architecture) - the query will not hold. One alternative (to what i proposed above) is to use lshw to enrich the "CPU" metadata node

ErwanAliasr1: That is opening an interesting point. How much the structure should be predictable ?
Does this means it would be better to "reduce" the number of reported information or format it into predefined structures ? @lebauce @safchain ?

agent/host.go
ErwanAliasr1: Yes, the switch is a remaining of the first implementation but have the interest of being more flexible if I have to reprocess some other structures. So yes it's a bit overkill for the current code but useful for testing/adding new features or quirks.

agent/host.go
safchain: I would prefer to have predefined structure so that we guaranty the Skydive output, better to do gremlin request, etc.

nplanel:
It would be better to use a golang library as will do not relay on external binary (ex : when skydive running in container ...)
https://github.com/jaypipes/ghw

PR #1312, opened 2018-09-13
Disable elasticsearch by default in devstack plugin

mangelajo:
Use the memory backend instead and save memory on the default
deployment. Most people is confused and believes that elasticsearch
is a mandatory component of a deployment, while it's not.

With this change I hope I can drive a wider usage in the openstack community.


safchain:
Hi Miguel, I'm OK with this change, thanks. 

We do have CI job to test devstack with elasticsearch and you change will break this job. Can you please enable elasticsearch in this file
https://github.com/skydive-project/skydive/blob/master/scripts/ci/devstack/run-devstack.sh#L80
so that elasticsearch will still be tested. Thanks

mangelajo:
Oh, thank you very much. I will do that.

---
irc: ajo / mangelajo
Miguel Angel Ajo Pelayo
+34 636 52 25 69
skype: ajoajoajo

mangelajo:
Done :)

safchain:
run devstack-tests

PR #1279, opened 2018-09-05
K8s refactoring

lebauce:

PR #1318, opened 2018-09-16
k8s: netpol convert policy for namespace to pods

hunchback:
Prior to this change network-policy would either connect pods or namespaces, but so as to facilitate ipRange which may be used alongside namespaceSelector (as in the below policy) we need to revert to network-policy to always connect pods. (as ipRange my need to intersect with namespaceSelector). So now rather than connecting namespace we will connect all the pods within the respective namespace(s).

All tests have been converted to always create pods within namespaces and then check that edges have been created correctly between network-policy and pod(s).

Note that ipRange support has not been added within this PR and will be added in a future PR.

To test the correctness of this change run the k8s test.

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: test-network-policy
  namespace: default
spec:
  podSelector:
    matchLabels:
      role: db
  policyTypes:
  - Ingress
  ingress:
  - from:
    - ipBlock:
        cidr: 172.17.0.0/16
        except:
        - 172.17.1.0/24
    - namespaceSelector:
        matchLabels:
          project: myproject
```

### DIFF
--- a/tests/k8s/networkpolicy-egress-allow.yaml
+++ b/tests/k8s/networkpolicy-egress-allow.yaml
@@ -1,3 +1,14 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: skydive-test-networkpolicy-egress-allow
+spec:
+  containers:
+  - name: nginx
+    image: nginx
+    ports:
+    - containerPort: 80
+---
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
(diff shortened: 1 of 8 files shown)


hunchback:
run skydive-unit-tests

hunchback:
was integrated into: https://github.com/skydive-project/skydive/pull/1279

PR #1320, opened 2018-09-17
ui: fix node selector with workflows

safchain:


safchain:
rerun skydive-ppc64le-tests

PR #1321, opened 2018-09-18
docker/ci: create-docker-image.sh remove bashism

grooverdan:
Running under dash generates the error:

scripts/ci/create-docker-image.sh: 20: scripts/ci/create-docker-image.sh: [[: not found

Replaced as the pattern /*/* doesn't match the D1 case is left unchanged:

$ D1=skydive/skydive
$ D2=registry.ng.bluemix.net:8080/skydive/skydive
$ echo ${D1%/*/*}
skydive/skydive
$ echo ${D2%/*/*}
registry.ng.bluemix.net:8080


grooverdan:
run skydive-create-docker-image

PR #1326, opened 2018-09-18
Fix afpacket timeout and capture deletion

safchain:


safchain:
run skydive-k8s-tests

PR #1341, opened 2018-09-28
[openflow] Openflow group tables support

pierrecregut:
The Openflow probe did not display group tables because there was no obvious way to monitor them. Those two commits add support for groups. The hard part is in the probe where we use a mostly hidden functionality of OpenFlow 1.4. As one can expect group support will only work for OF1.5 enabled switch (we cannot safely enable it for 1.4 due to a bug in openvswitch corrected last week).

### DIFF
--- a/topology/probes/ovsdb/ovs_of.go
+++ b/topology/probes/ovsdb/ovs_of.go
@@ -25,10 +25,15 @@ package ovsdb
 import (
 	"bufio"
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
+	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -63,14 +68,17 @@ type OvsOfProbe struct {
 // highest priority hides the other rules. It is important to handle rules with the same rawUUID as a group
 // because ovs-ofctl monitor does not report priorities.
 type BridgeOfProbe struct {
-	Host       string             // The global host
-	Bridge     string             // The bridge monitored
-	UUID       string             // The UUID of the bridge node
-	Address    string             // The address of the bridge if different from name
-	BridgeNode *graph.Node        // the bridge node on which the rule nodes are attached.
-	OvsOfProbe *OvsOfProbe        // Back pointer to the probe
-	Rules      map[string][]*Rule // The set of rules found so far grouped by rawUUID
-	cancel     context.CancelFunc
+	Host           string               // The global host
+	Bridge         string               // The bridge monitored
+	UUID           string               // The UUID of the bridge node
+	Address        string               // The address of the bridge if different from name
+	BridgeNode     *graph.Node          // the bridge node on which the rule nodes are attached.
+	OvsOfProbe     *OvsOfProbe          // Back pointer to the probe
+	Rules          map[string][]*Rule   // The set of rules found so far grouped by rawUUID
+	Groups         map[uint]*graph.Node // The set of groups found so far
+	groupMonitored bool
+	ctx            context.Context
+	cancel         context.CancelFunc
 }
 
 // Rule is an OpenFlow rule in a switch
@@ -84,6 +92,14 @@ type Rule struct {
 	UUID     string // Unique id
 }
 
+// Group is an OpenFlow group in a switch
+type Group struct {
+	ID        uint   // Group id
+	GroupType string // Group type
+	Contents  string // Anything else (buckets + selection)
+	UUID      string // Unique id
+}
+
 // Event is an event as monitored by ovs-ofctl monitor <br> watch:
 type Event struct {
 	RawRule *Rule   // The rule from the event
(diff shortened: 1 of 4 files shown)


topology/probes/ovsdb/ovsdb.go
```diff
@@ -129,6 +127,7 @@ func (o *OvsdbProbe) OnOvsBridgeAdd(monitor *ovsdb.OvsMonitor, uuid string, row
 			}
 		}
 	}
+	o.Graph.Unlock()
```
lebauce: The graph is still used in `OnOvsBridgeAdd` so should we stay under lock. Maybe we should just Unlock and RLock

topology/probes/ovsdb/ovs_of.go
```diff
@@ -556,10 +826,11 @@ func (o *OvsOfProbe) OnOvsBridgeAdd(bridgeNode *graph.Node) {
 	bridgeName, _ := bridgeNode.GetFieldString("Name")
 	bridgeProbe, err := o.NewBridgeProbe(hostName, bridgeName, uuid, bridgeNode)
 	if err == nil {
-		logging.GetLogger().Infof("Probe added for %s on %s (%s)", bridgeName, hostName, uuid)
+		logging.GetLogger().Debugf("Probe added for %s on %s (%s)", bridgeName, hostName, uuid)
 		o.BridgeProbes[uuid] = bridgeProbe
 	} else {
 		logging.GetLogger().Errorf("Cannot add probe for bridge %s on %s (%s)", bridgeName, hostName, uuid)
```
lebauce: logging.GetLogger().Errorf("Cannot add probe for bridge %s on %s (%s) : %s", bridgeName, hostName, uuid, err)

topology/probes/ovsdb/ovs_of.go
```diff
@@ -521,9 +781,19 @@ func (o *OvsOfProbe) NewBridgeProbe(host string, bridge string, uuid string, bri
...
 
-	return probe, probe.monitor(ctx)
+	err := probe.monitor(ctx)
+	if err != nil {
+		cancel()
+		return nil, err
+	}
+	err = probe.monitorGroup(ctx)
+	if err != nil {
+		logging.GetLogger().Errorf("Cannot add group probe on %s - %s", bridge, err.Error())
```
lebauce: nit: s/err.Error()/err/

topology/probes/ovsdb/ovsdb.go
pierrecregut: I have put locks on "atomic" graph operations like adding a node. OnOvsBridgeAdd does much more than that and locks are not recursive in go. So I think releasing the lock is sensible but using an explicit unlock is more error prone than defer. Another solution would be to push the unlock defer rlock to the callback but relocking for an immediate unlock is weird.

pierrecregut:
run skydive-ppc64le-tests

pierrecregut:
run skydive-functional-tests-backend-elasticsearch insanelock

pierrecregut:
run skydive-functional-tests-backend-elasticsearch  insanelock

topology/probes/ovsdb/ovsdb.go
pierrecregut: Functional tests detected a deadlock due to recursive use of graph lock. Invariants maintained here: graph is only locked on elementary actions related to it (no system stuff done while the lock is held). openflow probe lock must be locked *before* graph lock

pierrecregut: Race removed when we add  a group just after the creation of the bridge. Takes also into account the fact that the bridge does not recognized OF1.5 when it is created, only after a set bridge protocols=...

pierrecregut:

pierrecregut:
run skydive-k8s-tests

topology/probes/ovsdb/ovs_of.go
```diff
@@ -329,7 +386,7 @@ func launchContinuousOnSwitch(ctx context.Context, cmd []string) (<-chan string,
 
 				return nil
 			}
-			if err := common.Retry(retry, 100, 10*time.Millisecond); err != nil {
+			if err := common.Retry(retry, 100, 500*time.Millisecond); err != nil {
```
safchain: 50 sec seems to be too long ? Is there anything special requiring to wait so long time ?

topology/probes/ovsdb/ovs_of.go
```diff
@@ -506,6 +686,112 @@ func (probe *BridgeOfProbe) monitor(ctx context.Context) error {
 	return nil
 }
 
+const delayForFile = 30
+
+// Check if a file is created in time
+func waitForFile(path string) error {
```
safchain: should be replace by comm.Retry

topology/probes/ovsdb/ovs_of.go
```diff
@@ -506,6 +686,112 @@ func (probe *BridgeOfProbe) monitor(ctx context.Context) error {
...
+	}
+	return err
+}
+
+// monitorGroup monitors the openflow groups of a bridge by
+// launching a goroutine. It uses OpenFlow 1.4 ForwardRequest
+// command that is not directly available in ovs-ofctl
+//
+// Note: Openflow15 is really needed. Receiving an insert_bucket on an OF14
+// monitor crashes the switch (yes, the switch itself) on OVS 2.9
```
safchain: I like this comment :) it is scary but funny

topology/probes/ovsdb/ovs_of.go
```diff
@@ -548,23 +846,27 @@ func (o *OvsOfProbe) OnOvsBridgeAdd(bridgeNode *graph.Node) {
...
 	hostName := o.Host
 	bridgeName, _ := bridgeNode.GetFieldString("Name")
 	bridgeProbe, err := o.NewBridgeProbe(hostName, bridgeName, uuid, bridgeNode)
 	if err == nil {
-		logging.GetLogger().Infof("Probe added for %s on %s (%s)", bridgeName, hostName, uuid)
+		logging.GetLogger().Debugf("Probe added for %s on %s (%s)", bridgeName, hostName, uuid)
 		o.BridgeProbes[uuid] = bridgeProbe
 	} else {
 		logging.GetLogger().Errorf("Cannot add probe for bridge %s on %s (%s)", bridgeName, hostName, uuid)
+		logging.GetLogger().Error(err.Error())
```
safchain: nit: logging.GetLogger().Error(err)

topology/probes/ovsdb/ovs_of.go
pierrecregut: No but 1s may be too short on a slow machine if multiple bridges are created. This is the propagation delay between the admin state of the database and the real bridge in the switch daemon. Let us choose 5s

topology/probes/ovsdb/ovs_of.go
pierrecregut: and could be joined with previous line.

PR #1352, opened 2018-10-03
Openflow json

pierrecregut:
Schema and parser for openflow rules.

### DIFF
--- a/contrib/jsonof/Makefile
+++ b/contrib/jsonof/Makefile
@@ -1,0 +1,9 @@
+
+.PHONY: all
+all: jsonof
+
+jsonof: jsonof.go
+	govendor build jsonof.go
+
+clean:
+	rm -f jsonof
--- a/contrib/jsonof/jsonof.go
+++ b/contrib/jsonof/jsonof.go
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2018 Orange
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package main
+
+import (
+	"bufio"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/skydive-project/skydive/logging"
+	"github.com/skydive-project/skydive/topology/probes/ovsdb/jsonof"
+)
+
+func main() {
+	var err error
+	var result []interface{}
+	var jsBytes []byte
+	gFlag := flag.Bool("g", false, "parse input as groups not rules")
+	fileFlag := flag.String("f", "-", "name of file to parse")
+	prettyFlag := flag.Bool("p", false, "prettyfy the output")
+	flag.Parse()
+	cin := os.Stdin
+	if *fileFlag != "-" {
+		cin, err = os.Open(*fileFlag)
+		if err != nil {
+			logging.GetLogger().Errorf("Cannot open file %s: %s", *fileFlag, err)
+			os.Exit(1)
+		}
+	}
+	scanner := bufio.NewScanner(cin)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if *gFlag {
+			ast, err2 := jsonof.ToASTGroup(line)
+			if err2 != nil {
+				logging.GetLogger().Error(err)
+				continue
+			}
+			result = append(result, ast)
+		} else {
+			ast, err2 := jsonof.ToAST(line)
+			if err2 != nil {
+				logging.GetLogger().Error(err)
+				continue
+			}
+			result = append(result, ast)
+		}
+	}
+	if *prettyFlag {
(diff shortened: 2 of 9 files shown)


pierrecregut:
Just for discussion on the proposed Json schema.
It only covers rules, not groups and is pretty verbose. 
The parser is not ready yet.

pierrecregut:
The updated version modify the schema to make it less verbose (there is still a **very high overhead**). Note that the schema does not really distinguish values and functions in actions. It would be hard to do without a well known set of actions that we would then need to maintain.

The unit tests should give a good idea of what one can expect as output.

There is still no link with skydive. If we use it, it will mean that we have an unparser in the front-end.
We also need to add support for *parsing groups* (and handling meters would be nice too at least at the level of schemas).

pierrecregut:
openflow-json-overhead

Here is a chart of the ratio of the size of the json encoding and the original encoding for the 44000 rules extracted from the openvswitch test suite. The mean is 2.17

Similar results can be obtained for groups but the test suite is much shorter:
openflow-json-overhead-groups

safchain:
@pierrecregut Thanks for this WIP patch and the inputs, I'm going to start playing with it.

pierrecregut:
Last update adds a third commit that enables the new openflow -> json parser in the openflow probe.
As it impacts the metadata of rules, ui and functional tests have been updated.

pierrecregut:

lebauce:
@pierrecregut Thank a lot for this, it definitely makes querying rules with Gremlin easier.

We recently introduced the possibility to define a schema for metadata in https://github.com/skydive-project/skydive/pull/1427 (this commit shows how to convert an existing probe https://github.com/skydive-project/skydive/pull/1427/commits/aa24aff1903895d714df20ba4439d072e4c564ba). This way, you could define a Go struct for an OpenFlow rule.

Regarding the structure, I would prefer to not create an other `Meta` level inside the metadata. `duration`, `reset_counts`, ... could be at the same level than 'priority`, `cookie` and others. For flags such as `reset_counts`, they could be converted to booleans instead of using a key/value with an empty value.

It would be nice if they could be converted to camel case (`Priority`, `Cookie`, `ResetCounts`) as it's the convention we have been following so far for metadata.

Maybe in a latter PR, we could also allow the user to use the `Metrics` step on OpenFlow rules so that we can do this such as `g.V().Has('Type, 'ofrule', .....).Metrics().Aggregates() to get and graph the metrics for a set of rules.

For filters, it would be more user friendly to rename "k" to "Name", "v" to "Value", and "m" to "Match".
For actions, we could rename "f" to "Name", and "a" to "Args".

This way, we could write nicer Gremlin queries such as :
```
g.V().Has('Type', 'ofrule', 'OpenFlow.Priority', GT(150), 'OpenFlow.Action.Name', 'resubmit')
```

pierrecregut:
@lebauce please keep in mind that the probe does not understand Openflow. The Meta struct is probably misnamed (Extra ?) but I don't think its content should be mixed with the part Priority/Table/Filters/Actions that is well defined. For example if you transform the "n" field of Meta to a key in a dictionnary, you will loose the ability to type the schema. For similar reason, flags cannot be transformed into booleans: we just do not know they exist when they are not present. If the reason to do so are Gremlin limitations, we should think of an xpath equivalent for gremlin (JSONPath ?).

Do not forget the overhead of this proposal (see my graph above). My first try used full words as name/value/mask. From my point of view it proved to be too verbose (ratio size json/size original was much higher). 

For capitalized field names, no problem even if it feels like a go convention pushed to json. 
Do we want to have our format adopted outside skydive ? An idea proposed at the hackaton was to push the json format description to the openvswitch community so that the tools generate it directly when they are used with the right flag.

Gremlin filters will be possible but remember that filters are lists and actions are often trees. Can we have list predicates like any, forall on lists with Has ? Again, we need some kind of JSONPath.

safchain:
@pierrecregut I think having "real name" for the fields makes sense even with a the lite overhead, it will bring readability and will help users to understand what they see. For the JSONPath why not but I would start making this usable even with a reduced perimeter. 

Pushing the json to the ovs community sounds good but maybe we should try to use it/improve it to make sure that it fits our needs. (not only ours)
I think some of the fields in the schema could be enforced as their name and type are known (priority, etc.), but that can be done later probably.

pierrecregut:
I will rebase it on top of #1499 when it is merged.
Regarding the name fields, I am glad HTML did not impose "Bold", "Anchor" or "Italic" for common tag names.

lebauce:
@pierrecregut :-) That's a - probably - poor design choice on our side. But being consistent is important otherwise the user needs to know for every field if it's uppercase or lowercase

safchain:
I'm glad iptables imposes upper case for tables, or that c, javascript etc are case sensitive :) Please keep in mind that there are other contributors and that we have to maintain the whole project for a long term and we try to keep things consistant.

pierrecregut:
The rambling was rather against long names that become distracting rather than informative when they are repeated too much (look at the unit tests).
The only problem with capitalized fields is that it is not javascript conventions. Conventions tend to come back in unexpected ways (see metadata vs Metadata...). Also it would be easier to make it acceptable by another community if we want to submit our format.

safchain:
This not about javascript, this is about API (gremlin), etc. We should not consider the WebUI thing here. I had a look a the unit test and that is why I think that is important to make them readable otherwise I don't really know what "f" means. Regarding the format that we would send to ovs, there will be changes for sure requested by them and we can't predict them so I'll be focus on our current use case then if the schema change we will adapt.

lebauce:
@pierrecregut I pushed a PR (https://github.com/skydive-project/skydive/pulls) on top of your branch. It introduces a pure Go OpenFlow parser in addition to the ovs-ofctl one. Your PR is still marked as WIP, is it still true ? Thanks

pierrecregut:
I think we can remove the WIP. But I commited a last change : the schema was out of sync with the produced json (It still used short names).

pierrecregut:
All the commits have been merged in #1474, I am not sure I understand why but there is probably no reason to keep it open.

lebauce:
@pierrecregut ovs-ofctl and native versions share a bunch of code. It was also more convenient to test to ensure there was no regression with the native version. It was therefore a lot easier to put both version in the same PR

PR #1371, opened 2018-10-10
K8s istio merge

hunchback:
the istio probe will latch on the k8s objects (cluster, namespace, etc.)

to test verify there are no regressions (using k8s test)


hunchback:

hunchback:
run skydive-k8s-tests

safchain:
run skydive-k8s-tests

hunchback:
run skydive-ppc64le-tests

hunchback:
run skydive-k8s-tests

etc/skydive.yml.default
```diff
@@ -121,6 +121,7 @@ analyzer:
       # list of (sub) probes comprising k8s probe.
       # if list is empty then will resolve to all existing (sub) probes.
       probes:
+        - cluster
```
nplanel: it's enabled by default and will stay like that
why breaking everything .... where the feature ?

nplanel:
please add real functional tests on a running k8s
no test = no merge

hunchback:
@lebauce - can you check why k8s tests are not running?

topology/graph/indexer.go
```diff
@@ -203,14 +203,9 @@ func NewMetadataIndexer(g *Graph, listenerHandler ListenerHandler, m ElementMatc
...
 			if match := n.MatchMetadata(m); match {
-				switch len(indexes) {
-				case 0:
-					return map[string]interface{}{string(n.ID): nil}
-				default:
-					kv = make(map[string]interface{})
-					if values, err := getFieldsAsArray(n, indexes); err == nil {
-						kv[indexer.Hash(values...)] = values
-					}
+				kv = make(map[string]interface{})
```
lebauce: Why this change ?

topology/probes/k8s/namespace.go
```diff
@@ -62,43 +56,16 @@ func newNamespaceProbe(clientset *kubernetes.Clientset, g *graph.Graph) Subprobe
...
 	}
 
-	filter := filters.NewAndFilter(
-		filters.NewTermStringFilter("Manager", Manager),
-		filters.NewTermStringFilter("Type", "namespace"),
-		filters.NewNotNullFilter("Name"),
-	)
-
-	namespaceIndexer := graph.NewMetadataIndexer(g, nsSubprobe, graph.NewElementFilter(filter), "Name")
+	namespaceIndexer := graph.NewMetadataIndexer(g, g, graph.Metadata{"Manager": Manager, "Type": "namespace"}, "Name")
```
lebauce: This creates a namespace indexer per resource type. This is very costly

topology/probes/k8s/cache.go
```diff
@@ -41,16 +41,6 @@ import (
 type ResourceHandler interface {
 	Map(obj interface{}) (graph.Identifier, graph.Metadata)
 	Dump(obj interface{}) string
-	IsTopLevel() bool
```
lebauce: I added IsTopLevel at your request. In my previous implementation, the top level resources were passed as a string array. Just like in your proposition :-)

topology/probes/k8s/cluster.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+func newClusterLinker(g *graph.Graph, subprobes map[string]Subprobe, manager, ty string) probe.Probe {
+	dstCache := subprobes[ty]
+	if dstCache == nil {
+		return nil
+	}
+
+	clusterIndexer := graph.NewMetadataIndexer(g, g, graph.Metadata{"Manager": Manager, "Type": "cluster"})
```
lebauce: Creating a indexer for a single resource is probably overkill

topology/graph/indexer.go
hunchback: because when running `indexer.Get()` I want all the nodes meeting the criteria, which was not the case with the previous implementation.

topology/probes/k8s/cache.go
hunchback: @lebauce - you are correct - but when looking at the result I realized that there was no benefit to  using IsTopLevel() (for indicating a link the cluster) over the string array. And this became clearer when having to do the same with indicating l link to namespace(s).

topology/probes/k8s/cluster.go
hunchback: it simplified implementation (over sharing the indexer between all the cluster linkers) and I don't think there is a performance penalty

topology/probes/k8s/namespace.go
hunchback: see above (my comment on the cluster indexer)  - this simplifies implementation and no apparent execution cost.

etc/skydive.yml.default
hunchback: the change remove the cluster element from being created by default (for both k8s and istio) but also enables no to turn-off the cluster element.

topology/probes/k8s/cluster.go
lebauce: Yes, there is. For every change on the graph, one of OnNodeAdded/OnNodeUpdated/OnEdgeAdded/OnEdgeUpdated will be called. The only time this will do something is when we add the cluster node, and we exactly know when this will happen. I'm sorry but it makes no sense to index a single resource

topology/probes/k8s/cache.go
lebauce: It was just a remark. I don't think the `IsTopLevel` is useful.

topology/probes/k8s/namespace.go
lebauce: Did you try this on a platform with hundred or thousand of namespaces ? I really don't see the difference in terms of implementation between creating one indexer will a big filter and creating many indexers and iterating over a loop. Both a almost equal in terms of complexity but one has clearly a bigger execution cost

topology/graph/indexer.go
lebauce: If I'm correct, this is only used for the cluster indexer

hunchback:
@nplanel - we have had some discussions on this in the past but I wanted to know what would be the acceptable minimal tests required. Note that k8s/istio are not yet connected to the "physical" topology and have no flows associated with any nodes.

hunchback:
correct, only used for the cluster indexer

hunchback:
but when i want to add external istio object under namespaces i would want
to add a new indexer, and given this i rather use the same implementation
in both cases.

nplanel:
> @nplanel - we have had some discussions on this in the past but I wanted to know what would be the acceptable minimal tests required. Note that k8s/istio are not yet connected to the "physical" topology and have no flows associated with any nodes.

@hunchback Sorry my remark was a little bit short. 
First it's not negociable k8s functional tests must run with database; Changing the number of indecies fields limit in ES is not possible

Second, a full use case in k8s functional test will be more than welcome : spawn a pod, generate traffic, request topology/flow matching with the network/topo/pods

topology/graph/indexer.go
hunchback: correct

topology/probes/k8s/namespace.go
hunchback: I don't see the difference in performance between the two approaches (as it is we have never tested our system for thousands of namespaces, if the probe breaks it will not be due to the above change ...).

and again the change towards a separate linker per each namespace--><resource> (e.g. namespace-->pod; namespace-->service) was triggered by the need to add namespace--><istio resource> links thus suggesting that each namespace--><resource> linker must work independantly.

topology/probes/k8s/cluster.go
lebauce: The previous implementation did not make use of an  MetadataIndexerLinker for the links between cluster and the top level objects. A `classic` linker that just implements GetBALinks is enough.

topology/probes/k8s/namespace.go
lebauce: Before your change:
- Only one indexer was used to index namespaces
- Only one indexer was used to index the resources that must be linked to namespaces

After your change:
- One indexer is used to index namespace **per resource type** (let's say there is 10 resource types. For every change on the graph, 10 methods will be called, 10 hash will be computed and 10 entries will be created in indexes)
- One indexer is used per resource type (same calculus than above)

I don't understand what you mean by `was triggered by the need to add namespace--> links`. What is namespace -> links ?

etc/skydive.yml.default
hunchback: cluster is not a k8s resource (its' a child of pod), and in istio we may add workspace (a group of pods defined by selector such as v1) which again is not a k8s resource ... so while most of the nodes we are showing are direct k8s resource this is not true in the general case

topology/probes/k8s/cluster.go
hunchback: @lebauce - for have changed this a little such that a use a dedicated clusterEventHandler (so as not to be tirggered on every add/del/update of a node on the graph).

topology/probes/k8s/namespace.go
hunchback: in the previous implementation all the following were handled by a single linker:
- namespace --> pod
- namespace --> networkpolicy
- namespace --> etc.

but as we now also have the ability of istio probe to link its' resources under namespaces I wanted to expose the ability to link individual resources under namespace such as:
- namespace (k8s) --> destinationrule (istio)

this resulting implementation has now much simplified the namespace linker and I still don't see the significant impact on performance (yes adding/deleting/editing namespace resources will now  echo on 10 indexers (one per resource) rather than a single indexer ... but namespaces are small in number compared to the entire k8s landscape (1/100) and events on namespaces are very rare (compared to pods for example ...) so this is not a matter of true performance impact thus I prefer to go with the simpler implementation.

lebauce: If the goal is to be able to create `namespace (k8s) --> destinationrule (istio)` (and I understand this requirement), we can think of an other way.

Now, I really don't like to say it like this but I am not going to merge this PR as we don't seem to be able to agree on it.

hunchback:
> > @nplanel - we have had some discussions on this in the past but I wanted to know what would be the acceptable minimal tests required. Note that k8s/istio are not yet connected to the "physical" topology and have no flows associated with any nodes.
> 
> @hunchback Sorry my remark was a little bit short.
> First it's not negociable k8s functional tests must run with database; Changing the number of indecies fields limit in ES is not possible

@nplanel - I am handling this in a separate PR

> 
> Second, a full use case in k8s functional test will be more than welcome : spawn a pod, generate traffic, request topology/flow matching with the network/topo/pods

@nplanel - I have not yet implemented flows with relation to k8s ... we are still working on the topology representation and the current tests provide  a reasonable level of coverage to k8s topology (which creation of graph.Node and graph.Edge).
-

topology/graph/indexer.go
lebauce: I'm fine with it, it wasn't used anywau

etc/skydive.yml.default
lebauce: That still leaves the question : What is the use case for this ? Who would want to remove the creation of a single resource ?

hunchback: why have cluster handled this way:
- you can now disable to enable cluster object; it may be handy when merging k8s to physical to eliminate the cluster node
- all types (cluster included) are handled uniformly as Subprobes (including having cluster code in cluster.go file)
- when we support multiple clusters this will make even more sense ...

hunchback:
@safchain / @lebauce - with regards to namespace linker (see https://github.com/skydive-project/skydive/pull/1371/commits/13cedaafbfd355c830e047c72c5bc41fb796166b#diff-b5957612729a89bf2e7310786d10df78R61)

In the previous implementation you were using a single linker:
- indexer1 --> listening on namespace events
- indexer2 --> listening on ***all*** resource events (including changes in objects which are not under namespace such as container, storageclass, node, etc.)

In the new implementation I have linker per each resource type:
- indexer1 --> listening on namespace events
- indexer2 --> listening ***only*** on the per resource type events (e.g. only on pod changes)

so comparing the two - if I have 10 types (pod, networkpolicy, deployment, etc.) connected to namespace then I am generating 10x on indexer1 ***but*** while previously indexer2 was listening on all events (k8s, istio, physical - everything) I have now reduced this significantly to only the k8s resources potentially connected to a namespace. so complexity wise the current implementation in better than the previous one

lebauce:
@hunchback Please ping the others about this PR. I'm giving up

safchain:
A general comment after reading it. I'm not sure to get the real purpose of the patch. It seems to be more a refactoring patch not bringing any value to the end user. To be clear while I'm not against refactoring thing in general I prefer (want) to see it as part of something that bring feature/bug fixes. The patch seems to be related to some Istio integration preparation which is great but I would prefer to see it with the Istio integration directly otherwise we all know that you will have to refactor again once you will work on the real integration. That is the game of the software development we never know what will be the final implementation :). For the clustering thing, it is an handy object and I don't see (currently) any value to disable it. In the future if we plan to address (with a patch) multiple clusters or physical links so yes we will handle it differently. To summarize I'll keep the current implementation except if this patch fixes bugs. A side note, we have more and more users using Skydive with Kubernetes/Openshift and I would love to see a patch fixing the history, not having a ton of error messages in the logs and not being able to do history requests.
So a PR must be a Feature or Bugfix

hunchback:
ok will postpone until adding istio objects making use of these changes.
will address the noted issues (history, error messages, etc)

hunchback:
I added the istio resources (only the ones available w/o modifications to kiali) so as to demonstrate how they are making use of the above changes,

As to the two mentioned issues:
- histroy - https://github.com/skydive-project/skydive/pull/1381
- errors - https://github.com/skydive-project/skydive/pull/1403

hunchback:
replaced by: https://github.com/skydive-project/skydive/pull/1404 (github stopped enabling to update the changes on this PR)

hunchback:
replaced by: https://github.com/skydive-project/skydive/pull/1404 (github not updating branch anymore)

hunchback:
replaced by: https://github.com/skydive-project/skydive/pull/1404

PR #1373, opened 2018-10-11
root not required by agent to start

safchain:


lebauce:
run skydive-scale-tests

PR #1376, opened 2018-10-12
docker: fix race in create manifest, use image Id instead of tag

safchain:


lebauce:

PR #1381, opened 2018-10-15
es: don't index topology metadata fields under '*.Extra'

hunchback:
provided two mechanisms for handling large metadata fields (as the k8s metadata section):
- support configuring elasticsearch limit on number of fields (default 1000 fields)
- enable k8s probe to disable or prune size of the k8s metadata section (default is enabled with max 100 fields)

### DIFF
--- a/scripts/ci/jobs/jobs.yml
+++ b/scripts/ci/jobs/jobs.yml
@@ -281,7 +281,7 @@
           slave-name: baremetal
     builders:
       - skydive-test:
-          test: BACKEND=memory scripts/ci/run-istio-tests.sh
+          test: BACKEND=elasticsearch scripts/ci/run-istio-tests.sh
     publishers:
       - junit:
           results: tests.xml
--- a/tests/k8s_test.go
+++ b/tests/k8s_test.go
@@ -65,7 +65,7 @@ func makeHasArgsType(mngr, ty interface{}, args1 ...interface{}) []interface{} {
 }
 
 func makeHasArgsNode(node *graph.Node, args1 ...interface{}) []interface{} {
-	m := node.Metadata()
+	m := node.Metadata
 	args := []interface{}{}
 	for _, key := range []string{"Namespace", "Name"} {
 		if val, ok := m[key]; ok {
@@ -166,7 +166,7 @@ func testNodeCreation(t *testing.T, setupCmds, tearDownCmds []Cmd, mngr, typ, na
 				return err
 			}
 
-			m := obj.Metadata()
+			m := obj.Metadata
 			for _, field := range fields {
 				if _, ok := m[field]; !ok {
 					return fmt.Errorf("Node '%s %s' missing field: %s", typ, name, field)
--- a/tests/k8s_test.go
+++ b/tests/k8s_test.go
@@ -528,7 +528,7 @@ func TestK8sNetworkPolicyAllowIngressNamespaceToNamespaceScenario(t *testing.T) 
 }
 
 func TestK8sNetworkPolicyAllowIngressPodToPodPortsScenario(t *testing.T) {
-	testK8sNetworkPolicyObjectToObjectScenario(t, k8s.PolicyTypeIngress, k8s.PolicyTargetAllow, "ports", "Ports", ":80")
+	testK8sNetworkPolicyObjectToObjectScenario(t, k8s.PolicyTypeIngress, k8s.PolicyTargetAllow, "ports", "PolicyPorts", ":80")
 }
 
 func TestK8sServicePodScenario(t *testing.T) {
--- a/topology/graph/elasticsearch.go
+++ b/topology/graph/elasticsearch.go
@@ -39,11 +39,23 @@ const graphElementMapping = `
 {
 	"dynamic_templates": [
 		{
-			"strings": {
-				"match": "*",
-				"match_mapping_type": "string",
+                        "strings": {
+                                "path_match": "*",
+                                "path_unmatch": "*.Extra.*",
+                                "match_mapping_type": "string",
+                                "mapping": {
+                                        "type": "keyword"
+                                }
+                        }
+                },
+		{
+			"extra": {
+				"path_match": "*.Extra",
 				"mapping": {
-					"type": "keyword"
+					"type": "object",
+					"enabled": false,
+					"store": true,
+					"index": "no"
 				}
 			}
 		},
--- a/topology/probes/k8s/networkpolicy.go
+++ b/topology/probes/k8s/networkpolicy.go
@@ -255,7 +255,7 @@ func (npl *networkPolicyLinker) createLinks(np *v1beta1.NetworkPolicy, npNode, f
 	metadata := npl.newEdgeMetadata(ty, target, point)
 	for _, objNode := range podNodes {
 		if filterNode == nil || filterNode.ID == objNode.ID {
-			metadata.SetFieldAndNormalize("Ports", getFieldPorts(np, ty))
+			metadata.SetFieldAndNormalize("PolicyPorts", getFieldPorts(np, ty))
 			fields := []string{string(npNode.ID), string(objNode.ID)}
 			for k, v := range metadata {
 				fields = append(fields, k, v.(string))


storage/elasticsearch/client.go
```diff
@@ -156,13 +156,25 @@ func (c *Client) addMapping(index Index) error {
 	return nil
 }
 
+func (c *Client) putSetting(index Index) error {
+	const setting string = `{ "index.mapping.total_fields.limit": 10000 }`
```
safchain: I would prefer to know and to be sure about what we are going to store with a limited amount of fields. I think we need to handle this in the probes.

lebauce:
@hunchback There is a reason why the default is 1000 : an index has a real cost. The problem is that we put objects under the "K8s" metadata field that we have no idea what's inside. There are maps and maps of maps. Some of them can even be specified by the user. So we don't even know if 10000 is enough.

The solution we discussed during the hackathon is that we need to select the fields we want to store in the metadata.

storage/elasticsearch/client.go
hunchback: @safchain - alternative fixes could be:

1. arbitrarily limit the number of fields in k8s (as we have many types of resources we are handling its' not always clear which subfield will overflow and cause us to exceed ElasticSearch max field limit.
2. serialize the entire k8s field as a BLOB (for all backends) and then after reading from the backend and before transferring to UI we translate into a multi-level map which can be displayed correctly in the UI.

Let me know what you think about these options

safchain: @hunchback maybe we could do both, limiting the number of field to well known fields so that we can continue to do gremlin filter on them and if more is needed we can use blob. 

I would first limit the fields and then adding the blob thing. Maybe that is something we could extend (thinking about lazy loading, other es index)

hunchback:
run skydive-ppc64le-tests

hunchback:

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

lebauce:
- This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.
- With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.
- The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.
- We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

hunchback:
> * This PR must cherry pick the commits that enable history for the functional tests. Otherwise, we have no idea if this fixes anything.

ok

> * With the `Prune` mechanism, we have no idea of what will be recorded and what will be discarded. It will even be different for the same type of resource as the order of fields in maps is not predictable : some StorageClass will have a field "A", when others won't.

Prune sorts the keys before pruning so its' predictable

> * The `maxFields` only ensure that we won't index more than `maxFields` fields for a single resource. So maxFields could be exhausted after having indexed a few resources. This approach simply does not work. Until we have a way to serialize/unserialize on the fly, the only viable way is to select the field we want to index per resource type.

Please explain: `So maxFields could be exhausted after having indexed a few resources`

> * We are not going to bump the `index.mapping.total_fields.limit` until we know how many fields we are indexing.

This needs to be used in conjunction with the above, but generally i believe in productive settings one would not want to increase total_fields.limit beyond the default (which is 1000). So I am fine with *removing* this commit

lebauce:
@hunchback 
Let's say we have a new StorageClass object that has - let's say - 200 fields.
Then a Job object with also 200 fields. Then a NetworkPolicy. Then a pod. Then a deployment. We reached the 1000 field limit and Skydive will not be able to index any object of an other type.

That's why the only good solution is to index a selected set of fields.

hunchback:
> @hunchback
> Let's say we have a new StorageClass object that has - let's say - 200 fields.
> Then a Job object with also 200 fields. Then a NetworkPolicy. Then a pod. Then a deployment. We reached the 1000 field limit and Skydive will not be able to index any object of an other type.
> 
> That's why the only good solution is to index a selected set of fields.

ok ... in that case I guess this may be a good point to look farther into serialization of objects ... basically store them in elasticsearch as BLOBs and expand for the UI.

lebauce:
@hunchback Then you won't be able to able to do any query on it, which makes the history basically useless. Again, **we need to select the fields we want to index**. Either by selecting the fields we want at the probe level, or by explicitely tell Elasticsearch to index the fields we are interested in.

hunchback:
> @hunchback Then you won't be able to able to do any query on it, which makes the history basically useless. Again, **we need to select the fields we want to index**. Either by selecting the fields we want at the probe level, or by explicitly tell Elasticsearch to index the fields we are interested in.

All the fields we needed for searching have already been placed in the toplevel (we can add any specific fields if we discover they are missing) - so the K8s entry is just to help describe the node ***after*** its' selected. The new container code does conflict with this assumption - this should be fixed - see my comment within the patch - after fixing the container code we can simply disable the K8s field by default (leaving it for now as a debug option).

scripts/ci/jobs/jobs.yml
```diff
@@ -269,7 +269,7 @@
           cancel-builds-on-update: true
     builders:
       - skydive-test:
-          test: BACKEND=memory scripts/ci/run-istio-tests.sh
+          test: BACKEND=elasticsearch scripts/ci/run-k8s-tests.sh
```
lebauce: Why disabling the Istio tests ?

hunchback: mistake - will fix next push

safchain:
@hunchback @lebauce It seems we agreed that we need to keep only indexed fields that are relevant for : 

- making the probe working :)
- allowing the user to do basic k8s filtering

For the rest it should be a non-indexed sub key

So I would create the following structure

```
type K8S struct {
   field1
   field2

   Extra ....(everything else we want to keep, could be everything)
}
```

I would store this in the metadata under the sub K8S, so that this is easy for users to find K8s things.
I would add a rule in the ES schema to not index K8S.Extra

hunchback:
> @hunchback @lebauce It seems we agreed that we need to keep only indexed fields that are relevant for :
> 
> * making the probe working :)
> * allowing the user to do basic k8s filtering
> 
> For the rest it should be a non-indexed sub key
> 
> So I would create the following structure
> 
> ```
> type K8S struct {
>    field1
>    field2
> 
>    Extra ....(everything else we want to keep, could be everything)
> }
> ```
> I would store this in the metadata under the sub K8S, so that this is easy for users to find K8s things.
> I would add a rule in the ES schema to not index K8S.Extra

So I tried just that but I am not managing to setup the index such that the K8s field is really ignored, I was using the following commands attempting to do that:

```
echo
echo
echo "***** create index *****"
curl -X DELETE http://localhost:9200/*
curl -X PUT http://localhost:9200/skydive_test -d '{
  "mappings": {
    "graph_element": {
        "dynamic_templates": [
                {
                        "strings": {
                                "match": "*",
                                "match_mapping_type": "string",
                                "mapping": {
                                        "type": "keyword"
                                }
                        }
                },
                {
                        "k8s": {
                                "path_match": "K8s.*",
                                "mapping": {
                                        "type": "object",
                                        "enabled": false,
                                        "store": true,
                                        "index": "no"
                                }
                        }
                }
        ]
    }
  }
}'

curl -X PUT http://localhost:9200/skydive_test/_settings -d '{
        "index.mapping.total_fields.limit": 15
}'
echo
echo
echo "***** insert 2 fields *****"
curl -X PUT http://localhost:9200/skydive_test/graph_element/1 -d '{
  "k8s" : { "0": 2, "1": 2 }
}'

echo
echo
echo "***** insert 5 fields *****"
curl -X PUT http://localhost:9200/skydive_test/graph_element/1 -d '{
  "k8s" : { "0": 2, "1": 2, "2": 2, "3": 2, "4": 2 }
}'
```

The result I got:

```
[vagrant@dev skydive]$ source runme


***** create index *****
{"acknowledged":true}{"acknowledged":true,"shards_acknowledged":true,"index":"skydive_test"}{"acknowledged":true}

***** insert 2 fields *****
{"_index":"skydive_test","_type":"graph_element","_id":"1","_version":1,"result":"created","_shards":{"total":2,"successful":1,"failed":0},"created":true}

***** insert 5 fields *****
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Limit of total fields [15] in index [skydive_test] has been exceeded"}],"type":"illegal_argument_exception","reason":"Limit of total fields [15] in index [skydive_test] has been exceeded"},"status":400}[vagrant@dev skydive]$ vim runme
```

hunchback:
@safchain / @lebauce - if you can help disabling the K8s then fine - else we can reconsider pruning the K8s (this may anyway be needed for handling other potential limitations such as network bandwidth etc.)

safchain:
@hunchback As said let's start defining a K8S struct will see the indexing/blob later

hunchback:
> @hunchback As said let's start defining a K8S struct will see the indexing/blob later

For now I have just disabled storing/indexing K8s: https://github.com/skydive-project/skydive/pull/1426

As for re-org of K8s elements:
- important fields directly under <toplevel> (stuff like "Namespace", "Name", "Manager", and other fields)
- extra fields under <toplevel>.K8s

Are you proposing we store in:
- <toplevel>
- <toplevel>.K8s
- <toplevel>K8s.Extra
If so then what goes into <toplevel> vs. <toplevel>.K8s

lebauce:
`Name` and `Manager` are toplevel (as they are common to other resources)
`Namespace` would go under `K8s`
Extra fields would go under `K8s.Extra`

hunchback:
> `Name` and `Manager` are toplevel (as they are common to other resources)
> `Namespace` would go under `K8s`
> Extra fields would go under `K8s.Extra`

ok - fine - with regards to `Name` it should be the "fullname" of the k8s resource:
- for a node: `node1`
- for a pod: `namespace1/pod1`
- for a container: `namespace1/pod1/container1`

under K8s we would still hold the short name:
```
Name: "namespace1/pod1/container1",
K8s: {
  Namespace: "namespace1",
  Pod: "pod1",
  Name: "container1",
  ... and more important k8s fields ....
}
```

hunchback:
another point is that rather than:
- *toplevel*.K8s.Extra
I rather name extra:
- *toplevel*.Extra.K8s

this why our special handling of Extra data (in disabling indexing in elasticsearch; in presenting it in a separate panel in the UI; etc.) is generic (not k8s specific) such that other Probes can make use of this w/o requiring special handling

hunchback:
The restructuring of K8s is orthogonal to this PR (which focuses on supporting large K8s object in elasticsearch) this was handled in a separate PR: https://github.com/skydive-project/skydive/pull/1432

hunchback:
@lebauce / @safchain - I was not able to get elasticsearch to not index the K8s field - can you have a look at the script I used above ... thanks!

lebauce:
```
curl -X PUT http://localhost:9200/skydive_test -d '{
  "mappings": {
    "graph_element": {
        "dynamic_templates": [
                {
                        "strings": {
                                "path_match": "*",
                                "path_unmatch": ".Extra.*",
                                "match_mapping_type": "string",
                                "mapping": {
                                        "type": "keyword"
                                }
                        }
                },
                {
                        "extra": {
                                "path_match": "*.Extra",
                                "mapping": {
                                        "type": "object",
                                        "enabled": false,
                                        "store": true,
                                        "index": "no"
                                }
                        }
                }
        ]
    }
  }
}'
```

hunchback:
> ```
> curl -X PUT http://localhost:9200/skydive_test -d '{
>   "mappings": {
>     "graph_element": {
>         "dynamic_templates": [
>                 {
>                         "strings": {
>                                 "path_match": "*",
>                                 "path_unmatch": ".Extra.*",
>                                 "match_mapping_type": "string",
>                                 "mapping": {
>                                         "type": "keyword"
>                                 }
>                         }
>                 },
>                 {
>                         "extra": {
>                                 "path_match": "*.Extra",
>                                 "mapping": {
>                                         "type": "object",
>                                         "enabled": false,
>                                         "store": true,
>                                         "index": "no"
>                                 }
>                         }
>                 }
>         ]
>     }
>   }
> }'
> ```

Great - this works! I have updated the PR - but depends on: https://github.com/skydive-project/skydive/pull/1432

safchain:
need to be rebased

lebauce:
run skydive-k8s-tests

safchain:
Need to be rebased and fix the code with the new modification of the graph

lebauce:
@hunchback
```
tests/k8s_test.go:68:20: cannot call non-function node.graphElement.Metadata (type graph.Metadata)
tests/k8s_test.go:169:21: cannot call non-function obj.graphElement.Metadata (type graph.Metadata)
```

hunchback:
run skydive-k8s-tests

lebauce:

hunchback:
run skydive-k8s-tests

lebauce:
@hunchback
```
&{mapper_parsing_exception object mapping for [Metadata.Ports] tried to parse field [Ports] as object, but found a concrete value     false map[] [] []}
```

This is probably due to the fact that service.go does
```
m.SetFieldAndNormalize("Ports", srv.Spec.Ports)
```

and networkpolicy.go does:
```
metadata.SetFieldAndNormalize("Ports", getFieldPorts(np, ty))
```

So `Metadata.Ports` are different types, making Elasticsearch unable to index them

hunchback:
> @hunchback
> 
> ```
> &{mapper_parsing_exception object mapping for [Metadata.Ports] tried to parse field [Ports] as object, but found a concrete value     false map[] [] []}
> ```
> This is probably due to the fact that service.go does
> 
> ```
> m.SetFieldAndNormalize("Ports", srv.Spec.Ports)
> ```
> and networkpolicy.go does:
> 
> ```
> metadata.SetFieldAndNormalize("Ports", getFieldPorts(np, ty))
> ```
> So `Metadata.Ports` are different types, making Elasticsearch unable to index them

I see - so I should probably have a clear separation between generic K8s fields such as:
- K8s.Namespace
- K8s.Name
- K8s.Labels (not current included)

Vs. type specific fields
- K8s.Service.Ports
- K8s.NetworkPolicy.Ports

I will merge this PR into https://github.com/skydive-project/skydive/pull/1432 and then handle.

lebauce:
@hunchback I would prefer that you cherry pick https://github.com/skydive-project/skydive/pull/1432/commits/ccf6675062838b9e5d8fa56f85dda34a889e5b6e in this PR

lebauce:
run skydive-k8s-tests

lebauce:
@hunchback I did not see that this PR could not run the tests with the history enabled as it only modifies the mapping without moving things under `K8s.Extra`

PR #1389, opened 2018-10-17
k8s: add type based filters

hunchback:
to test run skydive with k8s probe enabled and verify that you get a list of filters (such as filter services, network, compute, storage, etc.)


hunchback:

PR #1398, opened 2018-10-17
gremlin: add HasEither step to generate a OR filter instead of AND

safchain:


safchain:

PR #1403, opened 2018-10-21
k8s: throttle internal k8s errors

hunchback:
all internal k8s errors are now redirected to `logging.getLogger().Debug()`

### DIFF
--- a/topology/probes/k8s/probe.go
+++ b/topology/probes/k8s/probe.go
@@ -23,8 +23,14 @@
 package k8s
 
 import (
+	"sync"
+	"time"
+
+	"github.com/skydive-project/skydive/logging"
 	"github.com/skydive-project/skydive/probe"
 	"github.com/skydive-project/skydive/topology/graph"
+
+	"k8s.io/apimachinery/pkg/util/runtime"
 )
 
 func int32ValueOrDefault(value *int32, defaultValue int32) int32 {
@@ -94,4 +100,43 @@ func NewProbe(g *graph.Graph, manager string, subprobes map[string]Subprobe, lin
 		subprobes: subprobes,
 		linkers:   linkers,
 	}
+}
+
+func logOnError(err error) {
+	logging.GetLogger().Warning(err)
+}
+
+type errorThrottle struct {
+	period   time.Duration
+	lastLock sync.RWMutex
+	last     time.Time
+}
+
+func (r *errorThrottle) onError(error) {
+	r.lastLock.RLock()
+	d := time.Since(r.last)
+	r.lastLock.RUnlock()
+
+	if d < r.period {
+		time.Sleep(r.period - d)
+	}
+
+	r.lastLock.Lock()
+	r.last = time.Now()
+	r.lastLock.Unlock()
+}
+
+func muteInternalErrors() {
+	throttle := errorThrottle{
+		period: time.Second,
+		last:   time.Now(),
+	}
+	runtime.ErrorHandlers = []func(error){
+		logOnError,
+		throttle.onError,
+	}
+}
+
+func init() {
+	muteInternalErrors()
 }


hunchback:
run skydive-k8s-tests

hunchback:
@lebauce - I was not able to trigger k8s tests on this PR

topology/probes/k8s/probe.go
```diff
@@ -74,8 +77,15 @@ func (p *Probe) Stop() {
...
 
+func muteInternalErrors() {
+	runtime.ErrorHandlers = []func(error){
+		func(err error) { logging.GetLogger().Debug(err) },
+	}
+}
+
 // NewProbe creates the probe for tracking k8s events
 func NewProbe(manager, clusterName string, g *graph.Graph, subprobes map[string]Subprobe, linkers []probe.Probe) (*Probe, error) {
+	muteInternalErrors()
```
nplanel: If we got error (panic and others) from k8s we don't want to close our eyes (in production) and say it's all is fine

Please think in the right way, if we got error(s) from k8s, where the error come from ? why we should ignore it (ex: only a kind of specific error) ?, are we sure using k8s API correctly, any bug in k8s module ? ....

Why k8s errors happens / bother you ? could you reproduce it in the CI ? why it's not reproduced in the CI ? ...

hunchback: We are getting the below error (and other errors as well) on some of our internal k8s landscapes (on both private and public infrastructure); the source of the issue being various permission settings per target k8s cluster ... something only known during runtime.

One less drastic solution may be to introduce a rate limit mechanism:
- simple - limit to one message per given time windows (such as one message every 5 minutes)
- complex - discard any duplicated error-message if identical was printed within the time window

```
ERROR: logging before flag.Parse: E1008 08:53:28.179347       1 reflector.go:205] github.com/skydive-project/skydive/topology/probes/k8s/cache.go:148: Failed to list *v1.PersistentVolume: persistentvolumes is forbidden: User "system:serviceaccount:monitoring:default" cannot list persistentvolumes at the cluster scope
```

Coming from here: https://gitlab.cncf.ci/kubernetes/kubernetes/blob/d56b9a69fa84248f5c9b27ee02976ad86b36671d/staging/src/k8s.io/client-go/tools/cache/reflector.go#L205

nplanel: So the message is 

> Failed to list *v1.PersistentVolume: persistentvolumes is forbidden: User "system:serviceaccount:monitoring:default" cannot list persistentvolumes at the cluster scope

Meaning your k8s user "system:serviceaccount:monitoring:default" used by Skydive k8s probe cannot list the ressource of **persistentvolumes** due to **access is forbidden**

So in other words the k8s probe try to list ressources that we don't have right to access to

1/ Is it your requirement this user can't list persistentvolumes ressource with this user ?
   1.a/ why enable a probe that can't be listed
   1.b/ are you using the right user (I supposed yes)

2/ I the Skydive k8s probe can't access/list persistentvolumes due to RBAC strategy, don't you think that the role of the Skdyive k8s probe to return the error to the user only once and stop to inquery k8s ?

eranra: Hi @nplanel from my point of view:

1. This message and all other log messages (for example I saw many repeating rbac messages)  should be throttled to some level ... we are filling the log with messages and we need a way to limit that noise. This is correct to every message not specific to this case.

2. From Skydive point of view, this is not an error, it should be reported as info (we can argue if maybe warning) but for sure this is not an error.

nplanel: 1/ throttled is not a solution, see point 2
2/ yes it is, from Skydive k8s probe point of view
  a/ Why Skydive k8s probe query k8s ; As the probe don't have permission granted ?
  b/ Why the probe continue to do so if didn't have permission ? (reported once will create only one error message)

eranra: - Your point 2 ... agree ... I was referring to different rbac, the code that @lebauce added is creating a lot of repeated log entries . So I totally agree we need to report only once (or once in a while, but not very frequent)

```
2018-10-24 19:00:16.622367 I | [Request: admin, capture, read ---> true]
2018-10-24 19:00:24.451974 I | [Request: admin, topology, read ---> true]
2018-10-24 19:00:24.781350 I | [Request: admin, topology, read ---> true]
2018-10-24 19:00:25.085122 I | [Request: admin, topology, read ---> true]
2018-10-24 19:00:25.565114 I | [Request: admin, topology, read ---> true]
2018-10-24 19:00:27.870318 I | [Request: admin, topology, read ---> true]
2018-10-24 19:00:31.006806 I | [Request: admin, topology, read ---> true]
2018-10-24 19:00:31.662669 I | [Request: admin, capture, write ---> true]
2018-10-24 19:00:42.474779 I | [Request: admin, topology, read ---> true]
2018-10-24 19:00:45.944516 I | [Request: admin, capture, write ---> true]
2018-10-24 19:00:45.964985 I | [Request: admin, capture, read ---> true]
```

eranra: I think we need a way to limit all those repeated messages ... not just in k8s but inn general

eranra: With regards to the level of log ... Error indicates something that need to be fixed, in cases that some of the info is available from  K8S and some is not (e.g. due to permissions/rbac) ... what we do in the prob is just trying to get access, the fact that we can't and we get error from k8s is not an error ... this is the behavior of the code == as expected ... we should not error on that. The prob code is using this to know that he can't access that resource ... I think info, or at max warning bu not an error. Maybe if the code can get that info with some other API ... list what he can get access to.

eranra: Your last point ... the prob must continue to check (I agree maybe reduce the granularity) because permissions might change ... they are dynamic ... so I do expect the code to continue (once in a while) and access ... if we get notification on change in permissions, maybe we can limit the call just when we see such event.

nplanel: 1/ Would be better to not having the conversation forking, as 
> 2018-10-24 19:00:25.565114 I | [Request: admin, topology, read ---> true]
these messages are from github.com/casbin/casbin used by Skydive RBAC
This PR talking about k8s messages (may Skydive RBAC messages are too verbose or incomplete)

2/ K8s having all these errors generated here : k8s.io/apimachinery/pkg/api/errors/errors.go
That the role of the probe to manage/report "permission errors" messages ... and not ignore it or continue to retry ... May the probe need to check permission at the Start()

3/ If the k8s RBAC changing (dynamically) would the client would received an API event ?
**Did you test** that you didn't need to re-auth (re-connect) to have the latest RBAC applied ? (I mean RBAC can be applied per connection/context not per request on server side)

Please let's start by doing : 
 A/ a simple implementation that didn't retry when got an error on a k8s ressource
 B/ check point 3
 C/ implementation depend on point 3 (from API event **or** timer, => transparent **or** re-connection)

lebauce: @nplanel @eranra Regarding the RBAC logs, it indeed pollutes the logs. I will post a PR to change this

lebauce:
@nplanel @eranra https://github.com/skydive-project/skydive/pull/1420

topology/probes/k8s/probe.go
lebauce: Regarding the `Failed to list *v1.PersistentVolume: persistentvolumes is forbidden: User "system:serviceaccount:monitoring:default" cannot list persistentvolumes at the cluster scope` message. This is a configuration issue on the Kubernetes side so it's true that it is not a Skydive error. A warning may be enough.

This is dynamic and can be fixed at runtime. If the RBAC is changed, we may receive an API event but it's not easy to interpret. I can confirm that fixing the RBAC does not it a reconnect.

hunchback:
I read through the comments above some points:

1. these errors are coming from the k8s server as such, configuration changes to he server may cause issues to come or go ***dynamically*** throughout the lifetime of the k8s cluster. (our probe will see these changes).

2. the errors at k8s server should be downgraded for now I have set them to ***warning*** but I would not disagree to set to ***info***

3. the errors repeat and high rate (originally as high as 1 error per minisecond)  this would flood skydive log ... so I have changed the default to 1 minute and have made it configurable via `analyzer.topology.k8s.log_period_seconds` configuration parameter.

lebauce:
I agree that warning is the right level for this message. The suggested solution seems ok to me. But I would prefer not to define an other configuration value for this. Let's use a hardcoded value of 1 minute for this and not make the configuration even more complex for such a corner case.

etc/skydive.yml.default
```diff
@@ -118,6 +118,9 @@ analyzer:
       # specify the path of k8s configuration YAML file.
       # config_file: /etc/skydive/kubeconfig
 
+      # specify the minimal period between log messages
+      # log_period_seconds: 60
```
safchain: maybe 5sec to start, 1min seems to be a bit long

nplanel:
@hunchback could you sync with @eranra 
We discuss about this PR during the last weekly call
Please read all _oudated_ comments

hunchback:
is there any specific comment i should take note of?

lebauce:
@hunchback Why putting back the configuration value when using 5 seconds ?
The permission error should not happen in a normal environment. This means that something needs to be fixed : either give the user the rights to list the resource or disable the subprobe if the corresponding resource is not supported on the environment.

hunchback:
because this would not work for our internal implementation (which is
representative of other productive environments)  it would still be too
noisy

lebauce:
@hunchback Why do you have such an error on an environment that you control ? You can either add the permission or disable the subprobe that is at fault.

hunchback:
ok will remove config option ...

hunchback:
@lebauce / @safchain - can this be merged ?

topology/probes/k8s/probe.go
```diff
@@ -95,3 +101,38 @@ func NewProbe(g *graph.Graph, manager string, subprobes map[string]Subprobe, lin
...
+}
+
+type errorThrottle struct {
+	period   time.Duration
+	lastLock sync.Mutex
+	last     time.Time
+}
+
+func (r *errorThrottle) onError(error) {
+	r.lastLock.Lock()
```
lebauce: The lock is global and is held during the sleep (5 seconds, which is a lot). We should at least use an atomic.Value

safchain:
@hunchback I'm not sure add a sleep is the right approach since it blocks the execution. What kind of issue are so often reported that you can't fix ? can you elaborate on this ? If we find that it is really needed I would prefer to have a throttling mechanism per message type and not blocking.

hunchback:
the sleep is based on the original implementation of the k8s logging so
this blocks an internal dedicated loop which has no effect on other parts.

one example is permission not set correctly but which can be fixed by user
on his private cluster (ICP).

another example is an overloaded k8s cluster dropping requests ...

topology/probes/k8s/probe.go
```diff
@@ -95,3 +101,38 @@ func NewProbe(g *graph.Graph, manager string, subprobes map[string]Subprobe, lin
...
+	defer r.lastLock.Unlock()
+	d := time.Since(r.last)
+	if d < r.period {
+		time.Sleep(r.period - d)
```
nplanel: time.Sleep() must not be called under lock
even if you got the model from k8s, doesn't mean it's good one

topology/probes/k8s/probe.go
```diff
@@ -95,3 +101,38 @@ func NewProbe(g *graph.Graph, manager string, subprobes map[string]Subprobe, lin
...
+	}
+	r.last = time.Now()
+}
+
+func muteInternalErrors() {
+	throttle := errorThrottle{
+		period: time.Duration(5) * time.Second,
```
nplanel: time.Duration(100) * time.Millisecond
would be enough here as it will generate only 10 lines/messages per second

topology/probes/k8s/probe.go
nplanel: To throttling the log output as the k8s client is event based only and you can't be hooked / do the right behavior due to k8s client design (need to be patched)

Please implement the throttling this way

```
func (r *errorThrottle) onError(error) {
	r.lastLock.RLock()
	d := time.Since(r.last)
        r.lastLock.RUnlock()

	if d < r.period {
		time.Sleep(r.period - d)
	}

	r.lastLock.Lock()
	r.last = time.Now()
	r.lastLock.Unlock()
}
```

hunchback: ok - fixed

topology/probes/k8s/probe.go
```diff
@@ -95,3 +101,42 @@ func NewProbe(g *graph.Graph, manager string, subprobes map[string]Subprobe, lin
...
+	}
+
+	r.lastLock.Lock()
+	r.last = time.Now()
+	r.lastLock.Unlock()
+}
+
+func muteInternalErrors() {
+	throttle := errorThrottle{
+		period: time.Second,
```
nplanel: time.Duration(100) * time.Millisecond
would be enough here as it will generate only 10 lines/messages per second

nplanel:
run skydive-k8s-tests

PR #1405, opened 2018-10-22
K8s istio merge2

hunchback:
replacing: https://github.com/skydive-project/skydive/pull/1371


hunchback:
run tests

lebauce:
@hunchback I suggest we keep the original PR instead of creating a new one. Everything was only due to temporary Github issue

PR #1407, opened 2018-10-22
K8s istio merge

hunchback:
replacing PR: https://github.com/skydive-project/skydive/pull/1371


hunchback:
@lebauce - I was not able to revive the original PR and was forced to open this

hunchback:
run tests

lebauce:
@hunchback You could not reopen the previous as you created this one from the same branch and with the same name. You should have closed this one before reopening the previous one

hunchback:
> @hunchback You could not reopen the previous as you created this one from the same branch and with the same name. You should have closed this one before reopening the previous one

Tried that (and deleted the associated branch) but I still did not get back the REOPEN button

hunchback:

hunchback:
ready for review ...

topology/probes/k8s/probe.go
```diff
@@ -74,37 +72,30 @@ func (p *Probe) Stop() {
...
-	clusterNode *graph.Node
 }
 
-func (cl *clusterLinker) GetBALinks(node *graph.Node) []*graph.Edge {
-	id := graph.GenID(string(cl.clusterNode.ID), string(node.ID), "RelationType", topology.OwnershipLink)
-	return []*graph.Edge{cl.graph.CreateEdge(id, cl.clusterNode, node, topology.OwnershipMetadata(), time.Now(), "")}
+// AppendNamespaceLinkers appends newly created namespace linker per type
+func (p *Probe) AppendNamespaceLinkers(types ...string) {
+	for _, ty := range types {
+		if namespaceLinker := newNamespaceLinker(p.graph, p.subprobes, p.manager, ty); namespaceLinker != nil {
```
lebauce: I know you don't care about my remarks but why don't you instanciate the namespace indexer and pass it to `newNamespaceLinker` ?

hunchback: done: https://github.com/skydive-project/skydive/pull/1407/commits/556d3f09ef6510b6a5e7c2d9ae7af488fa6886d4

topology/probes/k8s/probe.go
```diff
@@ -74,37 +72,32 @@ func (p *Probe) Stop() {
...
 }
 
-func (cl *clusterLinker) GetBALinks(node *graph.Node) []*graph.Edge {
-	id := graph.GenID(string(cl.clusterNode.ID), string(node.ID), "RelationType", topology.OwnershipLink)
-	return []*graph.Edge{cl.graph.CreateEdge(id, cl.clusterNode, node, topology.OwnershipMetadata(), time.Now(), "")}
+// AppendNamespaceLinkers appends newly created namespace linker per type
+func (p *Probe) AppendNamespaceLinkers(types ...string) {
+	namespaceIndexer := newNamespaceIndexer(p.graph)
+	for _, ty := range types {
+		if namespaceLinker := newNamespaceLinker(p.graph, namespaceIndexer, p.subprobes, p.manager, ty); namespaceLinker != nil {
```
safchain: I would rather create a filter as it was previously based on the array of the types using it to create the single indexer. So that is not statically defined and we keep only one indexer.

hunchback:
ok, will handle.

Note that after this change the single indexer on the destination side
needs to listen to all graph events (all k8s, all physical) while in my
implementation i am able to only listen on the destination side to the
relevant k8s events (pod, netpol but not container or storafeclass).

hunchback:
when using the composite object indexer (which needs to catch many types
such as: pod or networkpolicy or deployment or .. ) i guess you need to
listen on the entire graph - right?

hunchback:
i don't understand ..

hunchback:
so are you recommending that we also construct a composite listener?
(beyond the composite type indexer) ...

and i still dont understand why this is superior to last weeks
implementation with the 3 first commits ...

lebauce:
@hunchback No big deal. Let's do as you said (`the single indexer on the destination side needs to listen to all graph events`). We can push a PR later to improve it

lebauce:
@hunchback I think we already discussed about the number of indexers enough :-)

hunchback:
tx

hunchback:
:-))

topology/probes/k8s/cache.go
```diff
@@ -157,11 +147,16 @@ func (c *ResourceCache) Stop() {
 
 // NewResourceCache returns a new cache using the associed Kubernetes
 // client and with the handler for the resource that this cache manages.
-func NewResourceCache(restClient rest.Interface, objType runtime.Object, resources string, g *graph.Graph, handler ResourceHandler) *ResourceCache {
+func NewResourceCache(restClient rest.Interface, objType runtime.Object, resources string, g *graph.Graph, handler ResourceHandler, optionalEventHandler ...*graph.EventHandler) *ResourceCache {
 	watchlist := cache.NewListWatchFromClient(restClient, resources, api.NamespaceAll, fields.Everything())
 
+	eventHandler := graph.NewEventHandler(100)
```
lebauce: If no event handler, EventHandler will be null, making NotifyEvents call segfault

hunchback: How can EventHandler be initialized to nil ?

```
        eventHandler := graph.NewEventHandler(100)
        if len(optionalEventHandler) == 1 {
                eventHandler = optionalEventHandler[0]
        }

        c := &ResourceCache{
                EventHandler:   eventHandler,
                graph:          g,
                handler:        handler,
                stopController: make(chan struct{}),
        }
```

lebauce: @hunchback My bad, it can't be nil. Still, I don't see a reason to specify `optionalEventHandlers`

hunchback: this was introduced so as to let both k8s and istio share the same eventHandler; while previously k8s probe would just pass the list of Subprobes and one could get access to the needed subprobe (which is the event listener); but in the case that we want to setup a linker between and istio resource (such as destinationrule) and a k8s resource (such as namespace) this is no longer possible ... so the solution was to use a global eventHandler (which would not be tied to the Subprobe). 

Please see k8s/namespace.go for how this is used.

hunchback:
run skydive-k8s-tests

lebauce:
@hunchback We still don't see any use case for disabling cluster...

PR #1409, opened 2018-10-22
test : check if gopacket didn't panic

nplanel:


flow/flow_test.go
```diff
@@ -321,6 +321,17 @@ func compareFlow(expected, tested *Flow) bool {
 	return true
 }
 
+func checkGoPacketSanity(t *testing.T, p gopacket.Packet) {
+	if errLayer := p.ErrorLayer(); errLayer != nil {
+		if l := p.Layer(gopacket.LayerTypeDecodeFailure); l != nil {
+			df := l.(*gopacket.DecodeFailure)
+			if len(df.Dump()) > 0 {
```
safchain: if df.Dump() doesn't return anything we won't get the error.

nplanel: yes that the purpose, as DecodeFailure layer can be used to return an "legit" error like : "packet too small" but didn't crash/panic
df.Dump() return something only if a panic occured during a protocol DecodingLayer()

nplanel:
run  skydive-scale-tests

nplanel:
run  skydive-scale-tests

nplanel:
run scale-tests

flow/flow_test.go
safchain: ok thanks for the explanation

Issue #1410, opened 2018-10-23
Support libvirt


pierrecregut:
When we have agents on a host and a guest (eg. some deployments of control nodes with director, or on an agent in a VNF), we would like to see the links between the host and guest interfaces. Here is an example with two levels of nesting (an host hosting a VM deployed manually with devstack hosting a VM). The interfaces added are circled in red.
![skydive-stack](https://user-images.githubusercontent.com/7011215/47349148-403c8c00-d6b3-11e8-94d2-50dc2feacdbc.png)

safchain:
This should be already supported but it seems to be broken, I'll have a look to see what is not working

pierrecregut:
How do you get what to link ? I have some python code that makes the glue but it uses the mac address extracted from libvirt on the host side: https://gist.github.com/pierrecregut/fceb65572c0a1d4b0b3982dc683b54c1

safchain:
Ok got the issue :) so its uses this "probe" https://github.com/skydive-project/skydive/blob/master/topology/probes/peering/peering.go#L47

but this field is only filled by the neutron probe. I think we could fill it in ovsdb probe

pierrecregut:
It seems that something is going wrong. It does not work and one of the cause is that retrieveAttributes fails in the neutron updater. When it is launched, the metadata of the port node is not extended with ExtID information. The mac used is then the wrong MAC and gophercloud finds no port matching the host mac. ExtID is filled later but there is no second chance. The root cause is why actions are done in the wrong order or why is the nodeUpdater not launched with the extended information.

pierrecregut:
The root cause is that the enhancement of the port is obsolete for versions of neutron ovs using connection tracking instead of the intermediate bridge with qvo qvb veths. The enhancement is only done on the qvo node propagating it to the tap node. Today there is only one tap node and PeerIntfMAC should be filled at that point.

There is still another minor bug: I do not see the point of adding MAC to portMetadata if we do not have the attached-mac. This will not work and triggers an error message (but when the port is enhanced the right action will be taken).

Having the libvirt route may be safer and more independent from mechanism drivers as we are relying in fact on nova.

safchain:
libvirt probe makes definitely sense. You're right this part of the code should be adapted. Do you plan to provide a fix or do you want us to fix it ?

pierrecregut:
[...]

PR #1416, opened 2018-10-24
neutron: peering with ovs firewall

pierrecregut:
Adds meta-information for peering in the case where the firwall
is provided directly by openvswitch with conntrack and not by the
hybrid iptable firewall.

Both the legacy and the new implementations are supported.

This is the bugfix part of #1410.


safchain:
run devstack-tests

safchain:

PR #1419, opened 2018-10-24
enable k8s probe for OpenShift

duritong:
+ add info about RBAC

### DIFF
--- a/contrib/openshift/README.md
+++ b/contrib/openshift/README.md
@@ -14,6 +14,12 @@ Install the template
 oc create -f skydive-template.yaml
 ```
 
+Apply the rbac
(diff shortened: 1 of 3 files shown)


lebauce:
@duritong Thank you for this. As the Kubernetes probe is still experimental, I would prefer no to enable by default. Could you please update the contrib/openshift/README.md to give the instructions on how to enable it ? (using the ConfigMap stuff that we recently merged)

lebauce:
There is now instructions regarding RBAC in the README.md and add parameters to enable the k8s probe. So I think we can close this issue. Feel free to open it again if I'm wrong.

PR #1422, opened 2018-10-25
add a flow subscriber writing to object store

orozery:


config/config.go
```diff
@@ -156,18 +156,29 @@ func init() {
...
+	cfg.SetDefault("storage.orientdb.driver", "orientdb")                                          // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.orientdb.addr", "http://localhost:2480")                               // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.orientdb.database", "Skydive")                                         // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.orientdb.username", "root")                                            // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.orientdb.password", "root")                                            // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.driver", "objectstorage")                                // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.endpoint", "http://s3.amazonaws.com")                    // defined for backward compatibility and to set defaults
```
nplanel: from this line you can remove the comment 

> // defined for backward compatibility and to set defaults

flow/storage/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// SearchMetrics searches flow metrics matching filters in the database
+func (c *ObjectStorage) SearchMetrics(fsq filters.SearchQuery, metricFilter *filters.Filter) (map[string][]common.Metric, error) {
+	return nil, nil
+}
+
+// SearchFlows search flow matching filters in the database
+func (c *ObjectStorage) SearchFlows(fsq filters.SearchQuery) (*flow.FlowSet, error) {
+	return nil, nil
```
nplanel: please implement these Search* functions as there are mandatory for Skydive storage backend
You can tests it locally via this command (like our CI) by running : 

`make test.functionals ARGS="-graph.output ascii -standalone -analyzer.topology.backend objectstorage -analyzer.flow.backend objectstorage" VERBOSE=true`

etc/skydive.yml.default
```diff
@@ -65,7 +65,7 @@ analyzer:
 
   # Flow storage engine
   flow:
-    # Storage backend name: myelasticsearch, myorientdb
+    # Storage backend name: myelasticsearch, myorientdb, myobjectstorage
```
nplanel: Thanks for your contribution
Could you describe your main motivation of using aws storage instead of elasticsearch (for example) ?

flow/storage/objectstorage/objectstorage_test.go
```diff
@@ -0,0 +1,123 @@
...
+		flows[i] = fl
+	}
+
+	return flows
+}
+
+func newTestObjectStorage() (*ObjectStorage, *fake.Client) {
+	fakeClient := &fake.Client{}
+	ds := &ObjectStorage{
+		maxStreamDuration:          time.Second * time.Duration(86400),
```
nplanel: time.Duration(24*time.Hour)

flow/storage/objectstorage/objectstorage_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"runtime"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/skydive-project/skydive/flow"
+	"github.com/skydive-project/skydive/storage/objectstorage/fake"
+)
+
+func generateFlowArray(count int) []*flow.Flow {
```
nplanel: please use real flow from pcap, like :
` 	flows := flowsFromPCAP(t, "pcaptraces/simple-tcpv4.pcap", layers.LinkTypeEthernet, nil)
`

storage/objectstorage/fake/fake_client.go
```diff
@@ -0,0 +1,34 @@
+package fake
```
nplanel: for visibility please implement this fake client, directly in your _test.go file
Don't need to have a specific file/package used only by your test

config/config.go
hunchback: i would suggest the backend is renamed to s3 rather than objectstorage (as there are several objectstorage standards such s3, swift)

flow/storage/objectstorage/objectstorage.go
```diff
@@ -92,17 +93,78 @@ func (c *ObjectStorage) StoreFlows(flows []*flow.Flow) error {
...
+		}
+	}
+
+	return flowset, nil
 }
 
 // SearchFlows search flow matching filters in the database
 func (c *ObjectStorage) SearchFlows(fsq filters.SearchQuery) (*flow.FlowSet, error) {
-	return nil, nil
+	flowset, err := c.getAllFlows()
```
nplanel: this will never works on real life / usage ... you can't read all the database, put in memory and filter out
Think you having weeks, months, years of data
You need to query your storage at minima based on Context() window

orozery:
you are right.
searchflows should be used only for testing. not for actual querying.
object storage does not support indexing.

hunchback:
@nplanel / @orozery - basically objectstore is a write only backend - the only reason for supporting `SearchFlows()` is to enable running full functional tests on this backend. (if you can't read the content you can't verify write was really successful). For now I would not invent is protecting the use against incorrectly using objectstore as a real backend .. going forward we can consider introducing the notion of a secondary write only backend (which would work in tandem with a true searchable backend) ... but in the current scope this would be an overkill.

hunchback:
@orozery - can you kindly respond to each comment - if its' been done or will be done then respond accordingly, If you disagree with the comment then raise your objection so we can further discuss. 

Thanks :-)

lebauce:
@orozery Just wondering what is the use case for using S3 to store flows ? Is it to use store flows before putting the later in an other datastore for further analysis (such as Hadoop) ?

config/config.go
```diff
@@ -156,18 +156,29 @@ func init() {
...
+	cfg.SetDefault("storage.objectstorage.access_key", "user")                                     // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.secret_key", "password")                                 // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.entity_id", "0000")                                      // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.collector_type", "flows")                                // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.collector_id", 0)                                        // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.root_output_dir", "logs")                                // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.max_seconds_per_stream", 86400)                          // defined for backward compatibility and to set defaults
```
hunchback: it may be clearer to write `24 * 60 * 60` rather than `86400`

flow/storage/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,204 @@
...
+	"compress/gzip"
+	"encoding/json"
+	"fmt"
+	"math"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/google/gopacket/layers"
```
hunchback: add a space after this line to separate skydive imports from external imports

flow/storage/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,204 @@
...
+		"last-timestamp":  aws.String(strconv.FormatInt(lastTime, 10)),
+		"num-records":     aws.String(strconv.Itoa(len(flows))),
+	}
+
+	// gzip
+	var b bytes.Buffer
+	w := gzip.NewWriter(&b)
+	w.Write([]byte(flowsString))
+	w.Close()
+
```
hunchback: remove line

flow/storage/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,204 @@
...
+// StoreFlows writes a set of flows to the object storage service
+func (c *ObjectStorage) StoreFlows(flows []*flow.Flow) error {
+	if len(flows) == 0 {
+		return nil
+	}
+
+	flowsString, err := json.Marshal(flows)
+	if err != nil {
+		logging.GetLogger().Error("Error encoding flows: " + err.Error())
```
hunchback: s/err.Error()/err/

contrib/ansible/roles/objectstorage/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: Set facts
+  set_fact:
+    minio_path: "/opt/minio"
```
hunchback: place in  `/usr/bin/minio`

contrib/ansible/roles/objectstorage/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    minio_client_path: "/opt/mc"
```
hunchback: place in `/usr/bin/mc`

contrib/ansible/roles/objectstorage/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    minio_data_path: "/opt/minio-data"
```
hunchback: rename to `/opt/minio` or '/opt/minio/data`

contrib/ansible/roles/objectstorage/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    mode: 0775
+
+- name: Create minio data directory
+  file:
+    path: "{{ minio_data_path }}"
+    state: directory
+    mode: 0755
+
+- name: Run minio server
+  shell: "{{ minio_path }} server {{ minio_data_path }} &"
```
hunchback: You should rather setup a service - one option is to use scripts found here - https://github.com/minio/minio-service

flow/storage/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,204 @@
...
+
+	endpoint := config.GetConfig().GetString(path + ".endpoint")
+	region := config.GetConfig().GetString(path + ".region")
+	bucket := config.GetConfig().GetString(path + ".bucket")
+	accessKey := config.GetConfig().GetString(path + ".access_key")
+	secretKey := config.GetConfig().GetString(path + ".secret_key")
+	entityID := config.GetConfig().GetString(path + ".entity_id")
+	collectorType := config.GetConfig().GetString(path + ".collector_type")
+	collectorID := strconv.Itoa(config.GetConfig().GetInt(path + ".collector_id"))
+	rootOutputDir := config.GetConfig().GetString(path + ".root_output_dir")
```
hunchback: I don't see why we need four parameters:
- entity_id
- collector_type
- collector_id
- root_output_dir

These all are used to construct the path leading to where the data is stored - so maybe just use one parameter instead:
- data_path

hunchback:
@orozery - just to reiterate that this driver should be named `s3` not `objectstore` and for sure not `objectstorage` as its' currently named.

etc/skydive.yml.default
orozery: The main advantage of object storage is its price, so it is great for storing large amounts of data (say you want to keep an archive of years full of flow logs).
See Amazon VPC flow logs as example:
https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs.html

flow/storage/objectstorage/objectstorage_test.go
orozery: I don't really care for "real" flow values here. Why not just use empty flows like I'm doing?

config/config.go
```diff
@@ -156,18 +156,29 @@ func init() {
...
+	cfg.SetDefault("storage.orientdb.database", "Skydive")                                         // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.orientdb.username", "root")                                            // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.orientdb.password", "root")                                            // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.driver", "objectstorage")                                // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.endpoint", "http://s3.amazonaws.com")                    // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.region", "us-east-1")                                    // defined for backward compatibility and to set defaults
+	cfg.SetDefault("storage.objectstorage.bucket", "bucket")                                       // defined for backward compatibility and to set defaults
```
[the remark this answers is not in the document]
orozery: I think bucket is more a generic name that people can understand.
Naming it skydive may confuse one with thinking that the bucket must be named skydive for this driver to work (?)
If you insist, I will change it though...

config/config.go
```diff
@@ -156,18 +156,29 @@ func init() {
...
+	cfg.SetDefault("storage.objectstorage.access_key", "AKIAIOSFODNN7EXAMPLE")                     // defined for backward compatibility and to set defaults
```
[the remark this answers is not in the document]
orozery: This was change to user:password

orozery:
I made changes to the ansible scripts. How do I test them?

flow/storage/objectstorage/objectstorage.go
orozery: you are right.
searchflows should be used only for testing. not for actual querying.
object storage does not support indexing.

contrib/ansible/roles/objectstorage/tasks/main.yml
orozery: changed to /usr/local/bin/minio, as desired by the minio.service

contrib/ansible/roles/objectstorage/tasks/main.yml
orozery: changed to /usr/local/bin/mc

flow/storage/objectstorage/objectstorage.go
orozery: added object_prefix parameter to replace all of these, as you suggested

config/config.go
```diff
@@ -168,14 +168,14 @@ func init() {
...
 	cfg.SetDefault("storage.orientdb.password", "root")              // defined for backward compatibility and to set defaults
-	cfg.SetDefault("storage.objectstorage.driver", "objectstorage")
-	cfg.SetDefault("storage.objectstorage.endpoint", "http://localhost:9000")
-	cfg.SetDefault("storage.objectstorage.region", "local")
-	cfg.SetDefault("storage.objectstorage.bucket", "bucket")
-	cfg.SetDefault("storage.objectstorage.access_key", "user")
-	cfg.SetDefault("storage.objectstorage.secret_key", "password")
-	cfg.SetDefault("storage.objectstorage.object_prefix", "flow_logs")
-	cfg.SetDefault("storage.objectstorage.max_seconds_per_stream", 24 * 60 * 60)
+	cfg.SetDefault("storage.objectstore.driver", "objectstore")
```
lebauce: s/objectstore/s3

lebauce:
@orozery @hunchback We are thinking about using the `/ws/subscriber` WebSocket endpoint so that you can get notified about the flows seen by Skydive. This way, S3 doesn't have to be a real flow backend. We'll push a POC as soon as possible.

By the way, `objectstore != S3` :-) (I'm thinking about OpenStack Swift among others)

hunchback:
> @orozery @hunchback We are thinking about using the `/ws/subscriber` WebSocket endpoint so that you can get notified about the flows seen by Skydive. This way, S3 doesn't have to be a real flow backend. We'll push a POC as soon as possible.
> 
> By the way, `objectstore != S3` :-) (I'm thinking about OpenStack Swift among others)

@lebauce I would think such use of the subscriber is less suitable for a high performance s3 backend ... so our thinking was more in the direction having both types of backends:
- storage (with search capabilities)
- archive (with read but not search). 

For now given that this "archive" thing is not defined we wanted to implement s3 as a storage with limited-search only to be used for testing scenarios.

hunchback:
ok, i agree we should start with the ws api for now.

safchain:
@orozery @hunchback I submitted a patch that adds a flow websocket endpoint. Skydive websocket client (golang/python) can be used to get the flows
#1470

safchain:
I think most of the current implementation could be reused, using the skydive websocket package. Maybe we could put this subscriber into the contrib folder ?

hunchback:
+1

orozery:
@nplanel @hunchback I switched to use flow subscriber. Your feedback will be appreciated.

contrib/objectstore/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/skydive-project/skydive/contrib/objectstore/subscriber"
+	shttp "github.com/skydive-project/skydive/http"
+	"github.com/skydive-project/skydive/logging"
+)
+
+const defaultConfigurationFile = "/etc/skydive/skydive-objectstore.yml"
+
+func main() {
+	if err := config.InitConfig("file", []string{defaultConfigurationFile}); err != nil {
+		logging.GetLogger().Error(fmt.Errorf("Failed to initialize config: %s", err.Error()))
```
safchain: nit: GetLogger().Errorf instead of Error(fmt.

contrib/objectstore/main.go
```diff
@@ -0,0 +1,73 @@
...
+		os.Exit(1)
+	}
+
+	if err := config.InitLogging(); err != nil {
+		logging.GetLogger().Error(fmt.Errorf("Failed to initialize logging system: %s", err.Error()))
```
safchain: ditto

contrib/objectstore/main.go
```diff
@@ -0,0 +1,73 @@
...
+	maxSecondsPerStream := cfg.GetInt("max_seconds_per_stream")
+
+	authOptions := &shttp.AuthenticationOpts{
+		Username: subscriberUsername,
+		Password: subscriberPassword,
+	}
+
+	subscriberURL, err := url.Parse(subscriberURLString)
+	if err != nil {
+		logging.GetLogger().Error(fmt.Errorf("Failed to parse subscriber URL: %s", err.Error()))
```
safchain: ditto

contrib/objectstore/Makefile
```diff
@@ -0,0 +1,11 @@
+GOVENDOR:=${GOPATH}/bin/govendor
```
safchain: I would call this Makefile from the contrib target, llike the `snort` one:
https://github.com/skydive-project/skydive/blob/master/Makefile#L354

So that we are sure that we won't break it changing internal sutff

orozery:
@safchain I see that it fails building with the aws-sdk-go now that I moved to contrib.
I did add this dependency to the global `vendor.json`.
Should I try something else?

safchain:
I think you need to 

```
govendor fetch ...
```

all the packages that are missing and reported here:
http://ci-logs.skydive.network/skydive-unit-tests/builds/2313/log.html

orozery:
@safchain dependencies are now fixed

Makefile
```diff
@@ -349,10 +349,12 @@ static: skydive.clean govendor genlocalfiles
 .PHONY: contribs.clean
 contribs.clean:
 	$(MAKE) -C contrib/snort clean
+	$(MAKE) -C contrib/objectstore clean
```
safchain: Sorry a final remark, I think objectstore is too generic especially since the implementation is based on aws. I would use awss3 or something similar.

orozery: @safchain It supports all S3-compliant object stores, which includes: IBM Cloud Object Storage and Google Cloud Storage.
Also, the s3-specific part of the code is modular, and can be easily extended to support other object storage APIs (via the `Client` interface).

PR #1423, opened 2018-10-25
libvirt: initial version of libvirt probe

### LINKED ISSUE #1410: Support libvirt

pierrecregut:
Libvirt probe dynamically polls the status of libvirt domains and their network
interfaces. For each new interface, it tries to enrich the coresponding tun
device with the name of the domain, the guest mac and the complete address of the
device on the guest pci bus.

The probe relies on libvirt-go the official go binding for libvirt that requires
the c libvirt library.

closes #1410

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -80,6 +80,7 @@ func init() {
 	cfg.SetDefault("agent.flow.pcapsocket.max_port", 8132)
 	cfg.SetDefault("agent.listen", "127.0.0.1:8081")
 	cfg.SetDefault("agent.topology.probes", []string{"ovsdb"})
+	cfg.SetDefault("agent.topology.libvirt.url", "qemu:///system")
 	cfg.SetDefault("agent.topology.netlink.metrics_update", 30)
 	cfg.SetDefault("agent.topology.neutron.domain_name", "Default")
 	cfg.SetDefault("agent.topology.neutron.endpoint_type", "public")
--- a/contrib/ansible/roles/skydive_dev/tasks/main.yml
+++ b/contrib/ansible/roles/skydive_dev/tasks/main.yml
@@ -27,6 +27,7 @@
      - rpm-build
      - rpmlint
      - libvirt
+     - libvirt-devel
      - libvirt-client
      - vagrant-libvirt
      - ansible
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -216,6 +216,9 @@ agent:
       interfaces:
         # - eth0
 
+    libvirt:
+      # url: qemu:///system
+
   capture:
     # Period in second to get capture stats from the probe. Note this
     # stats_update: 1
--- a/topology/graph/processor.go
+++ b/topology/graph/processor.go
@@ -1,0 +1,125 @@
+/*
+ * Copyright (C) 2018 Orange
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package graph
+
+import (
+	"github.com/skydive-project/skydive/common"
+)
+
+// NodeAction is a callback to perform on a node. The action is kept
+// active as long as it returns true.
+type NodeAction interface {
+	ProcessNode(g *Graph, n *Node) bool
+}
+
+// deferred represents a node action with additional info if needed
+// for cancellation.
+type deferred struct {
+	action NodeAction
+}
+
+// Processor encapsulates an indexer that will process NodeActions
+// on the nodes that filter
+type Processor struct {
+	common.RWMutex
+	DefaultGraphListener
+	MetadataIndexer *MetadataIndexer
+	actions         map[string][]deferred
+}
+
+// NewProcessor creates a Processor on the graph g, a stream of
+// events controlled by listenerHandler, that match a first set
+// of metadata m. Actions will be associated to a given set
+// of values for indexes.
+func NewProcessor(g *Graph, listenerHandler ListenerHandler, m ElementMatcher, indexes ...string) (processor *Processor) {
+	processor = &Processor{
+		MetadataIndexer: NewMetadataIndexer(g, listenerHandler, m, indexes...),
+		actions:         make(map[string][]deferred),
+	}
+	processor.MetadataIndexer.AddEventListener(processor)
+	return
+}
+
+// DoAction will perform the action for nodes matching values.
+func (processor *Processor) DoAction(action NodeAction, values ...interface{}) {
+	processor.Lock()
+	defer processor.Unlock()
+	nodes, _ := processor.MetadataIndexer.Get(values...)
+	kont := true
+	for _, node := range nodes {
+		kont = action.ProcessNode(processor.MetadataIndexer.graph, node)
+		if !kont {
+			break
+		}
+	}
+	if kont {
+		act := deferred{action: action}
+		hash := processor.MetadataIndexer.Hash(values...)
+		if actions, ok := processor.actions[hash]; ok {
+			actions = append(actions, act)
+		} else {
+			actions := []deferred{act}
+			processor.actions[hash] = actions
+		}
+	}
+}
+
+// Start starts the processor
+func (processor *Processor) Start() {
+	processor.MetadataIndexer.Start()
+}
+
+// Stop stops the processor
+func (processor *Processor) Stop() {
+	processor.MetadataIndexer.Stop()
+}
+
+// Cancel the actions attached to a given set of values.
+func (processor *Processor) Cancel(values ...interface{}) {
+	processor.Lock()
+	defer processor.Unlock()
+	hash := processor.MetadataIndexer.Hash(values...)
+	delete(processor.actions, hash)
+}
+
+// OnNodeAdded event
+func (processor *Processor) OnNodeAdded(n *Node) {
+	if values, err := getFieldsAsArray(n, processor.MetadataIndexer.indexes); err == nil {
+		hash := processor.MetadataIndexer.Hash(values...)
+		processor.Lock()
+		defer processor.Unlock()
+		if actions, ok := processor.actions[hash]; ok {
+			var keep []deferred
+			for _, action := range actions {
+				if action.action.ProcessNode(processor.MetadataIndexer.graph, n) {
+					keep = append(keep, action)
+				}
+			}
+			if len(keep) == 0 {
+				delete(processor.actions, hash)
+			} else {
+				processor.actions[hash] = keep
+			}
+		}
+	}
+}
--- a/topology/probes/libvirt/libvirt.go
+++ b/topology/probes/libvirt/libvirt.go
@@ -1,0 +1,302 @@
+// +build libvirt,linux
+
+/*
+ * Copyright (C) 2018 Orange.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package libvirt
+
+import (
+	"context"
+	"encoding/xml"
+	"fmt"
+	"sync"
+
+	"github.com/skydive-project/skydive/config"
+
+	libvirtgo "github.com/libvirt/libvirt-go"
+	"github.com/skydive-project/skydive/logging"
+	"github.com/skydive-project/skydive/topology/graph"
+)
+
+// Probe is the libvirt probe
+type Probe struct {
+	sync.Mutex
+	graph        *graph.Graph          // the graph
+	conn         *libvirtgo.Connect    // libvirt conection
+	interfaceMap map[string]*Interface // Found interfaces not yet connected.
+	cidLifecycle int                   // libvirt callback id of monitor to unregister
+	cidDevAdded  int                   // second monitor on devices added to domains
+	uri          string                // uri of the libvirt connection
+	cancel       context.CancelFunc    // cancel function
+	root         *graph.Node           // root node for ownership
+	tunProcessor *graph.Processor      // metadata indexer
+}
+
+// Address describes the XML coding of the pci addres of an interface in libvirt
+type Address struct {
+	Type     string `xml:"type,attr"`
+	Domain   string `xml:"domain,attr"`
+	Bus      string `xml:"bus,attr"`
+	Slot     string `xml:"slot,attr"`
+	Function string `xml:"function,attr"`
+}
+
+// DomainStateMap stringifies the state of a domain
+var DomainStateMap map[libvirtgo.DomainState]string = map[libvirtgo.DomainState]string{
+	libvirtgo.DOMAIN_NOSTATE:     "UNDEFINED",
+	libvirtgo.DOMAIN_RUNNING:     "UP",
+	libvirtgo.DOMAIN_BLOCKED:     "BLOCKED",
+	libvirtgo.DOMAIN_PAUSED:      "PAUSED",
+	libvirtgo.DOMAIN_SHUTDOWN:    "DOWN",
+	libvirtgo.DOMAIN_CRASHED:     "CRASHED",
+	libvirtgo.DOMAIN_PMSUSPENDED: "PMSUSPENDED",
+	libvirtgo.DOMAIN_SHUTOFF:     "DOWN",
+}
+
+// Interface is XML coding of an interface in libvirt
+type Interface struct {
+	Mac struct {
+		Address string `xml:"address,attr"`
+	} `xml:"mac"`
+	Target struct {
+		Device string `xml:"dev,attr"`
+	} `xml:"target"`
+	Address Address `xml:"address"`
+	Alias   struct {
+		Name string `xml:"name,attr"`
+	} `xml:"alias"`
+	Host *graph.Node `xml:"-"`
+}
+
+// Domain is the subset of XML coding of a domain in libvirt
+type Domain struct {
+	Interfaces []Interface `xml:"devices>interface"`
+}
+
+// getDomainInterfaces uses libvirt to get information on the interfaces of a
+// domain.
+func (probe *Probe) getDomainInterfaces(
+	domain *libvirtgo.Domain, // domain to query
+	domainNode *graph.Node, // Node representing the domain
+	constraint string, // to restrict the search to a single interface (by alias)
+) (interfaces []*Interface) {
+	rawXML, err := domain.GetXMLDesc(0)
+	if err != nil {
+		logging.GetLogger().Errorf("Cannot get XMLDesc: %s", err)
+		return
+	}
+	d := Domain{}
+	if err = xml.Unmarshal([]byte(rawXML), &d); err != nil {
+		logging.GetLogger().Errorf("XML parsing error: %s", err)
+		return
+	}
+	for _, itf := range d.Interfaces {
+		if constraint == "" || constraint == itf.Alias.Name {
+			itf.Host = domainNode
+			interfaces = append(interfaces, &itf)
+		}
+	}
+	return
+}
+
+// registerInterfaces puts the information collected in the graph
+// interfaces is an array of collected information.
+func (probe *Probe) registerInterfaces(interfaces []*Interface) {
+	probe.graph.Lock()
+	defer probe.graph.Unlock()
+	for _, itf := range interfaces {
+		name := itf.Target.Device
+		if name == "" {
+			continue
+		}
+		logging.GetLogger().Debugf(
+			"Libvirt interface %s on %s", name, itf.Host)
+		probe.tunProcessor.DoAction(itf, name)
+	}
+}
+
+// ProcessNode adds the libvirt interface information to a node of the graph
+func (itf *Interface) ProcessNode(g *graph.Graph, node *graph.Node) bool {
+	logging.GetLogger().Debugf("enrich %s", itf.Alias.Name)
+	tr := g.StartMetadataTransaction(node)
+	tr.AddMetadata("Libvirt.MAC", itf.Mac.Address)
+	tr.AddMetadata("Libvirt.Domain", itf.Host)
+	address := itf.Address
+	formatted := fmt.Sprintf(
+		"%s:%s.%s.%s.%s", address.Type, address.Domain, address.Bus,
+		address.Slot, address.Function)
+	tr.AddMetadata("Libvirt.Address", formatted)
+	tr.AddMetadata("Libvirt.Alias", itf.Alias.Name)
+	tr.AddMetadata("PeerIntfMAC", itf.Mac.Address)
+	tr.Commit()
+	if !g.AreLinked(node, itf.Host, graph.Metadata{"RelationType": "vlayer2"}) {
+		g.Link(node, itf.Host, graph.Metadata{"RelationType": "vlayer2"})
+	}
+	return false
+}
+
+// getDomain access the graph node representing a libvirt domain
+func (probe *Probe) getDomain(d *libvirtgo.Domain) *graph.Node {
+	domainName, _ := d.GetName()
+	probe.graph.RLock()
+	defer probe.graph.RUnlock()
+	return probe.graph.LookupFirstNode(graph.Metadata{"Name": domainName, "Type": "libvirt"})
+}
+
+// createOrUpdateDomain creates a new graph node representing a libvirt domain
+// if necessary and updates its state.
+func (probe *Probe) createOrUpdateDomain(d *libvirtgo.Domain) *graph.Node {
+	g := probe.graph
+	g.Lock()
+	defer g.Unlock()
+	domainName, _ := d.GetName()
+	metadata := graph.Metadata{
+		"Name": domainName,
+		"Type": "libvirt",
+	}
+	domainNode := g.LookupFirstNode(metadata)
+	if domainNode == nil {
+		domainNode = g.NewNode(graph.GenID(), metadata)
+		g.Link(probe.root, domainNode, graph.Metadata{"RelationType": "ownership"})
+	}
+	state, _, err := d.GetState()
+	if err != nil {
+		logging.GetLogger().Errorf("Cannot update domain state for %s", domainName)
+	} else {
+		tr := g.StartMetadataTransaction(domainNode)
+		defer tr.Commit()
+		tr.AddMetadata("State", DomainStateMap[state])
+	}
+	return domainNode
+}
+
+// deleteDomain deletes the graph node representing a libvirt domain
+func (probe *Probe) deleteDomain(d *libvirtgo.Domain) {
+	domainNode := probe.getDomain(d)
+	if domainNode != nil {
+		probe.graph.Lock()
+		defer probe.graph.Unlock()
+		probe.graph.DelNode(domainNode)
+	}
+}
+
+// Start get all domains attached to a libvirt connection
+func (probe *Probe) Start() {
+	// The event loop must be registered WITH its poll loop active BEFORE the
+	// connection is opened. Otherwise it just does not work.
+	if err := libvirtgo.EventRegisterDefaultImpl(); err != nil {
+		logging.GetLogger().Errorf("libvirt event handler:  %s", err)
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	probe.cancel = cancel
+	go func() {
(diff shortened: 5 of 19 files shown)


pierrecregut:
There is an alternative libvirt library go-libvirt from digitalocean that is pure go. So one is official and so probably more stable while the other requires no additional c library. I made a choice but I can change it if go-libvirt correctly supports events.

pierrecregut:
Clearly libvirt-devel is not in the CI.

lebauce:
run tests

lebauce:
@pierrecregut Thank you for this, it's really appreciated.

Regarding the libvirt bindings, in general, we prefer to use pure Go libraries as it allows us to compile the probe statically without relying on a shared library on the system. I took a brief look at the DigitalOcean you pointed out. It seems to support events and a part of it is generated from XML definition from upstream `libvirt` and seems to be well maintained. It would be really nice if you could use it. Otherwise, we can do this in a following PR.

pierrecregut:
In fact it does not: only a very limited kind of events are supported. The types are recognized but nothing is done to handle the events. If you want to receive lifecycle events ;first you must call `ConnectDomainEventRegister`, so far so good and you have a callback id (always 0 for these, can be something else for `NetworkLifecycleEvent`).

The events will be received on a custom rpc id : 1 for which no callback is registered in private method rpc.callback. The payload is a `DomainEventLifecycleMsg`. I would have expected a `DomainEventCallbackLifecycleMsg`, because now I don't know how to guess the callback id I should associate to them. Other messages like `NetworkEventLifecycleMsg` contain the `CallbackId` and then it is easy to associate the right handler and so the right decoder for them.

The conclusions are:
- go-libvirt lacks the connection abstractions provided by libvirt : you connect directly on a socket that could be anywhere. I prefer "qemu:///system"
- go-libvirt does no support lifecycle events out of the box
- The minimal patch to ask from DigitalOcean is for the ability to register a custom handler on RPC messages with id 1
- Even then developing a clean well typed solution is not easy.

lebauce:
@pierrecregut Fair enough. We'll see later if we can move to a pure Go implementation. Let's focus on have something working :-)

pierrecregut:
New version should now compile on windows (libvirt only done with linux build tag). But that will probably not solve the failure on ppc64

lebauce:
@pierrecregut Can you add libvirt-devel to https://github.com/skydive-project/skydive/blob/master/contrib/ansible/roles/skydive_dev/tasks/main.yml#L55 (it's the ansible role that is used by the CI)

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+
+// NewProbe creates a libvirt topology probe
+func NewProbe(g *graph.Graph, uri string) *Probe {
+	probe := &Probe{
+		graph:        g,
+		interfaceMap: make(map[string]*Interface),
+		callbackID:   -1,
+		uri:          uri,
+	}
+	g.AddEventListener(probe)
```
lebauce: You could create an indexer like this :
```
tunIndexer := graph.NewMetadataIndexer(g, g, graph.Metadata{"Type": "tun"}, "Name")
```
then you can do:
```
tunIndexer.AddEventListener(probe)
```

You `OnNodeAdded/Updated/Deleted`  methods will be called a resource is indexed or uindexed

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// Domain is the subset of XML coding of a domain in libvirt
+type Domain struct {
+	Interfaces []Interface `xml:"devices>interface"`
+}
+
+// QueryDomain uses libvirt to get information on the interfaces of a domain.
+// domain is the libvirt domain queried and bag is an accumulator for results
+func (probe *Probe) QueryDomain(domain *libvirtgo.Domain, bag *[]*Interface) {
```
lebauce: nit: no need to export QueryDomain

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+	defer probe.graph.Unlock()
+	for _, itf := range interfaces {
+		name := itf.Target.Device
+		if name == "" {
+			continue
+		}
+		probe.interfaceMap[name] = itf
+		logging.GetLogger().Debugf(
+			"Libvirt interface %s on %s", name, itf.Host)
+		node := probe.graph.LookupFirstNode(
```
lebauce: With the indexer described above, you can do:
```
if nodes, _ := tunIndexer.Get(name); len(nodes) > 0 {
    node := nodes[0]
    ....
```

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+
+// NewProbeFromConfig initializes the probe
+func NewProbeFromConfig(g *graph.Graph) (*Probe, error) {
+	uri := config.GetString("libvirt.url")
+	probe := NewProbe(g, uri)
+	return probe, nil
+}
+
+// OnNodeUpdated event
+func (probe *Probe) OnNodeUpdated(n *graph.Node) {
```
lebauce: Not sure we should handle the case where the interface is renamed

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// OnNodeAdded event
+func (probe *Probe) OnNodeAdded(n *graph.Node) {
+	name, _ := n.GetFieldString("Name")
```
lebauce: Why the indexer, you could simply do:
```
	if itf, ok := probe.interfaceMap[name]; ok {
		probe.EnrichNode(n, itf)
	}
```

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+		c *libvirtgo.Connect, d *libvirtgo.Domain,
+		event *libvirtgo.DomainEventLifecycle,
+	) {
+		if event.Event == libvirtgo.DOMAIN_EVENT_STARTED {
+			var interfaces []*Interface
+			probe.QueryDomain(d, &interfaces)
+			probe.RegisterInterfaces(interfaces)
+		}
+	}
+	callbackID, err := conn.DomainEventLifecycleRegister(nil, callback)
```
lebauce: nit: `probe.callbackID, conn.DomainEventLifecycleRegister(nil, callback)`

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+	// The event loop must be registered WITH its poll loop active BEFORE the
+	// connection is opened. Otherwise it just does not work.
+	err := libvirtgo.EventRegisterDefaultImpl()
+	if err != nil {
+		logging.GetLogger().Errorf("libvirt event handler:  %s", err)
+	}
+
+	go func() {
+		for {
+			err = libvirtgo.EventRunDefaultImpl()
```
lebauce: nit: `if err := libvirtgo.EventRunDefaultImpl(); err != nil {`

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+// QueryDomain uses libvirt to get information on the interfaces of a domain.
+// domain is the libvirt domain queried and bag is an accumulator for results
+func (probe *Probe) QueryDomain(domain *libvirtgo.Domain, bag *[]*Interface) {
+	rawXML, err := domain.GetXMLDesc(0)
+	if err != nil {
+		logging.GetLogger().Error("Cannot get XMLDesc: %s", err)
+		return
+	}
+	d := Domain{}
+	err = xml.Unmarshal([]byte(rawXML), &d)
```
lebauce: nit: `if err = xml.Unmarshal([]byte(rawXML), &d); err != nil {`

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		logging.GetLogger().Error("XML parsing error: %s", err)
+		return
+	}
+	logging.GetLogger().Debugf("LIBVIRT DOMAIN %v", d)
```
lebauce: nit: remove log

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+	tr.AddMetadata("Libvirt.Address", formatted)
+	tr.AddMetadata("PeerIntfMAC", itf.Mac.Address)
+	tr.Commit()
+}
+
+// Start get all domains attached to a libvirt connection
+func (probe *Probe) Start() {
+	// The event loop must be registered WITH its poll loop active BEFORE the
+	// connection is opened. Otherwise it just does not work.
+	err := libvirtgo.EventRegisterDefaultImpl()
```
lebauce: nit: `if err := libvirtgo.EventRegisterDefaultImpl(); err != nil {`

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+	for i := range d.Interfaces {
+		itf := &d.Interfaces[i]
+		itf.Host = host
+		*bag = append(*bag, itf)
+	}
+}
+
+// RegisterInterfaces puts the information collected in the graph
+// interfaces is an array of collected information.
+func (probe *Probe) RegisterInterfaces(interfaces []*Interface) {
```
lebauce: nit: no need to expert RegisterInterfaces

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,238 @@
...
+		node := probe.graph.LookupFirstNode(
+			graph.Metadata{"Name": name, "Type": "tun"})
+		if node != nil {
+			probe.EnrichNode(node, itf)
+		}
+	}
+}
+
+// EnrichNode adds the libvirt interface information to a node of the graph
+func (probe *Probe) EnrichNode(node *graph.Node, itf *Interface) {
```
lebauce: nit: no need to export EnrichNode

etc/skydive.yml.default
```diff
@@ -288,6 +288,9 @@ ovs:
 docker:
   # url: unix:///var/run/docker.sock
 
+libvirt:
```
lebauce: We are in the process of moving the agent probes configuration from the top level.
Can you please move it to `agent.topology.probe.libvirt` ?

topology/probes/libvirt/libvirt.go
pierrecregut: I have reverted this change. Copied as is, it does not capture tun nodes as they are added.

topology/probes/libvirt/libvirt.go
pierrecregut: I don't think you can rename libvirt interface like this. But the code did not handle the dynamic addition of interfaces and this has been corrected with a new callback.

pierrecregut:

sergeyglazyrindev:
I am sorry for the delay but as I promised I made a pull request with libvirt probe. Please check it out also.
And UI contribution is coming, I hope it would be amazing contribution

sergeyglazyrindev:
Here it is
https://github.com/skydive-project/skydive/pull/1424

pierrecregut:

config/config.go
```diff
@@ -80,6 +80,7 @@ func init() {
 	cfg.SetDefault("agent.flow.pcapsocket.max_port", 8132)
 	cfg.SetDefault("agent.listen", "127.0.0.1:8081")
 	cfg.SetDefault("agent.topology.probes", []string{"ovsdb"})
+	cfg.SetDefault("agent.topology.libvirt.url", "qemu:///system")
```
sergeyglazyrindev: I think it shoud be moved here https://github.com/skydive-project/skydive/pull/1424

contrib/ansible/roles/skydive_dev/tasks/main.yml
```diff
@@ -27,6 +27,7 @@
      - rpm-build
      - rpmlint
      - libvirt
+     - libvirt-devel
```
sergeyglazyrindev: I think it shoud be moved here #1424

etc/skydive.yml.default
```diff
@@ -220,6 +220,9 @@ agent:
       interfaces:
         # - eth0
 
+    libvirt:
```
sergeyglazyrindev: I think it shoud be moved here #1424

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		logging.GetLogger().Errorf(
+			"Could not register the lifecycle event handler %s", err)
+	}
+	callbackDeviceAdded := func(
+		c *libvirtgo.Connect, d *libvirtgo.Domain,
+		event *libvirtgo.DomainEventDeviceAdded,
+	) {
+		var interfaces []*Interface
+		probe.queryDomain(d, &interfaces, event.DevAlias)
```
sergeyglazyrindev: this logic may be moved also here
https://github.com/skydive-project/skydive/pull/1424

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// OnNodeAdded event
+func (probe *Probe) OnNodeAdded(n *graph.Node) {
+	if typ, _ := n.GetFieldString("Type"); typ != "tun" {
+		return
+	}
+	name, _ := n.GetFieldString("Name")
+	logging.GetLogger().Debugf("Tun interface %s", name)
+	if itf, ok := probe.interfaceMap[name]; ok {
```
sergeyglazyrindev: I think here is an issue: how would this code behave if one interface being attached to multiple vms ? Is it possible ? Not sure about if its possible in hypervisor... but anyway.....
I propose to use my realtime handler which I commited here.
It solves a lot of issues and abstracts away all needed interaction with graph from concrete probe. It will help to reuse this package for future probes... What do you think ? Here is the link to the package which may handle this problem in quite good way
https://github.com/skydive-project/skydive/pull/1424/files#diff-c9c7c72fb06c6b8a0ed7b4c81a3bf3f4

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,267 @@
...
+		node := probe.graph.LookupFirstNode(
+			graph.Metadata{"Type": "tun", "Name": name})
+		if node != nil {
+			probe.enrichNode(node, itf)
+		}
+	}
+}
+
+// enrichNode adds the libvirt interface information to a node of the graph
+func (probe *Probe) enrichNode(node *graph.Node, itf *Interface) {
```
sergeyglazyrindev: I do love this extended metadata information. We can move it here: https://github.com/skydive-project/skydive/pull/1424

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	name, _ := n.GetFieldString("Name")
+	logging.GetLogger().Debugf("Tun interface %s", name)
+	if itf, ok := probe.interfaceMap[name]; ok {
+		probe.enrichNode(n, itf)
+	}
+}
+
+// OnNodeDeleted event
+func (probe *Probe) OnNodeDeleted(n *graph.Node) {
```
sergeyglazyrindev: I think we don't need this code. Basically, the interfaces will be removed when libvirt machine stops and then these interfaces will be added back when it will be started, so better to rely on handling of events from libvirt like it is done here.
https://github.com/skydive-project/skydive/pull/1424/files#diff-9934bc7c03f53af658e6552f294f1b7cR107

pierrecregut:
@sergeyglazyrindev Your proposal makes sense. I am on leave this week and I will look at it carefully next monday.

Just a little bit of background on my proposal. All I was really interested in is filling this IntfPeerMac field. There are no VM node and interface nodes make only sense when they exist and the VM is live.
So the probe is considered as a secondary probe that only make sense if there is another one that creates the real interface nodes. The libvirt probe adds additional information to those nodes.

Interface (the struct type I use) are kept as representing this additional information. They should probably be thrown away as soon as the graph node is enriched and not when it disappear. Keeping this kind of information only makes sense if you have some complex updates to perform. This is not the case. This is probably what your code solves more elegantly through the abstraction realtime (I will look at it more seriously next week) even if I am not sure that is the right name. I have more the feeling that we are trying to create a synchronous framework from asynchronous streams of events.

sergeyglazyrindev:
hi @pierrecregut 

> Just a little bit of background on my proposal. All I was really interested in is filling this IntfPeerMac field. There are no VM node and interface nodes make only sense when they exist and the VM is live.
> So the probe is considered as a secondary probe that only make sense if there is another one that creates the real interface nodes. The libvirt probe adds additional information to those nodes.

In my implementation we add real vm node to the graph add add layer2 connections to all network interfaces vm connected through
Would be that ok for your needs if we add this metadata only to real vm node ?
Why do you want to have all this information for each interface vm connected through ? Basically, this metadata is the same for all nodes vm connected through. I think adding real libvirt node to the graph is better solution.
What do you think ?

> Interface (the struct type I use) are kept as representing this additional information. They should probably be thrown away as soon as the graph node is enriched and not when it disappear. Keeping this kind of information only makes sense if you have some complex updates to perform. This is not the case. This is probably what your code solves more elegantly through the abstraction realtime (I will look at it more seriously next week) even if I am not sure that is the right name. I have more the feeling that we are trying to create a synchronous framework from asynchronous streams of events.

About your comment regarding realtime "package", it's not a synchronous framework, it's just a try to decouple work related to graph from the real probe. I don't want to complicate it in future. But it would be useful for some easy graph interactions. About complicated graph interactions, for me it's ok to have it inside the probe. But it all depends on the task requirements, etc

While you'll be on your leave, I'll move all things from your changeset to my. Enjoy your leave.

pierrecregut:
@sergeyglazyrindev The information I add is specific to each interface: it has a unique mac address in the guest and a unique pci bus address from which its name will be derived. 
As far as I understand it, your framework is not tailored for this kind of cooperative enrichment of nodes but to create links between nodes. It could be used by the peering probe but I guess that the peering probe as it is could also be enriched to add links between arbitrary nodes meeting some criteria.

On the other hand I have no domain node which may be useful for cases where there is no agent on the guest. For example to link multiple interfaces to a given guest. The next question is what we do when we have both the "libvirt guest node" in the host and the "host node" of the agent in the guest. They represent the same entity, should we enhance the peering probe to link them ? There is also an issue because now we have two L2 links from the interface on the host that represent the same datapath but one linked to the interface in the guest and the other to the virtual guest node. It is clear that I would prefer to keep only the one linked to the real interface.

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,267 @@
...
+
+// Start get all domains attached to a libvirt connection
+func (probe *Probe) Start() {
+	// The event loop must be registered WITH its poll loop active BEFORE the
+	// connection is opened. Otherwise it just does not work.
+	if err := libvirtgo.EventRegisterDefaultImpl(); err != nil {
+		logging.GetLogger().Errorf("libvirt event handler:  %s", err)
+	}
+
+	go func() {
```
pierrecregut: Should use a context.

sergeyglazyrindev:
Hi, I'll reply today

pierrecregut:
I have uploaded a new version that mix my enrichment of nodes with Sergei approach of adding a node per domain. Note that my links with the virtual domain are tagged vlayer2 and not layer2 to avoid this:

screenshot_20181105_174857

Members of interfaceMap are now thrown as soon as they are consumed. It also correctly lock the use of interfaceMap.

pierrecregut:

pierrecregut:
run skydive-ppc64le-tests

pierrecregut:
run skydive-ppc64le-tests

pierrecregut:
run skydive-ppc64le-tests

topology/probes/libvirt/libvirt.go
lebauce: @pierrecregut ping

sergeyglazyrindev:
> I have uploaded a new version that mix my enrichment of nodes with Sergey approach of adding a node per domain. Note that my links with the virtual domain are tagged vlayer2 and not layer2 to avoid this:
> 
> screenshot_20181105_174857
> 
> Members of interfaceMap are now thrown as soon as they are consumed. It also correctly lock the use of interfaceMap.

hi, thank you, I still think we should split responsibilites of the probe: communication with graph, waiting for realtime changes, and getting information from libvirt

topology/probes/libvirt/libvirt.go
pierrecregut: @lebauce I think I get what you mean. In that case we can get one step beyond. My map is only there to defer actions when the node does not exist. It is indexed in fact by what you call indexes in the MetadataIndexer, so we could wrap an indexer and a set of actions. This is what I call a processor (needs a better name) in the last iteration.

sergeyglazyrindev:
hi @pierrecregut 
I love your processor implementation. Thank you so much. So, take a lead, now I quite happy to see such libvirt probe implementation.

sergeyglazyrindev:
The only one thing: could you please add to the makefile part from my commit ? And compile libvirt.go file only if there's a tag libvirt. I think there should be use cases when we don't want to have suich probe: like separate jenkins jobs to compile skydive with different tags for different server roles, etc
please add something like:
if (WITH_LIBVIRT=true) {
   buildtags += libvirt
}
and then use this tag in libvirt.go

pierrecregut:
The changes for the newly pushed version are:
- added a libvirt tag controlled by WITH_LIBVIRT set to true by default.
- added unit tests for processor (and fixed a bug)
- changed libvirt commit to add coauthor (@sergeyglazyrindev please review if info ok).

pierrecregut:
It would be nice to have a functional test. One solution would be to leverage the devtack used for neutron probe test and create a VM (but finding the right image/flavour may be tricky). Another approach is calling a virt-install with the right parameters. This means that we have an image to launch.

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,309 @@
...
+	Type     string `xml:"type,attr"`
+	Domain   string `xml:"domain,attr"`
+	Bus      string `xml:"bus,attr"`
+	Slot     string `xml:"slot,attr"`
+	Function string `xml:"function,attr"`
+}
+
+// DomainStateMap stringifies the state of a domain
+var DomainStateMap map[libvirtgo.DomainState]string = map[libvirtgo.DomainState]string{
+	libvirtgo.DOMAIN_NOSTATE:     "Undefined",
```
lebauce: nit: why lowercase for this one ?

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,309 @@
...
+	if err != nil {
+		logging.GetLogger().Errorf("Cannot get XMLDesc: %s", err)
+		return
+	}
+	d := Domain{}
+	if err = xml.Unmarshal([]byte(rawXML), &d); err != nil {
+		logging.GetLogger().Errorf("XML parsing error: %s", err)
+		return
+	}
+	for i := range d.Interfaces {
```
lebauce: for _, itf := range d.Interfaces {

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,309 @@
...
+type Domain struct {
+	Interfaces []Interface `xml:"devices>interface"`
+}
+
+// queryDomain uses libvirt to get information on the interfaces of a domain.
+// domain is the libvirt domain queried and bag is an accumulator for results
+func (probe *Probe) queryDomain(
+	domain *libvirtgo.Domain, // domain to query
+	domainNode *graph.Node, // Node representing the domain
+	bag *[]*Interface, // to accumulate results
```
lebauce: nit: change the function to `func (probe *Probe) getDomainInterfaces(...) []*Interface`

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,309 @@
...
+	probe.cancel = cancel
+	go func() {
+		for ctx.Err() == nil {
+			if err := libvirtgo.EventRunDefaultImpl(); err != nil {
+				logging.GetLogger().Errorf("libvirt poll loop problem: %s", err)
+			}
+		}
+	}()
+
+	conn, err := libvirtgo.NewConnectReadOnly(probe.uri)
```
lebauce: nit: probe.conn, err = libvirtgo.NewConnectReadOnly(probe.uri)

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,309 @@
...
+		tunProcessor: tunProcessor,
+	}
+	tunProcessor.Start()
+	return probe
+}
+
+// NewProbeFromConfig initializes the probe
+func NewProbeFromConfig(g *graph.Graph, root *graph.Node) (*Probe, error) {
+	uri := config.GetString("agent.topology.libvirt.url")
+	probe := NewProbe(g, uri, root)
```
lebauce: nit: return NewProbe(g, config.GetString("agent.topology.libvirt.url"), root), nil

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,309 @@
...
+// interfaces is an array of collected information.
+func (probe *Probe) registerInterfaces(interfaces []*Interface) {
+	for _, itf := range interfaces {
+		name := itf.Target.Device
+		if name == "" {
+			continue
+		}
+		logging.GetLogger().Debugf(
+			"Libvirt interface %s on %s", name, itf.Host)
+		probe.tunProcessor.DoAction(itf, name)
```
lebauce: It seems that `DoAction` is not executed under the graph lock

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,309 @@
...
+	address := itf.Address
+	formatted := fmt.Sprintf(
+		"%s:%s.%s.%s.%s", address.Type, address.Domain, address.Bus,
+		address.Slot, address.Function)
+	tr.AddMetadata("Libvirt.Address", formatted)
+	tr.AddMetadata("Libvirt.Alias", itf.Alias.Name)
+	tr.AddMetadata("PeerIntfMAC", itf.Mac.Address)
+	tr.Commit()
+	go func() {
+		g.Lock()
```
lebauce: Why taking the lock for the link and not for the metadata ?
When called from `OnNodeAdded`, ProcessNode has the lock so you don't have to care about it.
When called from `DoAction`, there is no lock. So I suggest you make sure that DoAction is executed under the graph lock so you don't have to care about it in `ProcessNode`.

pierrecregut: There is a real problem with the graph API : 
* you have so called transactions and they do not care taking the lock. For metadata, nodes should have their own lock. That would be much cleaner.
* When called from OnNodeAdded, you probably have a Read Lock not a Write Lock. This is at least what you have with regular event listener and promoting a read lock is not possible.
So the solution is probably to put everything under the go func()

topology/graph/processor.go
```diff
@@ -0,0 +1,130 @@
...
+	processor = &Processor{
+		MetadataIndexer: NewMetadataIndexer(g, listenerHandler, m, indexes...),
+		actions:         make(map[string][]deferred),
+	}
+	processor.MetadataIndexer.AddEventListener(processor)
+	return
+}
+
+// DoAction willl perform the action for nodes matching
+func (processor *Processor) DoAction(action NodeAction, values ...interface{}) {
```
lebauce: Maybe calling it "RegisterAction" would make it clearer that the action will be repeated. What do you think ?

topology/graph/processor.go
```diff
@@ -0,0 +1,130 @@
...
+// for cancellation.
+type deferred struct {
+	action NodeAction
+}
+
+// Processor encapsulates an indexer that will process NodeActions
+// on the nodes that filter
+type Processor struct {
+	common.RWMutex
+	*MetadataIndexer
```
lebauce: As MetadataIndexer already implement the `EventListener` implement, it would be clearer if MetadataIndexer was not embeded but be a regular member of the Processor struct.
This way, you can embed the `DefaultGraphListener` structure so you don't have to define empty `OnNodeUpdated` and `OnNodeDeleted` functions.

topology/probes/libvirt/libvirt.go
lebauce: @pierrecregut In OnNodeAdded, you have a Write lock. So you safely update the metadata and create the link. Then, in DoAction, you can also get a Write lock so that ProcessNode is protected.

pierrecregut: ok. the confusion came from the fact that eventHandler have their own lock (read locked in notifyListeners) are part of the graph and are named g...

topology/probes/libvirt/libvirt.go
pierrecregut: ok

topology/probes/libvirt/libvirt.go
pierrecregut: ok

topology/probes/libvirt/libvirt.go
pierrecregut: yes. No need for 2 phase and a bag to collect all itfs.

topology/probes/libvirt/libvirt.go
pierrecregut: ok...

topology/probes/libvirt/libvirt.go
pierrecregut: Ok modified.

topology/graph/processor.go
pierrecregut: yes nicer.

sergeyglazyrindev:
@pierrecregut thank you so much for adding me to coauthors, I don't feel that I am coauthor :) I made almost nothing in your commit :) You did use correct email. So, thank you.

pierrecregut:

pierrecregut:
@sergeyglazyrindev the code handling the state of the domain node is extracted from your commit.

pierrecregut:

lebauce:

safchain:
run skydive-devstack-tests

lebauce:
@pierrecregut Could you please allow me to push some commits on top of your PR (https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) to fix packaging/docker/post merge jobs ? Thanks !

pierrecregut:
@lebauce  It should be already enabled. The "allow edit from maintainers" box is ticked

lebauce:
@pierrecregut My bad

lebauce:
run skydive-packaging-tests

lebauce:
run skydive-devstack-tests

lebauce:

lebauce:

PR #1424, opened 2018-10-26
area: graph - add libvirt probe

sergeyglazyrindev:
this probe responsible for handling libvirt functionality:
1. add existing vms into topology
2. updates its status in realtime
3. determines all interfaces virtual machines connected through

Closes-Bug: #SOFTID-197

Change-Id: Id604911e2d80b7897545caea2c1f19a7a5826f84
Signed-off-by: Sergey Glazyrin <s.glazyrin@partner.samsung.com>


topology/probes/libvirt/helpers/network.go
```diff
@@ -0,0 +1,74 @@
...
+		Dev string `xml:"dev,attr"`
+	} `xml:"target"`
+	Vlan struct {
+		Tag struct {
+			Id string `xml:"id,attr"`
+		} `xml:"tag"`
+	} `xml:"vlan"`
+}
+
+func GetAllInterfacesFromXmlString(XMLDesc string) *list.List {
```
sergeyglazyrindev: this function could be improved a lot using xml.UnMarshall approach.

topology/graph/realtime/realtime.go
```diff
@@ -0,0 +1,123 @@
+/**
```
sergeyglazyrindev: this abstraction would be useful to reuse in different probes, I'd like to place only specific logic unique to the probe in probe but communication with graph may be handled in this package: realtime

sergeyglazyrindev:
Sriov probe not ready yet
I'd love to  investigate how can we catch up the realtime changes, maybe there are kernel events, and also right now there's one issue: if we put down the physical interface which provides VF (for instance using ifconfig eth0 down) then status of node still UP until we restart skydive agent. Probably we can catch it using kernel events, not sure yet, will check it on next week.

sergeyglazyrindev:
@safchain I'll split up this commit into two.

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,234 @@
...
+
+	probe.Cancel = cancel
+	probe.Wg.Add(1)
+
+	probe.Connected.Store(true)
+	defer probe.Connected.Store(false)
+
+	go func() {
+		defer probe.Wg.Done()
+		defer probe.periodicLibvirtChecker()
```
pierrecregut: I am surprised that you had no problem with the loop containing EventRunDefaultImpl not active when the libvirt connection is created. Moving it from the end of the startProbe code to the begining made my code work and it should definitely be active before the initial list of domains is performed.

sergeyglazyrindev: To be honest I am not sure, I tested adding/removing vms in realtime, the graph updated, etc
So, maybe you are right, but I've not seen any real issue with it.

topology/graph/realtime/realtime.go
lebauce: @sergeyglazyrindev It seems like this is very close to the `Indexer` and `MetadataIndexer` (topology/graph/indexer.go). Their goal is to listen for the graph and index nodes (based on metadata for instance). It is also possible to subscribe to these indexer (with AddEventListener, just like on the graph) to be notified when a node was added or removed to the index.

sergeyglazyrindev: hi @lebauce 
Hm, it looks similar... but I'll dig into it in few days. Thank you for pointing me to this existing stuff.

sergeyglazyrindev:
Hello, I was busy with gui contribution, wanted to reach some point...
tomorrow I'll reply about comments from guys who reviewed my changes and tomorrow I'll move changes from @pierrecregut to my changeset so we may contribute together and we may satisfy desires of everyone :)

topology/graph/realtime/realtime.go
```diff
@@ -0,0 +1,123 @@
...
+		if n == nil {
+			continue
+		}
+		nodeConnectedToType.ConnectTwoNodes(handler.Graph, n, Node)
+		delete(nodes, nodeIdentifier)
+	}
+	handler.StartTrackingChangesForNode(NodeIdentifier, nodes)
+}
+
+func MakeGraphHandler(g *graph.Graph, PropertyToIdentifyNode string) *GraphHandler {
```
pierrecregut: uncapitalize parameter name.

topology/probes/libvirt/helpers/network.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"bytes"
+	"container/list"
+	"encoding/xml"
+
+	"github.com/skydive-project/skydive/logging"
+	"github.com/skydive-project/skydive/topology/graph/realtime"
+)
+
+type Interface struct {
```
pierrecregut: Yes for a single use, better to have structs inlined.

Makefile
```diff
@@ -172,6 +172,10 @@ ifeq ($(WITH_LXD), true)
   BUILD_TAGS+=lxd
 endif
 
+ifeq ($(WITH_LIBVIRT), true)
+  BUILD_TAGS+=libvirt
+endif
```
pierrecregut: Will not be activated by default. On the other hand, it should only be activated on Linux.

sergeyglazyrindev:
Hi, I'll reply today.

topology/graph/realtime/realtime.go
sergeyglazyrindev: > @sergeyglazyrindev It seems like this is very close to the `Indexer` and `MetadataIndexer` (topology/graph/indexer.go and topology/graph/indexer_test.go). Their goal is to listen for the graph and index nodes (based on metadata for instance). It is also possible to subscribe to these indexer (with AddEventListener, just like on the graph) to be notified when a node was added or removed to the index.

Hello, I checked this stuff. It is different thing from what I developed. But it maybe used as a handler for indexers. The goal of this realtime package is to decouple and simplify writing new probes, the goal of new probes is to add a connections, discover new nodes, update metadata, etc (except core ones (netlink for instance))
According to SOLID development principle "Single responsibility principle
a class should have only a single responsibility (i.e. changes to only one part of the software's specification should be able to affect the specification of the class)'

I see at least two responsibilities in this libvirt probe:
1. get information from libvirt
2. track changes in probe and make corresponding changes in the graph: update metadata, add connections, etc

That's why I added realtime package, now I extended it, and it is able to track a changes in the probe and update metadata for node once it was added to the graph

Such modular system helps me to keep my code cleaner and maintainable, easily testable, etc
For instance, thankfully to this module I removed all locks from probe because it's not a job for probe to keep the consistent state of go maps, to make sure that we lock graph in proper places, I think this implementation is better than mixing two different responsibilities in one module.

sergeyglazyrindev:
> Hi @sergeyglazyrindev can you create two different PR, one for libvirt, one for sriov ? @pierrecregut also worked on a libvirt probe #1423 it could be nice to see if we/you can work on a merge of both approaches.

hello, sorry for delay, I was busy, I removed the commit related to sriov from branch and also merged the changes from @pierrecregut 's commit to mine. If you agree on the idea, then I can write extra tests, fix some failed checks, etc

Makefile
sergeyglazyrindev: it's hard to say, in my environment we have two separated jenkins jobs: build analyzer and agent separately, it's not mandatory for analyzers to have a libvirt installed on the system, it depends on your architecture, etc, so it's better to have a possibility to compile with libvirt support on demand.

topology/probes/libvirt/libvirt.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	return
+}
+
+func (probe *LibVirtProbe) isLibvirtCallbackConnected() bool {
+	return probe.libvirtCallbackId != -1
+}
+
+func (probe *LibVirtProbe) periodicLibvirtChecker() {
+	for true {
```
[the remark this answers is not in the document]
sergeyglazyrindev: that's good idea, but if I don't want to use anonymous functions, then I need to keep the context in the probe ? To be honest, I don't want to use anonymous functions for such things....

topology/probes/libvirt/libvirt.go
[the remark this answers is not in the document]
sergeyglazyrindev: ah, it makes sense

topology/probes/libvirt/helpers/network.go
```diff
@@ -0,0 +1,74 @@
...
+			Id string `xml:"id,attr"`
+		} `xml:"tag"`
+	} `xml:"vlan"`
+}
+
+func GetAllInterfacesFromXmlString(XMLDesc string) *list.List {
+	interfaceList := list.New()
+	buf := bytes.NewBufferString(XMLDesc)
+	decoder := xml.NewDecoder(buf)
+	var VMInterface Interface
```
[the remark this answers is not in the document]
sergeyglazyrindev: good point

agent/probes.go
```diff
@@ -107,6 +108,8 @@ func NewTopologyProbeBundleFromConfig(g *graph.Graph, hostNode *graph.Node) (*pr
 			probes[t] = opencontrail
 		case "socketinfo":
 			probes[t] = socketinfo.NewSocketInfoProbe(g, hostNode)
+		case "libvirt":
+			probes["libvirt"] = libvirt.NewProbe(g, n)
```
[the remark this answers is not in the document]
sergeyglazyrindev: is it mandatory to have an owner for each node in the graph ? to be honest I am not sure. It complicates the graph, is there any benefit ?

topology/graph/realtime/realtime.go
sergeyglazyrindev: hey @lebauce 
I am sorry for such stupid questions. But these indexers are a magic for me. Okay, let's split the tasks into independent ones:
1. if we want to connect two nodes automatically when it added to the graph, we may use MetadataLinker which will be able to link two nodes automatically. Is it correct ?
so, to solve it I may use such approach:
```
        // filter to match libvirt nodes
        libvirtFilter := NewElementFilter(filters.NewTermStringFilter("Type", "LibVirt"))
        indexer1 := NewMetadataIndexer(g, g, libvirtFilter, "Libvirt")
        indexer1.Start()

        // the filter to match nodes which are connected 
        libvirtPassThroughFilter := NewElementFilter(filters.NewNotNullFilter("Libvirt.Domain"))
        indexer2 := NewMetadataIndexer(g, g, libvirtPassThroughFilter, "libvirtPassThrough")
        indexer2.Start()

        linker := NewMetadataIndexerLinker(g, indexer1, indexer2, Metadata{"RelationType": "vlayer2"})
        linker.Start()

        **// and then when such nodes will be added to the graph it will be automatically connected
        // is it correct ?**

```

2. Update some metadata for nodes when it will be added to the graph

In theory we can use NewMetadataIndexer and add as eventlistener our probe and then handle updating of metadata for node directly in the probe ?

I've no found other way, is it what you suggest ? or is there more elegant way ?

sergeyglazyrindev: Btw , @lebauce , is there a way to create an ElementFilter with callable, something like this:
ElementFilter(FilterElementUsingCallable(func (n *graph.Node) { return true if element matches our expectations}))

sergeyglazyrindev: Aha, I think I figured it out....

lebauce: @sergeyglazyrindev I just pushed a POC of what it could look like when using existing framework here : https://github.com/lebauce/skydive/tree/libvirt-indexer

The 2 first commits are fixes to handle arrays. The last one is to make the libvirt probe use linkers and indexers.

sergeyglazyrindev: hello @lebauce 
Thank you, for me it's clear how to use linker, but there's still one problem, how can we schedule metadata update for node when it's added to the graph, @pierrecregut in his version of probe updates metadata for nodes vm connected through, and adding Interface->Address information to this node. Should we do that directly in a probe and subscribe it as eventlistener to one of metadata indexers ?

lebauce: @sergeyglazyrindev I'm not sure to understand what you mean by "schedule metadata update for node when it's added to the graph"

sergeyglazyrindev: I am talking about the thing provided by @pierrecregut .
https://github.com/skydive-project/skydive/pull/1423/files#diff-9934bc7c03f53af658e6552f294f1b7cR138
Previously he was listening on the graph events but now he added processor ( https://github.com/skydive-project/skydive/pull/1423/files#diff-9934bc7c03f53af658e6552f294f1b7cR138 ) which looks interesting for me. this processor decouples graph manipulation from the probe and for me it's good.

PR #1440, opened 2018-11-05
OpenFlow rule filter

dvandra:
OpenFlow rule filter: By Gremlin request with "OR" condition, using "as/select" method. (without removing "regex" and not using "HasEither".)


lebauce:
@dvandra Why closing this one and opening a new one ?

dvandra:
Some more files were added, those were not changed by me!

lebauce:
@dvandra You probably added files by mistake (git add on a folder maybe). Closing/reopening a PR would not fix the problem, you need to fix your local git branch

PR #1344, opened 2018-10-02
LLDP probe

lebauce:

PR #1450, opened 2018-11-08
security: Improve TLS

maglo:
The implementation of TLS in Skydive is severely lacking and a little bit unorthodox. 

Here is an attempt at making the implementation a little more standard and easier to understand.

**Agent Does not use TLS unless CA Certificate and Keys are specified**
Previously, the logic for checking if the admin wanted TLS was to check whether the `analyzer:cert` was specified. This is counter-intuitive and bad security practice.

This MR changes this so that the logic is changed. 

**Concepts regarding certificates and services confusing**
Previously, the code used `analyzer:cert` for server certificates, and `agent:cert` for client certificates. This gets confusing, since when Skydive runs in "agent" mode, a server is set up. Why would it use the certificates from the Analyzer configuration for that.

This MR proposes to move the specification of certificates to a separate node, like so:

```
tls:
 client_cert: /pki/client.crt.pem
 client_key: /pki/client.key.pem
 server_cert: /pki/server.crt.pem
 server_key: /pki/server.key.pem
 ca_cert: /pki/ca.crt.pem
```
When Skydive is started in Analyzer mode, it uses the Server certificates for the service, and uses Client certificates when connecting out to other Agents.

When Skydive is started in Agent mode, it also uses the Server certificates for the Agent service, and uses Client certificates when connecting to the Analyzers.

**Inappropriate use of certificate trust**
Previously, the examples and implementation did not allow for an administrator to dictate what certificates to trust in a secure way. 

This MR proposes to allow the admin to specify `tls:ca_cert` that is used to verify peers (either from client to server or from server to client). 

This way, the admin does not have to "give away the keys to the kingdom" by giving all Agents the certifcate AND key for the issuing CA.

It also matches better with the standard way of implementing PKI, where a Root CA issues a Subordinate CA that issues Certificates. The Root and Subordinate CAs private keys must be kept secret.

### DIFF
--- a/common/tls.go
+++ b/common/tls.go
@@ -29,8 +29,8 @@ import (
 	"io/ioutil"
 )
 
-// SetupTLSLoadCertificate creates a X509 certificate from file
-func SetupTLSLoadCertificate(certPEM string) (*x509.CertPool, error) {
+// SetupTLSLoadCA creates an X509 certificate from file
+func SetupTLSLoadCA(certPEM string) (*x509.CertPool, error) {
 	rootPEM, err := ioutil.ReadFile(certPEM)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to open root certificate '%s' : %s", certPEM, err)
--- a/config/config.go
+++ b/config/config.go
@@ -335,13 +335,22 @@ func GetEtcdServerAddrs() []string {
 	return []string{fmt.Sprintf("http://localhost:%d", port)}
 }
 
-// IsTLSEnabled returns true is the analyzer certificates are set
+// IsTLSEnabled returns true is the client / server certificates are set
 func IsTLSEnabled() bool {
-	certPEM := GetString("analyzer.X509_cert")
-	keyPEM := GetString("analyzer.X509_key")
-	if len(certPEM) > 0 && len(keyPEM) > 0 {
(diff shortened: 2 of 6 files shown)


nplanel:
Hi Maglo,


Thank you for your contribution.

Thanks to pin point us the inappropriate use of certificate trust in our implementation
Mainly due to the most common fault : a bit of code history, as the connection was only agents -> analyzer at the beginning.

Could you please accept our github invitation (email), aim to run our CI tests on this PR.

If you have 2min could you give us some clue how do you use/have interest in Skydive ?

maglo:
I’m currently not using it, but I’m considering it as a source of truth regarding networking. Also want to provide ad-hoc inspection capabilities to operators of private clouds or container platforms.

maglo:
I’m hoping that in the future Skydive will integrate better with physical network equipment via f.ex OSPF, Netflow etc.

nplanel:
@maglo good to known :)

We support LLDP discovery probe from #1344
OSPF and NetFlow/IPFix could be interesting, but NetFlow is pretty limited

You will may be interested by contrib/snort example, basically to map snort alerting on the topology

nplanel:
run python-tests

maglo:
Great work.

Thanks for the help with CI.

PR #1459, opened 2018-11-13
Adds NUMA topology details for host

Jaganathancse:
Displays NUMA topology details like RAM, CPU Cores, thread siblings
and NICs data for each NUMA node.

### DIFF
--- a/agent/host.go
+++ b/agent/host.go
@@ -28,12 +28,15 @@ import (
 	"os"
 	"strings"
 
+	"github.com/Jaganathancse/numautils"
+
 	"github.com/shirou/gopsutil/cpu"
 	"github.com/shirou/gopsutil/host"
 
 	"github.com/skydive-project/skydive/common"
 	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/graffiti/graph"
+	"github.com/skydive-project/skydive/logging"
 )
 
 // CPUInfo defines host information
@@ -50,6 +53,14 @@ type CPUInfo struct {
 	Mhz        int64  `json:"Mhz,omitempty"`
 	CacheSize  int64  `json:"CacheSize,omitempty"`
 	Microcode  string `json:"Microcode,omitempty"`
+}
+
+// NUMATopology defines NUMA topology details
+type NUMATopology struct {
+	NUMA int64
+	RAM  string               `json:"RAM,omitempty"`
+	NICs []string             `json:"NICs,omitempty"`
+	CPUs []*numautils.CPUInfo `json:"CPUs,omitempty"`
 }
 
 // createRootNode creates a graph.Node based on the host properties and aims to have an unique ID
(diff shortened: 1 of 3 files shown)


agent/host.go
```diff
@@ -28,6 +28,8 @@ import (
 	"os"
 	"strings"
 
+        "github.com/Jaganathancse/numautils"
```
safchain: You need to add a licence to your project

agent/host.go
```diff
@@ -107,6 +117,25 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
...
+                n := &NUMATopology{
+                        NUMA: numa.NUMA,
+                        RAM:  numa.RAM,
+                        NICs: numa.NICs,
+                        CPUs: numa.CPUs,
+                }
+                numaTopology = append(numaTopology, n)
+        }
+
+        m.SetField("NUMATopology", numaTopology)
```
safchain: I would prefer just "NUMA"

agent/host.go
```diff
@@ -107,6 +117,25 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
 
 	m.SetField("CPU", cpus)
 
+        // Gets NUMA topology details
+        numaInfo, err := numautils.GetNumaTopology()
```
safchain: could be nice to add functional tests so that we can ensure that it is working on multiple environment

agent/host.go
```diff
@@ -52,6 +54,14 @@ type CPUInfo struct {
 	Microcode  string `json:"Microcode,omitempty"`
 }
 
+// NUMATopology defines NUMA topology details
+type NUMATopology struct {
+        NUMA int64
+        RAM  string               `json:"RAM,omitempty"\`
```
safchain: not sure about the \\`

safchain:
run tests

agent/host.go
Jaganathancse: Currently numautils library supports only RHEL, fedora and centos, and also not tested in other env's.
So Planning to add metadata only if no error in function call numautils.GetNumaTopology().
This change also avoids the creating issue in skydive code for unsupported env's. is this fine?

agent/host.go
Jaganathancse: will update it.

agent/host.go
Jaganathancse: will remove and test it.

agent/host.go
lebauce: @Jaganathancse @safchain Our CI is based on CentOS and Fedora so I don't see any reason why we should not be able to add a test for it in this PR

safchain: right ! let's test what we can.

safchain: BTW we need to be sure that it will work with SELinux

lebauce:
run skydive-go-fmt

lebauce:
run skydive-go-fmt

lebauce:
Closing it as it hasn't moved in a while. Please reopen if necessary

PR #1471, opened 2018-11-20
k8s: fixed compilation of tests

hunchback:
broken due to changes done in topology/graph/graph.go

maybe we should consider re-adding k8s tests to list of defaults

### DIFF
--- a/topology/probes/istio/istio.go
+++ b/topology/probes/istio/istio.go
@@ -36,9 +36,12 @@ type Probe struct {
 	*k8s.Probe
 }
 
+type resourceHandler func(client *kiali.IstioClient, g *graph.Graph) k8s.Subprobe
+
 // NewIstioProbe creates the probe for tracking istio events
 func NewIstioProbe(g *graph.Graph) (*k8s.Probe, error) {
 	configFile := config.GetString("analyzer.topology.istio.config_file")
+	enabledSubprobes := config.GetStringSlice("analyzer.topology.istio.probes")
 	config, err := k8s.NewConfig(configFile)
 	if err != nil {
 		return nil, err
@@ -49,14 +52,16 @@ func NewIstioProbe(g *graph.Graph) (*k8s.Probe, error) {
 		return nil, err
 	}
 
-	subprobes := map[string]k8s.Subprobe{
-		"destinationrule":  newDestinationRuleProbe(client, g),
-		"gateway":          newGatewayProbe(client, g),
-		"quotaspec":        newQuotaSpecProbe(client, g),
-		"quotaspecbinding": newQuotaSpecBindingProbe(client, g),
-		"serviceentry":     newServiceEntryProbe(client, g),
-		"virtualservice":   newVirtualServiceProbe(client, g),
+	subprobeHandlers := map[string]k8s.SubprobeHandler{
+		"destinationrule":  newDestinationRuleProbe,
+		"gateway":          newGatewayProbe,
+		"quotaspec":        newQuotaSpecProbe,
+		"quotaspecbinding": newQuotaSpecBindingProbe,
+		"serviceentry":     newServiceEntryProbe,
+		"virtualservice":   newVirtualServiceProbe,
 	}
+
+	subprobes := k8s.InitSubprobes(enabledSubprobes, subprobeHandlers, client, g)
 
 	probe := k8s.NewProbe(g, Manager, subprobes, nil)
 
--- a/topology/probes/k8s/cluster.go
+++ b/topology/probes/k8s/cluster.go
@@ -34,38 +34,65 @@ const ClusterName = "cluster"
 
 var clusterEventHandler = graph.NewEventHandler(100)
 
+var clusterNode *graph.Node
+
 type clusterCache struct {
 	*graph.EventHandler
 	graph *graph.Graph
 }
 
-func (c *clusterCache) Start() {
+func (c *clusterCache) addClusterNode() {
 	c.graph.Lock()
 	defer c.graph.Unlock()
 
 	m := NewMetadata(Manager, "cluster", nil, ClusterName)
-	node := c.graph.NewNode(graph.GenID(), m, "")
-	c.NotifyEvent(graph.NodeAdded, node)
+	clusterNode = c.graph.NewNode(graph.GenID(), m, "")
+	c.NotifyEvent(graph.NodeAdded, clusterNode)
 	logging.GetLogger().Debugf("Added cluster{Name: %s}", ClusterName)
+}
+
+func (c *clusterCache) Start() {
 }
 
 func (c *clusterCache) Stop() {
 }
 
 func newClusterProbe(clientset interface{}, g *graph.Graph) Subprobe {
-	return &clusterCache{
+	c := &clusterCache{
 		EventHandler: clusterEventHandler,
 		graph:        g,
 	}
+	c.addClusterNode()
+	return c
+}
+
+type clusterLinker struct {
+	graph.DefaultLinker
+	*graph.ResourceLinker
+	g             *graph.Graph
+	objectIndexer *graph.MetadataIndexer
+}
+
+func (linker *clusterLinker) createEdge(cluster, object *graph.Node) *graph.Edge {
+	id := graph.GenID(string(cluster.ID), string(object.ID))
+	return linker.g.CreateEdge(id, cluster, object, topology.OwnershipMetadata(), graph.TimeUTC(), "")
+}
+
+// GetBALinks returns all the incoming links for a node
+func (linker *clusterLinker) GetBALinks(objectNode *graph.Node) (edges []*graph.Edge) {
+	edges = append(edges, linker.createEdge(clusterNode, objectNode))
+	return
 }
 
 func newClusterLinker(g *graph.Graph, manager string, types ...string) probe.Probe {
-	clusterIndexer := graph.NewMetadataIndexer(g, clusterEventHandler, graph.Metadata{"Manager": Manager, "Type": "cluster"})
-	clusterIndexer.Start()
-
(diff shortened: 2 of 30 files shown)


hunchback:
run skydive-k8s-tests

topology/probes/k8s/probe.go
```diff
@@ -94,6 +94,11 @@ func (p *Probe) AppendNamespaceLinkers(types ...string) {
 
 // NewProbe creates the probe for tracking k8s events
 func NewProbe(g *graph.Graph, manager string, subprobes map[string]Subprobe, linkers []probe.Probe) *Probe {
+	names := []string{}
+	for k, _ := range subprobes {
+		names = append(names, k)
+	}
+	logging.GetLogger().Infof("probe %s subprobes %v", manager, names)
```
lebauce: s/probe/Probe/

hunchback: ok

hunchback:
run skydive-k8s-tests

hunchback:
run skydive-k8s-tests

hunchback:
run skydive-k8s-tests

hunchback:
@lebauce - can you assist with k8s failure:

```
panic: test timed out after 20m0s
```

topology/graph/indexer.go
```diff
@@ -204,7 +204,7 @@ func NewMetadataIndexer(g *Graph, listenerHandler ListenerHandler, m ElementMatc
 		Indexer: NewIndexer(g, listenerHandler, func(n *Node) (kv map[string]interface{}) {
 			if match := n.MatchMetadata(m); match {
 				kv = make(map[string]interface{})
-				if values, err := getFieldsAsArray(n, indexes); err == nil && len(values) > 0 {
+				if values, err := getFieldsAsArray(n, indexes); err == nil {
```
lebauce: Why ???

lebauce:
Why do we have to change `resourceHandlers` and `EnabledSubprobes` again ?

lebauce:
@hunchback The test that timeouts is `TestK8sClusterNode` so the problem is probably related to the K8s probe (or this PR). I don't think this is a CI issue

hunchback:
so i can reuse by istio

hunchback:
the commit broke links between cluster to its children

hunchback:
will check ...

topology/probes/istio/istio.go
```diff
@@ -49,15 +52,22 @@ func NewIstioProbe(g *graph.Graph) (*k8s.Probe, error) {
...
+		"gateway":          newGatewayProbe,
+		"quotaspec":        newQuotaSpecProbe,
+		"quotaspecbinding": newQuotaSpecBindingProbe,
+		"serviceentry":     newServiceEntryProbe,
+		"virtualservice":   newVirtualServiceProbe,
+	}
+
+	subprobes := make(map[string]k8s.Subprobe)
+	for name, handler := range resourceHandlers {
+		subprobes[name] = handler(client, g)
```
lebauce: All the subprobes are created even if they are not enabled

hunchback:
run skydive-k8s-tests

hunchback:
@lebauce  - I think we may need to increase k8s execution time - later we can optimize and make the test run faster

topology/probes/istio/istio.go
hunchback: ok - have corrected

hunchback:
> @hunchback The test that timeouts is `TestK8sClusterNode` so the problem is probably related to the K8s probe (or this PR). I don't think this is a CI issue

this test PASSes when run locally on my devmode VM

hunchback:
run skydive-k8s-tests

hunchback:

lebauce:
run skydive-k8s-tests

topology/graph/indexer.go
lebauce: @hunchback That's why before 0b81a07 we had a special case when len(indexes) == 0. Anyway, I'm against removing && len(values) > 0 as it simply makes no sense to have an empty list (values) in kv.

hunchback:
but this breaks k8s which previously worked ...

hunchback:
so are you suggesting i replace the indexlinker and implement my custor
aablinks  balinks methods?

lebauce:
@hunchback Yes, but we were indexing all the objets that matched the filters,**even if they did not have the specified indices** . So basically all the nodes, with an empty values in the hashmap. Restoring the `case 0` you removed should fix Kubernetes

hunchback:
not sure i understand ... what is the fix you propose needed to get cluster
links working again ?

hunchback:
but this breaks cluster as indexer.Get() will not return nothing-nil
(rather than everything)

lebauce:
@hunchback Then yes. Creating a metadata indexer for a single node (cluster) makes little sense. Every time an object that needs to be linked to the cluster appears, we were querying the indexer that had a "nil" hash. That returned all the nodes : the cluster node in our case.

I think for the cluster linker, a classic linker that implements the GetBALinks() is simpler : we just return a link from the cluster node (that we can simply store as an attribute of the clusterLinker) to the node.

hunchback:
ok, will do

hunchback:
@lebauce - I want to propose the following:

(1) k8s is currently broken and the k8s tests disabled blocking k8s/istio PRs - I would be happy if this currently PR could be merged as-is.

(2) regarding your proposition - will your change not also break listing on the children nodes (namespaces, etc.)?

```
func (linker *clusterLinker) GetABLinks(clusterNode *graph.Node) (edges []*grpah.Edge) {
        nodes := linker.ObjectIndexer.Get()           <===== wi
        for _, n := nodes {
                // append to edges
        }
        return
}
```

lebauce:
@hunchback No need for GetABLinks as the cluster node already exists when the linker is started. You simply need to retrieve it and keep it as an attribute

hunchback:

hunchback:
run skydive-k8s-tests

topology/probes/k8s/cluster.go
```diff
@@ -27,47 +27,79 @@ import (
...
 }
 
-func newClusterProbe(clientset *kubernetes.Clientset, g *graph.Graph) Subprobe {
-	return &clusterCache{
+func newClusterProbe(clientset interface{}, g *graph.Graph) Subprobe {
+	c := &clusterCache{
 		EventHandler: clusterEventHandler,
 		graph:        g,
 	}
+	c.addClusterNode()
```
lebauce: This can be moved to the Start method

topology/probes/k8s/cluster.go
```diff
@@ -27,47 +27,79 @@ import (
...
+
+// GetBALinks returns all the incoming links for a node
+func (linker *clusterLinker) GetBALinks(objectNode *graph.Node) (edges []*graph.Edge) {
+	edges = append(edges, linker.createEdge(clusterNode, objectNode))
+	return
 }
 
 func newClusterLinker(g *graph.Graph, manager string, types ...string) probe.Probe {
-	clusterIndexer := graph.NewMetadataIndexer(g, clusterEventHandler, graph.Metadata{"Manager": Manager, "Type": "cluster"})
+	clusterIndexer := newObjectIndexerFromFilter(g, g, newTypesFilter(Manager, "cluster"))
```
lebauce: This doesn't seem to be used any more. This can be removed an nil be passed to NewResourceLinker instead

topology/probes/k8s/cluster.go
```diff
@@ -27,47 +27,79 @@ import (
...
+	objectIndexer *graph.MetadataIndexer
+}
+
+func (linker *clusterLinker) createEdge(cluster, object *graph.Node) *graph.Edge {
+	id := graph.GenID(string(cluster.ID), string(object.ID))
+	return linker.g.CreateEdge(id, cluster, object, topology.OwnershipMetadata(), graph.TimeUTC(), "")
+}
+
+// GetABLinks returns all the incoming links for a node
+func (linker *clusterLinker) GetABLinks(objectNode *graph.Node) (edges []*graph.Edge) {
```
lebauce: You can embed graph.DefaultListener in clusterLinker and remove the GetABLinks

topology/probes/k8s/cluster.go
hunchback: if moved to the start then we may miss handling of node Add events:
- as k8s.Probe Start() first starts the links
- then k8s.Probe.Start() Starts the subprobes in *random* order (looping via a go map works that way).

Thus if for example `namespace` will start before `cluster` then clusterLinker.GetBALinks() would get a `namespace` creation event *before* the singleton cluster node exists and thus would not be able to add the edge. As we are not implementing `clusterLinker.GetABLinks()` we have no chance to add the missing edge ....

topology/probes/k8s/cluster.go
hunchback: ok

topology/graph/indexer.go
```diff
@@ -204,7 +204,7 @@ func NewMetadataIndexer(g *Graph, listenerHandler ListenerHandler, m ElementMatc
 		Indexer: NewIndexer(g, listenerHandler, func(n *Node) (kv map[string]interface{}) {
 			if match := n.MatchMetadata(m); match {
 				kv = make(map[string]interface{})
-				if values, err := getFieldsAsArray(n, indexes); err == nil && len(values) > 0 {
+				if values, err := getFieldsAsArray(n, indexes); err == nil && len(indexes) == len(values) {
```
lebauce: This doesn't seem related to the compilation of the tests

hunchback: removed from the PR

lebauce:

lebauce:
run skydive-ppc64le-tests

PR #1472, opened 2018-11-21
ui: icon for libvirt

pierrecregut:
Libvirt VM image were using the default interface icon. Now, they have their
own icon.


pierrecregut:
This is a pretty dumb commit that should have been part of the libvirt probe. It just change the icon used by libvirt vms as the default interface icon is really misleading in this context. I just mixed two icons (host and the rounded square of interface) as my artistic skills are low.

sergeyglazyrindev:
Hello
It's very similar to host image:
https://github.com/skydive-project/skydive/blob/24725d53c927b65f0a9b2513df6eb304d871245c/statics/img/host.png
I think we need to use different image....

pierrecregut:
Yes and that's the point. It is similar because a vm is a virtualized host:

* the square around is often used for vm
*  the computer is a reminder that it is in some sense a host also especially when there is an agent in the guest.

The computer image could be reduced in size if there is a risk of confusion. Frankly I find them quite different.

lebauce:

pierrecregut:
Another option is two staggered rounded squares. But I don't know the origin of the convention and I don't find it intuitive.

sergeyglazyrindev:
Hello @pierrecregut 
I am sorry for the delay, I've got an issue with release, will try to propose an alternatives tomororow

sergeyglazyrindev-samsung:
hi, I am sorry, busy with release, will reply tomorrow or friday. Sorry for the delay...

sergeyglazyrindev-samsung:
What do you think about such icon ?
I'd like to use non black icon to make it more suitable for different backgrounds, etc
https://iconsplace.com/wp-content/uploads/_icons/ffa500/256/png/virtual-machine-icon-11-256.png

safchain:
Icon should not have background, and we need to be sure about copyright

sergeyglazyrindev-samsung:
Yes, its not hard to add a transparent background here but I asked about the idea....
And it seems like it is free icon. Google has shown this icon as free.

pierrecregut:
1. That is what I call "two staggered rounded squares", as I knew this option and decided against it because I found it confusing, I cannot say I favor it now. 
2. When icons are black, it is easy to use colors for status and other things without worrying about color conflicts.

sergeyglazyrindev-samsung:
okay, then maybe we can merge it as is. you know it's always hard to make decision regarding such things. It's mostly about understanding of art :) And there's no one truth :)

pierrecregut:
Finally it is very close to the Replicaset icon

pierrecregut:
Again if you think that the machine is too big, it can be easily adjusted by one or two pixels.

pierrecregut:
I could use one of the  machine of the "cluster" icon for example (after some cleanup because there is some white in it).

lebauce:
@sergeyglazyrindev-samsung @pierrecregut We would like to create a new Skydive release. An icon for the libvirt probe would be nice s I suggest we merge it as it. We can definitely change it later if we find a nicer one (I have no opinion on it :-) )

PR #1480, opened 2018-11-23
added sriov probe

sergeyglazyrindev-samsung:
responsibilites are:
1. take care of realtime changes in vf configuration
2. take care of realtime change for physical devices which provide sriov function
3. add all vfs currently configured on system
4. update state of the physical interface which provide sriov function
Partial-Bug: #SOFTID-200

Change-Id: I7736e052460ec83489c7360722d22e345f2d1960
Signed-off-by: Sergey Glazyrin <s.glazyrin@partner.samsung.com>

### DIFF
--- a/agent/probes.go
+++ b/agent/probes.go
@@ -41,6 +41,7 @@ import (
 	"github.com/skydive-project/skydive/topology/probes/ovsdb"
 	"github.com/skydive-project/skydive/topology/probes/runc"
 	"github.com/skydive-project/skydive/topology/probes/socketinfo"
+	"github.com/skydive-project/skydive/topology/probes/sriov"
 )
 
 // NewTopologyProbeBundleFromConfig creates a new topology probe.Bundle based on the configuration
@@ -121,6 +122,12 @@ func NewTopologyProbeBundleFromConfig(g *graph.Graph, hostNode *graph.Node) (*pr
 				return nil, fmt.Errorf("Failed to initialize runc probe: %s", err)
 			}
 			probes[t] = runc
+		case "sriov":
+			sriov, err := sriov.NewProbe(g, hostNode)
+			if err != nil {
+				return nil, fmt.Errorf("Failed to initialize Sriov probe: %s", err)
+			}
+			probes[t] = sriov
 		default:
 			logging.GetLogger().Errorf("unknown probe type %s", t)
 		}
--- a/topology/probes/sriov/sriov.go
+++ b/topology/probes/sriov/sriov.go
@@ -28,111 +28,82 @@ import (
 	"context"
 	"fmt"
 	"regexp"
-	"strconv"
 	"strings"
 	"time"
 
 	"github.com/vishvananda/netlink"
 
 	"github.com/skydive-project/skydive/common/fs"
-	"github.com/skydive-project/skydive/filters"
 	"github.com/skydive-project/skydive/graffiti/graph"
 	"github.com/skydive-project/skydive/logging"
-	"github.com/skydive-project/skydive/topology"
 )
 
 // Probe is a SRIO-V probe designed to handle all vfs on server
 type Probe struct {
 	graph   *graph.Graph
 	root    *graph.Node
-	cancel  context.CancelFunc // cancel function
+	cancel  context.CancelFunc
 	watcher *fs.SysProcFsWatcher
 }
 
-// Manager is a const to be used as a string to describe who handles sriov nodes
-const Manager = "vfio"
-
-// GenerateNameForNodeFromPhysInterfaceNameAndVFID - generates a name for vf based on physical device name and id of of this vf
-func GenerateNameForNodeFromPhysInterfaceNameAndVFID(physName string, id string) string {
-	return fmt.Sprintf("%s-vf-%s", physName, id)
-}
-
-func allZero(s []byte) bool {
-	for _, v := range s {
-		if v != 0 {
-			return false
-		}
-	}
-	return true
-}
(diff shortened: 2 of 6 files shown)


sergeyglazyrindev-samsung:
Now it is correct pull request but why I am unable to execute tests ?

safchain:
run tests

safchain:
you need to be a member of skydive-project, I'll send you an invitation

sergeyglazyrindev-samsung:
run tests

sergeyglazyrindev-samsung:
run skydive-k8s-tests

sergeyglazyrindev-samsung:
Hm, any idea why tests for k8s failed ?

lebauce:
@sergeyglazyrindev-samsung It's currently broken, it's been fixed

lebauce:
run skydive-packaging-tests

lebauce:

agent/probes.go
```diff
@@ -114,6 +115,12 @@ func NewTopologyProbeBundleFromConfig(g *graph.Graph, hostNode *graph.Node) (*pr
 				return nil, fmt.Errorf("Failed to initialize Libvirt probe: %s", err)
 			}
 			probes[t] = libvirt
+		case "sriov":
```
safchain: can you add this probe in the default config file (commented)

topology/probes/sriov/sriov.go
```diff
@@ -0,0 +1,177 @@
...
+					continue
+				}
+				h.graph.DelNode(node)
+			}
+			h.graph.Unlock()
+		}
+	}
+}
+
+func (h *Probe) addVfNode(name string, parentNode *graph.Node, vf *netlink.VfInfo) {
```
safchain: I think it would be better to add Metadata on the node created/seen by netlink instead of creating a new Node and I would create a link between the physical interface and the vfs one. I would add extra srio-v information in a SRIO-V section in the netlink vfs node metadata. Otherwise there is no way understand the relation between the vfs and the netlink interface. I could be also nice to add the PCI ID.

sergeyglazyrindev-samsung:
hi, I am busy with release, will reply tomorrow or friday.

topology/probes/sriov/sriov.go
sergeyglazyrindev-samsung: > I think it would be better to add Metadata on the node created/seen by netlink instead of creating a new Node and I would create a link between the physical interface and the vfs one. I would add extra srio-v information in a SRIO-V section in the netlink vfs node metadata. Otherwise there is no way understand the relation between the vfs and the netlink interface. I could be also nice to add the PCI ID.

I believe there's some misunderstanding, here we don't create a new node which provides vf stuff. We use a node created/seen by netlink probe and simply adding a vf node as child of specific physical device. Would you like to use layer2 connection instead of parent connection ?
for example, if there's a physical device which provides a network interface ens20 and it has vf functionality then we add a node with name: ens2f0-vf-VFID.
About PCI ID, I'll do that.
About updating metadata for node seen/created by netlink, I think it would be excessive information which we may get analyzing graphs.

lebauce: Let's say you have a SRIO-V card, device `enp5s0f0`. It will be seen by the netlink probe. If you create two VFs on this device. Will will end up with 2 interfaces : `enp5s16` and `enp5s16f2`. Those will also be seen by the `netlink` probe.

The probe also creates 2 VF child nodes. These are a bit of a duplicate of the nodes created by `netlink`. Instead, we suggest to not create the VF node and use the 2 nodes already created by `netlink`, and link them to the physical device `enp5s0f0`.

sergeyglazyrindev-samsung: > Let's say you have a SRIO-V card, device `enp5s0f0`. It will be seen by the netlink probe. If you create two VFs on this device. Will will end up with 2 interfaces : `enp5s16` and `enp5s16f2`. Those will also be seen by the `netlink` probe.
> 
> The probe also creates 2 VF child nodes. These are a bit of a duplicate of the nodes created by `netlink`. Instead, we suggest to not create the VF node and use the 2 nodes already created by `netlink`, and link them to the physical device `enp5s0f0`.

Oh, I was stupid. It seems like netlink probe handles a lot of things I added here.......
I'll check and let you know...... I am sorry :(

lebauce: It seems there is only 3 attributes for this VF nodes : `Name`, `Manager` and `Type`. If we don't create new nodes and used the netlink ones, I don't see any attribute to add to them. The real value is on the edges from the physical devices : we can use a dedicated `RelationType`, such as `sriov`.

sergeyglazyrindev-samsung: > It seems there is only 3 attributes for this VF nodes : `Name`, `Manager` and `Type`. If we don't create new nodes and used the netlink ones, I don't see any attribute to add to them. The real value is on the edges from the physical devices : we can use a dedicated `RelationType`, such as `sriov`.

yes, let me revisit it again cause you broke my mind :) I realized I added some duplication :)

sergeyglazyrindev-samsung: It's a shame on me :(

lebauce: @sergeyglazyrindev-samsung  :-) I did not mean to. Thanks for working on it. It worked nicely on my test machine by the way.

lebauce:

safchain:
Closing it as it has been addressed in others PR, feel free to summit patches on top on the current probe.

PR #1481, opened 2018-11-23
area: ui

sergeyglazyrindev-samsung:
Hello guys
I'll fix conflicts when we get an agreement about general idea.

Here is an updated gui for skydive
Few things, mostly minor need to be fixed and I hope to finalize these things on this week.
you may start review this changeset and in the end of week I'll ask you to start using this new gui
There's a safe way to try out this ui. Just add to the query string: use_hardcoded_data=1&newui_approach=1

An example: http://192.168.56.101:8082/topology?use_hardcoded_data=1&newui_approach=1
If you want to try it out on fresh topology, remove use_hardcoded_data=1 from query string

skip skydive-go-fmt
skip skydive-functional-tests-backend-elasticsearch
skip skydive-scale-tests
skip skydive-unit-tests
skip skydive-compile-tests
skip skydive-k8s-tests
skip skydive-ppc64le-tests
skip skydive-opencontrail-tests


sergeyglazyrindev-samsung:

sergeyglazyrindev-samsung:
run tests

statics/js/topologymanager_package/src/topologymanager/data_source/websocket.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    send(msg: any) {
+        this.conn.send(JSON.stringify(msg));
+    }
+
+    tryToReconnect() {
+        if (!this.reconnect) {
+            return;
+        }
+        this.conn.onopen = function() { }
```
safchain: I think there is an issue here. Once reconnected you will miss the events and then won't resync the graph

statics/js/topologymanager_package/src/topologymanager/topology_layout/infra/index.ts
```diff
@@ -0,0 +1,117 @@
...
+        this.uiBridge.remove();
+        $(this.selector).empty();
+        this.dataManager.removeOldData();
+        this.e.emit('ui.update');
+    }
+
+    addDataSource(dataSource: DataSourceI, defaultSource?: boolean) {
+        this.dataSources.addSource(dataSource, !!defaultSource);
+    }
+    reactToDataSourceEvent(dataSource: DataSourceI, eventName: string, ...args: Array<any>) {
```
safchain: I'm not sure about the datasource pattern. Finally layout and datasource are really tied and it seems that the datasource is more level of indirection for broadcasting websocket messages. Layout could just subscribe to a graphdatasource not using an interface that basically does nothing. There is no difference between all data sources except the gremlin filter.

sergeyglazyrindev-samsung:
hi, I am busy with release, will reply tomorrow or friday.

statics/js/topologymanager_package/src/topologymanager/data_source/base_skydive_data_source.ts
```diff
@@ -0,0 +1,56 @@
+import DataSourceI from "./interface";
+import * as events from 'events';
+import WsHandler from './websocket';
+
+export default class BaseSkydiveDataSource implements DataSourceI {
+    sourceType: string = "skydive";
+    dataSourceName: string = "base_skydive_data_source";
+    e: events.EventEmitter = new events.EventEmitter();
```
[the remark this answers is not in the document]
sergeyglazyrindev-samsung: > Relying on events and not callback makes to call harder to understand and to debug.
> 
> Or at least you could defines const that could be used.

I agree about that but it's a point for next refactoring. It would be small changes comparing to the current changeset size.

> There was already a subscriber mechanism, I do not see the point to use this EventEmitter
> https://github.com/skydive-project/skydive/blob/master/statics/js/websocket.js#L62
> 
> Other components could already subscribe to the graph to do something else, ex:
> https://github.com/skydive-project/skydive/blob/master/statics/js/components/rule-detail.js#L750

Such subscription mechanism breaks the one responsibilty SOLID principle. Each object which needs to have such mechanism needs to implement it on its own but it makes the code hard to reuse, etc
that's why we have an object EventEmitter in nodejs and it works fine in browser, and basically it is de-facto standart for nodejs.

statics/js/topologymanager_package/src/topologymanager/data_source/websocket.ts
sergeyglazyrindev-samsung: > I think there is an issue here. Once reconnected you will miss the events and then won't resync the graph

I don't like this part as well. But it works. About reconnecting -> I tested and it works. But yes you are right it is a suspicious part. Unfortunately it's impossible to make the code perfect :) So if it works, then it's ok :) And most important it is wrapped in one class. So, later we can easily fix it.

statics/js/topologymanager_package/src/topologymanager/topology_layout/infra/index.ts
sergeyglazyrindev-samsung: > I'm not sure about the datasource pattern. Finally layout and datasource are really tied and it seems that the datasource is more level of indirection for broadcasting websocket messages. 

Yes, but it's mostly about proper layering of the systems. It should be very easy to replace any communication gateway we use in our app. For example, right now I am not happy about websocket WsHandler I copied mostly from previous implementation. But I've no found better alternative. So, I think it is very good layer above websocket communication way to abstract away all websocket internal details of communication.

> Layout could just subscribe to a graphdatasource not using an interface that basically does nothing. There is no difference between all data sources except the gremlin filter.

You are right but it's very important to have an ability to use multiple data sources for layout. For instance, right now I copied link label implementation from previous javascript BUT there we use api to get an information about link latencies, etc
It's bad idea because presentation should know nothing about data sources it gets information from. So, later I'll create a separated data source based on api call to get a link latencies information to get rid of such improper system layering.

sergeyglazyrindev-samsung:
About infra layout. Right now its identical to another skydive default layout but there's already one difference, we set minimum collapse level for skydive default topology for host as 1. So the host gets expanded automatically when you switch to the host. so, there's already one difference and will be much more differences during product evolving.

statics/js/components/old_topology_approach.js
```diff
@@ -0,0 +1,3321 @@
+function WSHandler() {
```
safchain: I don't understand the purpose of merging two files into one big file. It seems to me that you could keep the original files, make modification if needed. It will make the review easier.

statics/js/topologymanager_package/src/topologymanager/topology_layout/skydive_default/index.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    addDataSource(dataSource: DataSourceI, defaultSource?: boolean) {
+        this.dataSources.addSource(dataSource, !!defaultSource);
+    }
+    reactToDataSourceEvent(dataSource: DataSourceI, eventName: string, ...args: Array<any>) {
+        console.log('Skydive default layout got an event', eventName, args);
+        let e;
+        switch (eventName) {
+            case "SyncReply":
+                if (this.config.getValue('useHardcodedData')) {
```
safchain: We should have hard coded topology. That is only for testing. There is already a way to inject topology to Skydive.

tests/selenium.go
```diff
@@ -554,6 +621,24 @@ func (s *seleniumHelper) stopVideoRecord() {
 	s.currVideoName = ""
 }
 
+func (s *seleniumHelper) switchToGremlinFilteredQueryTopology(topologyFilterQuery string) error {
```
safchain: not used

sergeyglazyrindev:
Okay, I've got an idea how to simplify it and allow to use old skydive version together with new version. Will explain tomorrow.

pierrecregut:
May be a first step could be to transition the full ui as it is today to webpack along the lines of this patch but using ES6 modules first (but still generate ES5 compatible code through babel). This would probably help to make dependencies more visible, we would immediately gain ES6 class syntax. Refactoring would be easier and the transition to typescript could then be progressive. Added bonus: you can use require on vuejs templates and store them in separate html files that editors understand.

Using modules should be only a matter of adding the right import/export

I would also recommend using typescript with the strictest settings for type-checking (at least no implicit any) from the start. Updating it later will mean changes that span over the whole front-end (or very complex compilation settings). Without them, the combination of gradual typing and type inference often mean 'any' type everywhere.

safchain:
I agree on this approach as we already have some part on the API in typescript and we do have a npm module for that in strict mode. Converting step by step seems to be a reasonable way. I'll submit what we have for the API(not vuejs) tomorrow as WIP/Draft.

statics/js/app.js
```diff
@@ -181,10 +179,18 @@ var app = new Vue({
...
+    globalEventHandler.onWebsocketEvent('websocket.connected', self.onConnected.bind(self));
+    globalEventHandler.onWebsocketEvent('websocket.disconnected', self.onDisconnected.bind(self));
+    globalEventHandler.onWebsocketEvent('websocket.error', self.onError.bind(self));
+    this.$store.subscribe(function(mutation) {
+      if (mutation.type === "login") {
+        globalEventHandler.websocket().toggleReconnectMode(true);
+	globalEventHandler.websocket().connect();
+      } else if (mutation.type === 'logout'){
+        globalEventHandler.websocket().toggleReconnectMode(false);
+	globalEventHandler.websocket().disconnect();
```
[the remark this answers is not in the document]
sergeyglazyrindev-samsung: as I said before, there are formatting issues (I wanted to fix formatting issues when we get an agreement about whole idea), but now it doesn't matter because I removed everything what I developed and added a simple way to initialize layouts, need to get agreement about code design and then we can discuss further steps.

statics/js/components/old_topology_approach.js
sergeyglazyrindev-samsung: okay

tests/selenium.go
sergeyglazyrindev-samsung: It was in use here
https://github.com/skydive-project/skydive/pull/1481/commits/36a1aa5453204330c8e4ecaedc7cfefacc354c5a#diff-afb2b0460e46d8b8ce3da9f8b974b27eR147

sergeyglazyrindev-samsung:
> I really think that this patch is too big, too many changes. I really think that we could change things step by step making this more acceptable. I'm pretty sure that there is a way to introduce a way to switch from one view/implementation seamlessly. I'll come with a proposal tomorrow.

> Moreover as said the host view doesn't bring a real value, we can already use filtering. We could simply add another vuejs component creating a host list. With this implementation we got disconnect every time we change the host. One more thing we can't do anymore cross host packet capture/node selection. So we need to keep both view at least. There is some problem with the design : there is scrollbar appearing while not needed and overlapping between the host list and the filtering (which is not working anymore).

Okay, I removed all my changes and added only a way to add new layouts.
Now there's a topology manager which has registry of topologies and we can activate needed layout using alias.
For instance, http://192.168.56.101:8082/topology?layout=test

About your another comments:
1. if we start skydive in really huge topology we get a lot of issues with big data in browser. Because we load full topology on page loading. My approach optimizes usage of skydive in huge topologies
2. You said "With this implementation we got disconnect every time we change the host.". And that's correct logic because we shouldn't load a topology of each host by default. I prefer to lazy load data. Maybe we may resubscribe with another gremlin query but it's another story. It's about adaptation of skydive to really huge topologies.
3.You said " One more thing we can't do anymore cross host packet capture/node selection. ".
We can. I implemented it using gremlin query: https://github.com/skydive-project/skydive/pull/1481/commits/36a1aa5453204330c8e4ecaedc7cfefacc354c5a#diff-afb2b0460e46d8b8ce3da9f8b974b27eR147
And it works. The idea is developer/operator needs to undestand what kind of thing he is investigating and for this we may use gremlin queries and if he wants to track/capture packets then he may filter the topology using specific query to be able to track packets on different hosts
4. I didn't say that it's ideal changeset. I proposed to merge and then step by step improve. I asked community to test my approach and find issues. What you found are serious issues but could be fixed quite fast.
But I agree with you it's better to improve things step by step.

sergeyglazyrindev-samsung:
Tests are still not working, I'll fix it when we understand how, what are the condition to have it merged, etc

sergeyglazyrindev-samsung:
And I am pretty sure in my old implementation not so many issues. We just need test it carefully, how it reacts to realtime changes, etc.

sergeyglazyrindev:
When I was talking about adaptation to huge topologies -> I meant it is done using infra layout which loads only necessary information for operator to be started.

sergeyglazyrindev:
But we may split it up into few phases:
1. provide a way to simply add a new layout.
2. adapt ui  to huge topologies, right now we load whole topology on page load.
3. etc

safchain:
@sergeyglazyrindev thanks I'll try to add a layout selector as suggested by our UX team to test your approach (seems to be good at first read).

sergeyglazyrindev:
fingers crossed.

lebauce:
Efforts moved to https://github.com/skydive-project/skydive-ui

PR #1485, opened 2018-11-27
ansible: bug - different websocket host_id

### LINKED ISSUE #1484: ansible: support multi-threading
When skydive_node.py or skydive_edge.py are used in an ansible playbook, it is necessary to set --fork=1

pierrecregut:
Generates a host_id based on the host parameter. Necessary to avoid
conflicts on the analyzer.

Adds some notes and fixes some minor problems in the online
documentation.

closes #1484

### DIFF
--- a/contrib/ansible/library/skydive_edge.py
+++ b/contrib/ansible/library/skydive_edge.py
@@ -39,7 +39,8 @@ description:
 options:
     analyzer:
         description:
-            - analyzer address, default: localhost:8082
+            - analyzer address
+        default: localhost:8082
         required: true
     ssl:
         description:
@@ -59,7 +60,7 @@ options:
         required: true
     relation_type:
         description:
-            - relation type of the node, ex: ownership, layer2, layer3
+            - "relation type of the node, ex: ownership, layer2, layer3"
         required: true
     node1:
         description:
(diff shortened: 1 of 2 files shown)


pierrecregut:
The commit solves the problem, but is not perfect. The missing point is that even with a bug, ansible should have failed but not hanged. The analyzer has sent an error message but it is swallowed somewhere.

safchain:
Correct, I think that is more a python skydive library issue than ansible module

pierrecregut:
Anyway, the fix is not enough even if it often works. There are still some conflicts sometimes during the second phase that enumerates the ports. My guess is that the loop gives work items to different workers with may be an affinity that hides the problem most of the time. 

Is there a good reason for not using a random id as X-Host-id ? A simple fix is to add os.getpid()

safchain:
No good reason getpid is fine

pierrecregut:
run skydive-unit-tests

pierrecregut:
unit tests are dependent on the availability of the mawi server in japan.

safchain:
rerun skydive-unit-tests

safchain:
rerun skydive-unit-tests

PR #1488, opened 2018-11-28
K8s: use stucts in-place of maps for node meta

hunchback:
NOTE: WIP as there is still a bug in k8s.Extra object serialization - @safchain / @lebauce  - can you check why this is happening? 

This PR replaced the K8s field which was a `map[string]interface{}` with a real structure. So as to support filters/gremlin I use a set of generic methods which traverse the structure and according to content of the `skydive` tag ("string", "int" or empty) implement the common.Getter interface.

Using this technique maybe more elegant that implementing specific `GetFieldString()` `GetFieldInt64()` etc. methods as for example is the case in `netlink/neighbor.go` so going forward methods I implemented in `k8s/graph.go` can be moved to `common`.

I assume we could consider later on relaying directly on the `skydive` tags thus enabling to cancel all-together the need to implement the `common.Getter` interface, but this can only be done once all searchable struct fields are tagged (using `skydive` tag) or if we agree to some convention as to which fields are exposed as `string` or `int64`.

In any case - code is ready for review.

### DIFF
--- a/analyzer/server.go
+++ b/analyzer/server.go
@@ -388,10 +388,10 @@ func init() {
 	// add decoders for specific metadata keys, this aims to keep the same
 	// object type between the agent and the analyzer
 	// Decoder will be used while unmarshal the metadata
-	graph.NodeMetadataDecoders["RoutingTables"] = netlink.RoutingTablesMetadataDecoder
-	graph.NodeMetadataDecoders["FDB"] = netlink.NeighborMetadataDecoder
-	graph.NodeMetadataDecoders["Neighbors"] = netlink.NeighborMetadataDecoder
-	graph.NodeMetadataDecoders["Metric"] = topology.InterfaceMetricMetadataDecoder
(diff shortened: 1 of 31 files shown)


lebauce:
@hunchback With a decoder on the Type, we are not protected against specifying conflict types for the same attribute (like `Ports`). I don't understand why we need such a change as a structure embedding other structure should work
```
type Metadata struct {
    *JobMetadata
    *ServiceMetadata
    ...
}
```

hunchback:
run skydive-k8s-tests

hunchback:

hunchback:
@safchain - do you want us to move forward with this PR? the last was that the k8s Extra data was munged  when viewed in the UI - indicating some the generic serialization of arbitrary objects was not working correctly.

hunchback:
the serialization of the K8s.Extra does not work properly:

image

hunchback:
@lebauce / @safchain - is ready for final review

PR #1495, opened 2018-11-30
ui: fix link between NS, stop simulation afte…

safchain:
…r 2sec

Disable link animation as it consumes lot of CPU.

### DIFF
--- a/statics/js/components/topology.js
+++ b/statics/js/components/topology.js
@@ -143,6 +143,24 @@ var TopologyComponent = {
                  :class="{\'glyphicon-resize-full\': currentNode.group.collapsed, \'glyphicon-resize-small\': !currentNode.group.collapsed}" aria-hidden="true"></span>\
             </span>\
           </button>\
+          <button id="pin-all" type="button" class="btn btn-primary" \
+                  title="Collapse" @click="pinAll">\
+            <span class="expand-icon-stack">\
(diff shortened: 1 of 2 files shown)


safchain:
@sergeyglazyrindev-samsung the goal of this was not to be merged. it is the result of some UI improvement that was needed for a demo. That is going to be used by the network service mesh project. I thought it would be nice to keep it as a WIP so that users can test if the result is better.

safchain:
run skydive-python-tests

statics/js/components/topology.js
```diff
@@ -143,6 +143,24 @@ var TopologyComponent = {
...
+          <button id="pin-all" type="button" class="btn btn-primary" \
+                  title="Collapse" @click="pinAll">\
+            <span class="expand-icon-stack">\
+              <i class="glyphicon glyphicon-pushpin icon-main"></i>\
+            </span>\
+            <span class="expand-icon-stack">\
+              <i class="glyphicon glyphicon-ok-circle icon-sub"></i>\
+            </span>\
+          </button>\
+          <button id="pin-all" type="button" class="btn btn-primary" \
```
matrohon: It should be unpin-all

matrohon:
I doesn't seem to work as I expected. With #1647, when I highlight g.E().HasKey('RelationType'), every node and edge is highlighted, but with this patch, only nodes are highlighted.

PR #1499, opened 2018-11-30
ui,openflow: quick fix of portnames

### LINKED ISSUE #1497: ui, openflow: port names are no more displayed in openflow rules tables
#1441 relies on the use of a subgraph. But this subgraph does not contain all the information used by ruledetail (we use interface linked to ovsports because they have the number OfPort metadata.)

pierrecregut:
The rule filtering functionality relies on a pre-filtering of the
graph. It breaked the display of port names in OpenFlow tables.

closes #1497

### DIFF
--- a/statics/js/components/rule-detail.js
+++ b/statics/js/components/rule-detail.js
@@ -431,7 +431,7 @@ var BridgeLayout = (function () {
     var nodes = this.interfaces[p];
     if (nodes === undefined) return '???';
     var port = nodes[1];
-    var portname = (port === undefined) ? '???' : port.metadata.Name;
+    var portname = (port === undefined) ? '???' : port.Metadata.Name;
     return portname;
   };
 
@@ -683,10 +683,11 @@ Vue.component('rule-detail', {
           return undefined
         },
         getTargets: function(node) {
+          var nodeid = node.id ? node.id : node.ID; // TODO: remove when capitalization is fixed.
           var targets = [];
           for (var i in this.edges) {
             var e = this.edges[i];
-            if (e.Parent === node.id) {
+            if (e.Parent === nodeid) {
               var n = this.getNode(e.Child);
               if (n) targets.push(n);
             }
@@ -788,7 +789,7 @@ Vue.component('rule-detail', {
       var self = this;
       console.log(this.filters);
       var queryBridge = "G.V('" + self.bridge.id + "').As('bridge')";
-      var queryPorts = queryBridge + ".Out().Has('Type', 'ovsport').As('ovsports')";
+      var queryPorts = queryBridge + ".Out().Has('Type', 'ovsport').As('ovsports').Out().As('ovsinterfaces')";
       var queryRules = queryBridge + ".Out().Has('Type', 'ofrule')";
       var has = "";
       var list = [];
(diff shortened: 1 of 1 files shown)


statics/js/components/rule-detail.js
```diff
@@ -683,10 +683,11 @@ Vue.component('rule-detail', {
           return undefined
         },
         getTargets: function(node) {
+          var nodeid = node.id ? node.id : node.ID;
```
sergeyglazyrindev-samsung: could you please add here some mention // @todo, get rid of this line when we refactor default skydive layout.
Right now we need to keep in mind such differences between different node representation: backend returns data with ID property but on frontend level we use id property. This is a thing which should be optimized in future.

pierrecregut:
This is a quick fix but I don't like the approach of passing a subgraph to rule-detail:

1. This is a sledgehammer to crack a nut.
2. It puts constraints on what rule-detail can do with the graph. 

Another problem is why on earth some fields (id, metadata) are lowercase and others are capitalized. My guess is that it is linked with the way the subgraph is handled.

sergeyglazyrindev-samsung:
> This is a quick fix but I don't like the approach of passing a subgraph to rule-detail:
> 
> 1. This is a sledgehammer to crack a nut.
> 2. It puts constraints on what rule-detail can do with the graph.
> 
> Another problem is why on earth some fields (id, metadata) are lowercase and others are capitalized. My guess is that it is linked with the way the subgraph is handled.

Yeah, I know. I added temporary bridge to handle this problem in this changeset:
https://github.com/skydive-project/skydive/pull/1481
But we need to comment the lines which should be doublechecked later. When we fix these bad things.

safchain:
So the current version may be not perfect, but at least we have a way to filter which was not there before. I think it brings a value for the user (except the bug OK). 
I don't see why using Subgraph doesn't make sense, it is there for that to get part of the graph according to filter. The only thing I see is we could use another websocket connection to really subscribe to a subgraph. But things are there to be improved.

lebauce:

pierrecregut:
There is another functionality that is broken by the subgraph approach: port highlighting no longer works.
I will submit another small patch.
What I really do not like about the approach is that the graph should be our unique intermediate between ui components, more precisely between actions requests and rendering. If we followed that, having alternative rendering will be just a matter of registering the right callbacks at the right place in the model.
But a subgraph obtained as it is is by an ajax query is just inert data. It no longer carry local model information such as, is this element highlighted, focused, etc.

pierrecregut:
"the graph" is the unique mutable data structure in which you can easily navigate between nodes and change their ui status. The idea of the gremlin filter is the right approach, but the result should only be an evolving set of rule UUID that are then used to filter the rule nodes really considered in BridgeLayout.extract and those nodes come from "the graph". @sergeyglazyrindev  The problem is not just a ducktyping problem, we want the real nodes in the subgraph if we are going to work with such structures. For the problem at hand, I don't think we neeed the subgraph. Followup in #1504

sergeyglazyrindev-samsung:
Hello, I agree with subgraph (you removed it recently in another changeset), but again, it is about keeping field names consistent with backend. So, my fix still has a sense because then we we get rid of this graph in raw javascript and migrate it to typescript, I prefer to have a proper field names, etc, so, such code would be more easier to migrate to the typescript skydive implementation.

PR #1504, opened 2018-12-03
ui: fix bugs in OpenFlow rule filters

pierrecregut:
* Use the real graph as BridgeLayout support structure.
* Fix filter computed for multiple values of a single field.

### DIFF
--- a/statics/js/components/rule-detail.js
+++ b/statics/js/components/rule-detail.js
@@ -254,12 +254,12 @@ function extractPort(c, graph, itfs) {
   var pot = graph.getTargets(c);
   for (var i = 0; i < pot.length; i++) {
     var cc = pot[i];
-    var ofport = cc.Metadata.OfPort;
+    var ofport = cc.metadata.OfPort;
     if (ofport === undefined)
       continue;
     var itf = [c, cc];
     itfs[ofport] = itf;
-    if (cc.Metadata.Type === 'patch') {
+    if (cc.metadata.Type === 'patch') {
       var ccc = graph.getNeighbor(cc, 'patch');
       if (ccc === undefined)
         return;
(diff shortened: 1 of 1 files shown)


pierrecregut:
Note that if the expressiveness of filters is kept as it is, it would be much faster to perform the filtering in the browser.

pierrecregut:
Note also that the use of regular expressions that I have not changed makes the analyzer vulnerable to ReDoS if you consider the user as hostile.

pierrecregut:

pierrecregut:

safchain:

lebauce:

statics/js/components/rule-detail.js
```diff
@@ -788,41 +754,25 @@ Vue.component('rule-detail', {
...
+      var keys = Object.keys(this.filters);
+      if(keys.length == 0) {
+        // Short cut no filter.
+        this.filtered = null;
+        this.memoBridgeLayout = null;
+        return;
       }
-      else{
-        query += ", 'ofrules'";
+      var queryRules = "G.V('" + self.bridge.id + "').Out().Has('Type', 'ofrule')";
```
sergeyglazyrindev-samsung: @safchain 
Hello, I do like this "gremlin query builder" implementation.
Is it possible to use it here ? to build a gremlin query ? Right now this component responsible for so many things...
https://github.com/skydive-project/skydive/blob/master/js/api.ts#L675

safchain: @sergeyglazyrindev-samsung It will be possible as it is part of the NPM module. I will push a draft of this package today as part of the skydive sources. But currently I think this implementation is OK. Just need to test it.

lebauce:

PR #1510, opened 2018-12-05
k8s: fix indexer needed for cluster linker

hunchback:
@lebauce - apparently the commit replacing `len(values) > 0` with `len(indexes) == len(values)` was needed - I did not retest after removing the commit as a result the cluster linker is now still broken even after merging: https://github.com/skydive-project/skydive/pull/1432 

please advise - I need an MetadataIndexer for cluster to object links which would capture   all graph Nodes from one of several types ("namespace", "node", etc.) - as currently len(values) > 0 - such a MetadataIndexer is not allowed -  I an reluctant to implement yest another variant of MetadataIndexer for handling the `len(indexes) == 0` case.


lebauce:
@hunchback Please check https://github.com/skydive-project/skydive/pull/1512

hunchback:
replaced by #1512

PR #1515, opened 2018-12-09
ui: remove k8s namespace filters

hunchback:
The current namespace filters implementation is broken:

- they don't include the namespace (and cluster) nodes
- they should come as an addition to existing filters (e.g. k8s service + k8s namespace <X>) not a stand-alone filter
- they should be implemented as checkbox - such that one can choose several namespaces

Given all the above this feature has been removed and will be re-implemented correctly is the future.

To test this setup configuration to include  `ui.k8s_enabled=true` and verify that no `k8s namespace ...` filters exists anymore.

### DIFF
--- a/statics/js/components/topology.js
+++ b/statics/js/components/topology.js
@@ -353,10 +353,6 @@ var TopologyComponent = {
         self.topologyFilterQuery();
       }
     });
-
-    if (self.isK8SEnabled()) {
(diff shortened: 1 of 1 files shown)


safchain:
rerun skydive-k8s-tests

safchain:
rerun skydive-k8s-tests

PR #1519, opened 2018-12-10
ansible: actions on list of elements

pierrecregut:
Provides actions on lists  of elements that are much faster
than repeating actions on one element at a time.

### DIFF
--- a/contrib/ansible/library/skydive_edge_list.py
+++ b/contrib/ansible/library/skydive_edge_list.py
@@ -1,0 +1,217 @@
+#
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+ANSIBLE_METADATA = {
+    'metadata_version': '1.1',
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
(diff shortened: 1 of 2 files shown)


contrib/ansible/library/skydive_node_list.py
```diff
@@ -0,0 +1,197 @@
...
+        finally:
+            self.factory.client.loop.call_soon(self.stop)
+
+
+def run_module():
+    module_args = dict(
+        analyzer=dict(type='str', default="127.0.0.1:8082"),
+        ssl=dict(type='bool', default=False),
+        insecure=dict(type='bool', default=False),
+        username=dict(type='str', default=""),
```
EmilienM: shouldn't username & passwords be required?

pierrecregut: no, if you use the noauth backend for skydive.

contrib/ansible/library/skydive_edge_list.py
```diff
@@ -0,0 +1,216 @@
...
+        description:
+            - host of the node
+
+notes:
+  - Using list reduces the communication overhead for creating edges.
+  - Node definitions must be UUIDs. Gremlin queries require a communication
+    roundtrip.
+
+author:
+    - Sylvain Afchain (@safchain)
```
safchain: You should maybe change this line :)

contrib/ansible/library/skydive_node_list.py
```diff
@@ -0,0 +1,197 @@
...
+        description:
+            - list of nodes
+        required: true
+
+notes:
+  - The default value of seed may be unsufficient to disambiguate nodes.
+  - Using list reduces the communication overhead for creating nodes.
+
+author:
+    - Sylvain Afchain (@safchain)
```
safchain: ditto

contrib/ansible/library/skydive_edge_list.py
pierrecregut: I will add a line as secondary author, my changes are minor compared to your code. This is one of the thing I do not like much about this code: there is a lot of boiler plate shared between modules. But it seems to be hard to share code between modules in ansible without putting constraints on the user (https://groups.google.com/forum/#!topic/ansible-devel/Tv_9GXnDJRI).

safchain: And it was a such excellent code :) Yes not easy to share code between modules it seems. Maybe we could add helpers into the python lib directly but that is something we can address later. Thanks for the PR.

safchain:
rerun skydive-k8s-tests

safchain:
rerun skydive-k8s-tests

PR #1523, opened 2018-12-11
Support for multiple interfaces of libvirt domain

JmilkFan:
Libvirt Domain has multiple interfaces, but Libvirt Topology Probes
function getDomainInterfaces can only get one of them. Due to the
overlapping of the memory addresses of the loop variable `itf`.

issue: 1522

### DIFF
--- a/topology/probes/libvirt/libvirt.go
+++ b/topology/probes/libvirt/libvirt.go
@@ -112,7 +112,8 @@ func (probe *Probe) getDomainInterfaces(
 	for _, itf := range d.Interfaces {
 		if constraint == "" || constraint == itf.Alias.Name {
 			itf.Host = domainNode
-			interfaces = append(interfaces, &itf)
+			itfObj := itf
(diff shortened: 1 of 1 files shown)


topology/probes/libvirt/libvirt.go
```diff
@@ -112,7 +112,8 @@ func (probe *Probe) getDomainInterfaces(
 	for _, itf := range d.Interfaces {
 		if constraint == "" || constraint == itf.Alias.Name {
 			itf.Host = domainNode
-			interfaces = append(interfaces, &itf)
+			itf_obj := itf
```
safchain: The golinter will raise an error because of "_" use. Could you please use itfObj instead, thanks

JmilkFan: Of course. This is a normative question, you are right.

JmilkFan:
Pls review again ：）

PR #1525, opened 2018-12-11
k8s/istio: small refactor to linkers initialization code

hunchback:
Now individual links can be disabled (as there is no why via Gremlin to do this) and this is required in various demo/productive scenarios.

To test you can disable the node2pod links by using:

```
analyzer:
  topology:
    k8s:
      linkers:
      - container2docker
      - pod2container
      - host2node
      #- node2pod
      - ingress2service
      - networkpolicy2pod
      - service2pod
```

### DIFF
--- a/topology/probes/istio/istio.go
+++ b/topology/probes/istio/istio.go
@@ -25,7 +25,6 @@ package istio
 import (
 	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/graffiti/graph"
-	"github.com/skydive-project/skydive/probe"
 	"github.com/skydive-project/skydive/topology/probes/k8s"
 
 	kiali "github.com/kiali/kiali/kubernetes"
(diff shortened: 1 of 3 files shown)


safchain:
Is this to workaround the WebUI ? If there is no need to have them, then we can remove them ? Why having them optional ? I think "demo" is not a valid reason. Production make sense but we need to understand the exact reason of this. The configuration is already quite complexe.

hunchback:
> Is this to workaround the WebUI ? If there is no need to have them, then we can remove them ? Why having them optional ? I think "demo" is not a valid reason. Production make sense but we need to understand the exact reason of this. The configuration is already quite complexe.

The feature I really want is to have links which don't create a gravity pull and thus don't distort the view ... as what is happening is that all the pods are pulled towards their owning node (thus making `pod --> networkpolicy --> pod` connections very difficult to decipher). The default (when nothing is specified will still work exactly as it works now) so the added complexity of configuration is only apparent when someone wants to use the new capability (of being able to disable specific links).

safchain:
It seems to be a WebUI issue and not a model issue. I don't think we should fix this by disabling/enabling links.

hunchback:
@safchain - ok have removed most of the PR - only left some minor improvements in initialization of linkers

hunchback:

hunchback:
@safchain - please process

PR #1532, opened 2018-12-17
Add Flow life cycle fields

orozery:
Per @eranra suggestion, we suggest to add the  following fields to the Flow struct:
1. `Status` - will be one of {started, updated, ended}. (`started` for the first time a flow is reported, `ended` for the last time it is reported (expired or otherwise detected as ended by TCP flags), and `updated` otherwise).
2. `FinishType` for the reason the flow has ended (will be omitempty for cases where flow has not yet ended). Possible values are currently timeout, tcp_rst, tcp_fin. This can be extended to other types.
3. `ReportCounter` counts the number of times a flow has been reported.

I have not yet wrote any unit testing.
Would first like to get your opinion on this: @lebauce @nplanel 

Thanks!

### DIFF
--- a/flow/flow.proto
+++ b/flow/flow.proto
@@ -191,10 +191,8 @@ message Flow {
 /* number of raw packet captured */
   int64 RawPacketsCaptured = 37;
 
-/* counts the number of times this flow was reported */
-  int64 ReportCounter = 60;
 /* describes the way the flow was ended (e.g. by RST, FIN) */
-  FlowFinishType FinishType = 61;
+  FlowFinishType FinishType = 60;
 }
 
 message FlowSet {
--- a/flow/flow.proto
+++ b/flow/flow.proto
@@ -56,9 +56,10 @@ message TransportLayer {
 }
 
 enum FlowFinishType {
-  TIMEOUT = 0;
-  TCP_FIN = 1;
-  TCP_RST = 2;
+  NOT_FINISHED = 0;
+  TIMEOUT = 1;
+  TCP_FIN = 2;
+  TCP_RST = 3;
 }
 
 enum ICMPType {
--- a/flow/flow.proto
+++ b/flow/flow.proto
@@ -55,6 +55,18 @@ message TransportLayer {
   int64 ID = 4;
 }
 
+enum FlowStatus {
+  STARTED = 0;
+  UPDATED = 1;
+  ENDED = 2;
+}
+
+enum FlowFinishType {
+  TIMEOUT = 0;
+  TCP_FIN = 1;
+  TCP_RST = 2;
+}
+
 enum ICMPType {
   UNKNOWN = 0;
   DESTINATION_UNREACHABLE = 1;
@@ -183,6 +195,13 @@ message Flow {
   repeated RawPacket LastRawPackets = 36;
 /* number of raw packet captured */
   int64 RawPacketsCaptured = 37;
+
+/* Flow life cycle status (Started, Updated, Ended) */
+  FlowStatus Status = 60;
+/* counts the number of times this flow was reported */
+  int64 ReportCounter = 61;
+/* describes the way the flow was ended (e.g. by RST, FIN) */
+  FlowFinishType FinishType = 62;
 }
 
 message FlowSet {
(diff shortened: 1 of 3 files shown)


flow/table.go
```diff
@@ -199,6 +199,9 @@ func (ft *Table) expire(expireBefore int64) {
 			}
 
 			logging.GetLogger().Debugf("Expire flow %s Duration %v", f.UUID, duration)
+			f.ReportCounter++
+			f.Status = FlowStatus_ENDED
```
hunchback: could flow have already ended (due to protocol flow termintation), if so then you would not want to handle as timeout

flow/table.go
```diff
@@ -248,10 +251,14 @@ func (ft *Table) update(updateFrom, updateTime int64) {
 	logging.GetLogger().Debugf("flow table update: %d, %d", updateFrom, updateTime)
 
 	var updatedFlows []*Flow
-	for _, f := range ft.table {
+	for k, f := range ft.table {
 		if f.XXX_state.updateVersion > ft.updateVersion {
 			ft.updateMetric(f, updateFrom, updateTime)
+			f.ReportCounter++
 			updatedFlows = append(updatedFlows, f)
+			if f.Status == FlowStatus_ENDED {
```
hunchback: if ended you may want to add to expiredFlows not updatedFlowd

flow/flow.proto
```diff
@@ -183,6 +195,13 @@ message Flow {
   repeated RawPacket LastRawPackets = 36;
 /* number of raw packet captured */
   int64 RawPacketsCaptured = 37;
+
+/* Flow life cycle status (Started, Updated, Ended) */
+  FlowStatus Status = 60;
```
safchain: I'm not sure about the approach since we are adding more field for all the flows to reduce the memory used for some of them. 

Adding fields here will have memory impact and storage impact. I would rather add a unique field here
https://github.com/skydive-project/skydive/blob/master/flow/flow.go#L68 which won't be serialized, let say `terminated bool`, and in the update function check this field to generate a list and then call `ft.expireHandler.callback(expiredFlows)`

orozery: Are you referring specifically to the field Status? or also to the other new fields FinishType & ReportCounter?

safchain: Just one field in this structure. This structure will be not stored in the datastore but adding a field still has cost in term of memory.

flow/table.go
```diff
@@ -248,10 +250,14 @@ func (ft *Table) update(updateFrom, updateTime int64) {
 	logging.GetLogger().Debugf("flow table update: %d, %d", updateFrom, updateTime)
 
 	var updatedFlows []*Flow
-	for _, f := range ft.table {
+	for k, f := range ft.table {
 		if f.XXX_state.updateVersion > ft.updateVersion {
 			ft.updateMetric(f, updateFrom, updateTime)
+			f.ReportCounter++
 			updatedFlows = append(updatedFlows, f)
+			if f.FinishType != "" {
```
safchain: that is an integer, I think the comparison if wrong

flow/flow.proto
```diff
@@ -183,6 +189,11 @@ message Flow {
   repeated RawPacket LastRawPackets = 36;
 /* number of raw packet captured */
   int64 RawPacketsCaptured = 37;
+
+/* counts the number of times this flow was reported */
+  int64 ReportCounter = 60;
```
safchain: I wouldn't keep this. If we want to know how many time a flow has been reported we can check the `_version` field of elasticsearch or even count the number of metric records. It seems to be more a debugging information ?

safchain:
ok to test

safchain:
run tests

flow/table.go
orozery: oops :) fixed (I hope)

flow/flow.proto
orozery: haha
https://github.com/skydive-project/skydive/pull/1532#discussion_r242832475
But OK, we'll ask @eranra about this

safchain:
rerun skydive-functional-tests-backend-elasticsearch

PR #1533, opened 2018-12-17
Adding OVS SFlow-Counter Information to Metadata

SH118415:
Co-authored-by: Darshan Vandra   <dvandra.ai@gmail.com>

### DIFF
--- a/flow/probes/ovssflow.go
+++ b/flow/probes/ovssflow.go
@@ -56,6 +56,7 @@ type OvsSFlowProbesHandler struct {
 	probes       map[string]OvsSFlowProbe
 	probesLock   common.RWMutex
 	Graph        *graph.Graph
+	Node         *graph.Node
 	fpta         *FlowProbeTableAllocator
 	ovsClient    *ovsdb.OvsClient
 	allocator    *sflow.AgentAllocator
@@ -178,7 +179,7 @@ func (o *OvsSFlowProbesHandler) UnregisterSFlowProbeFromBridge(bridgeUUID string
 }
 
 // RegisterProbeOnBridge registers a new probe on the OVS bridge
-func (o *OvsSFlowProbesHandler) RegisterProbeOnBridge(bridgeUUID string, tid string, capture *types.Capture) error {
+func (o *OvsSFlowProbesHandler) RegisterProbeOnBridge(bridgeUUID string, tid string, capture *types.Capture, n *graph.Node) error {
 	headerSize := flow.DefaultCaptureLength
 	if capture.HeaderSize != 0 {
 		headerSize = uint32(capture.HeaderSize)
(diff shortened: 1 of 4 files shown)


sflow/agent.go
```diff
@@ -89,8 +92,12 @@ func (sfa *Agent) feedFlowTable() {
 
 		// TODO use gopacket.NoCopy ? instead of gopacket.Default
 		p := gopacket.NewPacket(buf[:n], layers.LayerTypeSFlow, gopacket.DecodeOptions{NoCopy: true})
+		logging.GetLogger().Infof("value of gopacket %s", p)
```
safchain: It would be better to not log at all or at least to log using DEBUG otherwise in production it will generate a lot of noise

sflow/agent.go
```diff
@@ -103,7 +110,16 @@ func (sfa *Agent) feedFlowTable() {
 				// records each generating Packets.
 				sfa.FlowTable.FeedWithSFlowSample(&sample, bpf)
 			}
+			var counters []layers.SFlowCounterSample
+			for _, sample := range sflowPacket.CounterSamples {
```
safchain: this loop doesn't seem to be useful why not pushing sflowPacket.CounterSamples directly ? In the future we will need to define what we want to keep, not storing everything but only what make sense for us.

flow/probes/ovssflow.go
```diff
@@ -242,6 +243,7 @@ func (o *OvsSFlowProbesHandler) registerProbe(n *graph.Node, capture *types.Capt
 
 // RegisterProbe registers a probe on a graph node
 func (o *OvsSFlowProbesHandler) RegisterProbe(n *graph.Node, capture *types.Capture, e FlowProbeEventHandler) error {
+	o.Node = n
```
safchain: it doesn't seem to be required. as it is passed as argument of registerProbe

flow/probes/ovssflow.go
```diff
@@ -202,7 +203,7 @@ func (o *OvsSFlowProbesHandler) RegisterProbeOnBridge(bridgeUUID string, tid str
 	}
 
 	addr := common.ServiceAddress{Addr: address, Port: 0}
-	agent, err := o.allocator.Alloc(bridgeUUID, probe.flowTable, capture.BPFFilter, headerSize, &addr)
+	agent, err := o.allocator.Alloc(bridgeUUID, probe.flowTable, capture.BPFFilter, headerSize, &addr, o.Node, o.Graph)
```
safchain: `n` could be used, no need to use a member of OvsSFlowProbesHandler ?

sflow/agent.go
SH118415: Okay

sflow/agent.go
SH118415: Done

flow/probes/ovssflow.go
SH118415: Done

flow/probes/ovssflow.go
SH118415: Done

masco-enovance:
rerun skydive-cdd-overview-tests

PR #1550, opened 2018-12-23
gremlin: fix NE predicate returned element not have key

safchain:
Previously a node not have the key used by the NE predicate
was returned, like:

G.V().Has('AAAAAAAAAAAAAAAAAAAAAA', NE(""))

Now it ensures that the key exists before appliying the
NE predicate.


masco-enovance:
rerun skydive-unit-tests

graffiti/graph/traversal/traversal.go
```diff
@@ -122,18 +122,26 @@ func KeyValueToFilter(k string, v interface{}) (*filters.Filter, error) {
...
+			neFilter = filters.NewTermBoolFilter(k, !t)
 		default:
 			i, err := common.ToInt64(t)
 			if err != nil {
 				return nil, err
 			}
-			return filters.NewNotFilter(filters.NewTermInt64Filter(k, i)), nil
+			neFilter = filters.NewNotFilter(filters.NewTermInt64Filter(k, i))
 		}
+		if err != nil {
```
masco-enovance: I think this is not required since we are setting ```err``` only in default case and have a check there itself.

PR #1553, opened 2018-12-24
cleanup old variables in topology rules

SH118415:


masco-enovance:
run skydive-python-tests

PR #1566, opened 2019-01-06
netlink: Device speed is shown as 65535

jonnary:
This patch fix device speed 65535 to "unknown"

### DIFF
--- a/topology/probes/netlink/netlink.go
+++ b/topology/probes/netlink/netlink.go
@@ -401,6 +401,8 @@ func (u *NetNsProbe) addLinkToTopology(link netlink.Link) {
 	if speed, err := u.ethtool.CmdGet(&ethtool.EthtoolCmd{}, attrs.Name); err == nil {
 		if speed != math.MaxUint32 {
 			metadata["Speed"] = int64(speed)
+		}else {
+			metadata["Speed"] = "Unknown"
 		}
 	}
 
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -1954,10 +1954,10 @@
 			"revisionTime": "2015-01-06T09:31:45Z"
 		},
 		{
-			"checksumSHA1": "7Fqh2nZs8wQCD0AYGsqfLodWHAk=",
+			"checksumSHA1": "FleLGwYi5mNFHFUv43XcwB0NYcY=",
 			"path": "github.com/safchain/ethtool",
-			"revision": "6e3f4faa84e1d8d48afec75ed064cf3611d3f8bf",
-			"revisionTime": "2018-05-04T15:00:54Z"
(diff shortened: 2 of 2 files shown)


topology/probes/netlink/netlink.go
```diff
@@ -401,6 +401,8 @@ func (u *NetNsProbe) addLinkToTopology(link netlink.Link) {
 	if speed, err := u.ethtool.CmdGet(&ethtool.EthtoolCmd{}, attrs.Name); err == nil {
 		if speed != math.MaxUint32 {
 			metadata["Speed"] = int64(speed)
+		}else {
+			metadata["Speed"] = "Unknown"
```
nplanel: may not be a good idea to have a metadata of type int64 and string
I think the real fix would be to change the line to speed != math.MaxUint16   => need to double check ethtool package and ethtool API

jonnary: 1、ethtool result
[root@host_192_51_151_25 ~]# ethtool ens44f1
Settings for ens44f1:
Supported ports: [ FIBRE ]
Supported link modes: 1000baseT/Full
10000baseT/Full
Supported pause frame use: No
Supports auto-negotiation: Yes
Advertised link modes: 1000baseT/Full
10000baseT/Full
Advertised pause frame use: No
Advertised auto-negotiation: Yes
Speed: Unknown!
Duplex: Unknown! (255)
Port: Other
PHYAD: 0
Transceiver: external
Auto-negotiation: on
Supports Wake-on: umbg
Wake-on: g
Current message level: 0x00000007 (7)
drv probe link
Link detected: no
2、safchain/ethtool
pull a request: https://github.com/safchain/ethtool/pull/22
3、in skydive maybe we can changed as under?
                // TODO: it's better to fix math.MaxUint16 value in safchain/ethtool project
		if speed != math.MaxUint32 || speed != math.MaxUint16 {
			metadata["Speed"] = int64(speed)
		} else {
			metadata["Speed"] = "Unknown"
		}

safchain: Sorry I'm going to elaborate a bit. The datastore won't support multiple types for a same attribute, moreover we plan to convert metadata to protobuf which avoid multiple types as well. 

If the speed is unknown I propose to not report it.

jonnary:
OK.
As https://github.com/safchain/ethtool/pull/22 has merged, the  fix in skydive can be as follow?
1.upgrade skydive/vendor.json
2.If the speed is unknown we choice not report it or we report it int64(-1)?

safchain:
I would prefer to not report it

jonnary:
OK:)

lebauce:
Fixed by https://github.com/skydive-project/skydive/pull/1576

PR #1521, opened 2018-12-11
gremlin: Add function for without predicate

masco-enovance:

PR #1567, opened 2019-01-06
ui: add the 'k8s none' filter

hunchback:
when running with `ui.k8s_enabled: true` you should now see a new filter `k8s none` which shows all that is not k8s.


statics/js/components/topology.js
```diff
@@ -593,6 +597,10 @@ var TopologyComponent = {
       }
 
       if (self.isK8SEnabled()) {
+        // FIXME: Without() is not yet supported by backend so for now don't filter out istio
```
masco-enovance: without is added in backend from this patch https://github.com/skydive-project/skydive/pull/1521

PR #1572, opened 2019-01-08
gremlin: dedup, hash a node even if not having a key

safchain:
It is useful in case like following

Dedup('LastUpdateMetric.Start', 'SFlow.LastUpdateMetric.Start')

In such case some of the nodes have either LastUpdateMetric
and some other have SFlow.LastUpdateMetric.

### DIFF
--- a/graffiti/graph/traversal/traversal_test.go
+++ b/graffiti/graph/traversal/traversal_test.go
@@ -631,8 +631,8 @@ func TestTraversalParser(t *testing.T) {
 	// next traversal test
 	query = `G.V().Dedup("Type")`
 	res = execTraversalQuery(t, g, query)
-	if len(res.Values()) != 1 {
-		t.Fatalf("Should return 1 nodes, returned: %v", res.Values())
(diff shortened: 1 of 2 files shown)


graffiti/graph/traversal/traversal_test.go
```diff
@@ -803,3 +810,47 @@ func TestLimit(t *testing.T) {
...
+
+	// next traversal test
+	query = `G.V().Dedup("Name", "Type")`
+	res = execTraversalQuery(t, g, query)
+	if len(res.Values()) != 3 {
+		t.Fatalf("Should return 3 nodes, returned: %v", res.Values())
+	}
+
+	// next traversal test
+	query = `G.V().Dedup("Name", "Type")`
```
lebauce: It seems to be the same test than the previous one

safchain: Yes but I really want to be sure !

safchain:
rerun skydive-functional-tests-backend-elasticsearch

PR #1579, opened 2019-01-09
ci: create docker image supporting SKIP_BUILD, SKIP_PUSH, SKIP_MANIFEST

hunchback:
the PR improves create-docker-image.sh:

by enabling to skip various parts of the script by setting:
- SKIP_BUILD
- SKIP_PUSH
- SKIP_MANIFEST

by enabling to set DOCKER_TAG directly (rather than being derived from REF, TAG, VERSION)

an example test run:

```
$ cat >docker.env <<EOF
export SKIP_BUILD=
export SKIP_PUSH=
export SKIP_MANIFEST=
export ARCHES=amd64
export DOCKER_TAG=$(git rev-parse HEAD)
export DOCKER_IMAGE=skydive/skydive
export DOCKER_USERNAME=skydiveproject
export DOCKER_PASSWORD=place-correct-password-here
EOF

$ source docker.env

$ sudo -E scripts/ci/create-docker-image.sh
```


scripts/ci/create-docker-image.sh
```diff
@@ -14,75 +14,87 @@ set -e
 TAG=${REF##*/}
 VERSION=${TAG#v}
 
-[ -n "$VERSION" ] && DOCKER_TAG=$VERSION || DOCKER_TAG=latest
+: ${DOCKER_TAG:=$VERSION}
```
safchain: what about if VERSION is empty previously `latest` was used ?

hunchback: `REF` is set to "latest" if not set ... as a result `TAG` and `VERSION` are set to "latest"

hunchback:
@safchain - any further comments ?

scripts/ci/create-docker-image.sh
safchain: I could be useful to set it if we want to call it by hand

scripts/ci/create-docker-image.sh
```diff
@@ -21,68 +21,80 @@ if [ "${DOCKER_IMAGE%/*/*}" != "${DOCKER_IMAGE}" ]; then
...
-for arch in $ARCHES
-do
-    digest=$( docker inspect --format='{{index .RepoDigests 0}}' ${DOCKER_IMAGE}:${arch}-${DOCKER_TAG} )
-    docker manifest annotate --arch $arch "${DOCKER_IMAGE}:${DOCKER_TAG}" $digest
-done
+    docker manifest inspect "${DOCKER_IMAGE}:${DOCKER_TAG}"
+    docker manifest push --purge "${DOCKER_IMAGE}:${DOCKER_TAG}"
+}
+
+if [ -n "$DRY_RUN" ]; then
```
lebauce: The goal of DRY_RUN is to build the image but not push it. So this should be move after the `docker_build`

scripts/ci/create-docker-image.sh
hunchback: ok reverted

scripts/ci/create-docker-image.sh
hunchback: ok

lebauce:

PR #1586, opened 2019-01-10
istio: add link from virtualservice to service

tomerlange:
This PR creates link between virtualservice and service probes. 
It also extends Istio test such that now it checks this connection.


tomerlange:
run skydive-go-fmt

PR #1589, opened 2019-01-14
gremlin: allow to specify what field to use for Metrics step

safchain:


safchain:
@dvandra @SH118415 please have a look

PR #1594, opened 2019-01-15
lldp: prefer SysName and MgmtAddress to ChassisID for switch node id

lebauce:
Some switches - such as Cisco Nexus - sends a different chassis ID
for each port, so you use SysName and MgmtAddress if present and
fallback to chassis ID otherwise.

### DIFF
--- a/topology/probes/lldp/lldp.go
+++ b/topology/probes/lldp/lldp.go
@@ -344,7 +344,10 @@ func (p *Probe) getOrCreate(id graph.Identifier, m graph.Metadata) *graph.Node {
 // - when the interface is listed in the configuration file or we are in auto discovery mode
 func (p *Probe) handleNode(n *graph.Node) {
 	firstLayerType, _ := probes.GoPacketFirstLayerType(n)
-	mac, _ := n.GetFieldString("MAC")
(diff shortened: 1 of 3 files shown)


lebauce:

lebauce:

jrosser:
I've built this locally and installed the binaries to test, however no port 8082 is opened on the analyzer so I cannot visit the UI.

lebauce:
It seems only the embeded etcd is started. Could you please set the logging level to DEBUG ? https://github.com/skydive-project/skydive/blob/master/etc/skydive.yml.default#L348

lebauce:
It would also help if you could send the configuration file.

jrosser:
OK - so the binary starts and opens port 8082 on a bionic host but not on a xenial host.

jrosser:
The root of a lot of this is I had not appreciated I need to 'make static' in order to create a binary that would hopefully be portable from a bionic build host to a xenial target host. However 'make static' fails like this http://paste.openstack.org/show/743021/

lebauce:
@jrosser On Ubuntu, because of packaging issues, you need to do `make static WITH_OPENCONTRAIL=false`

PR #1601, opened 2019-01-17
Build: clean generated file

pierrecregut:
Also cleans the generated.proto file. Not doing so gives weird
behaviour when switching to old  branches.


lebauce:

PR #1607, opened 2019-01-20
Added ebpf support for TCP flag parsing

andrekassis:
Hi Guys
We enhance our eBPF parsing mechanism in this commit to allow skydive to track TCP flags (SYN,FIN,RST). It is worth mentioning that PCAP still offers a much wider functionality so there is a lot more to do but we think that this is the most pressing issue because it allows better tracking of the flows' life-cycle
Best,
Andre

### DIFF
--- a/topology/probes/netlink/netlink.go
+++ b/topology/probes/netlink/netlink.go
@@ -74,6 +74,7 @@ type NetNsProbe struct {
 	state                int64
 	wg                   sync.WaitGroup
 	quit                 chan bool
+	netNsNameTry         map[graph.Identifier]int
 }
 
 // Probe describes a list NetLink NameSpace probe to enhance the graph
@@ -355,7 +356,7 @@ func newInterfaceMetricsFromNetlink(link netlink.Link) *topology.InterfaceMetric
 	}
 }
 
-func (u *NetNsProbe) updateLinkNetNsName(intf *graph.Node, link netlink.Link, metadata graph.Metadata) {
+func (u *NetNsProbe) updateLinkNetNsName(intf *graph.Node, link netlink.Link, metadata graph.Metadata) bool {
 	var context *common.NetNSContext
 
 	lnsid := link.Attrs().NetNsID
@@ -392,10 +393,12 @@ func (u *NetNsProbe) updateLinkNetNsName(intf *graph.Node, link netlink.Link, me
 					metadata["LinkNetNsName"] = name
 				}
 
-				return
+				return true
 			}
 		}
 	}
+
+	return false
 }
 
 func (u *NetNsProbe) updateLinkNetNs(intf *graph.Node, link netlink.Link, metadata graph.Metadata) {
@@ -418,7 +421,22 @@ func (u *NetNsProbe) updateLinkNetNs(intf *graph.Node, link netlink.Link, metada
 		return
 	}
 
-	u.updateLinkNetNsName(intf, link, metadata)
+	const try = 3
+
+	i := u.netNsNameTry[intf.ID]
+	if i > try*3 {
+		// try again after a while
+		i = 0
+	}
+
+	if i < try {
+		if u.updateLinkNetNsName(intf, link, metadata) {
+			// avoid resolution if success
+			i = try - 1
+		}
+	}
+
+	u.netNsNameTry[intf.ID] = i + 1
 }
 
 func (u *NetNsProbe) addLinkToTopology(link netlink.Link) {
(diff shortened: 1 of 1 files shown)


lebauce:

probe/ebpf/flow.c
```diff
@@ -357,6 +366,27 @@ int bpf_flow_table(struct __sk_buff *skb)
 			__sync_fetch_and_add(&prev->rtt, tm - prev->start);
 			prev->_flags = RTT_SET;
 		}
+		if(prev->transport_layer.port_src==flow.transport_layer.port_src){
```
safchain: maybe it would be better to add all the tcp metrics related code in the update_metrics function ? you will have the ab/ba order and the timestamp is passed.

flow/probes/ebpf.go
```diff
@@ -156,6 +156,14 @@ func (p *EBPFProbe) flowFromEBPF(ebpfFlow *EBPFFlow, updatedAt int64) *flow.Flow
 				A:        portA,
 				B:        portB,
 			}
+			f.TCPMetric = &flow.TCPMetric{
+				ABSynStart: int64(kernFlow.transport_layer.ab_syn),
```
safchain: timestamp won't be correct regarding `start` and `last` of the flow because it is a monotonic clock. Please check line 338 to see how to get the real one.

andrekassis: 4) Can u please provide me with a few more details? What do you mean "a monotonic clock" and why won't this work..
Thx

probe/ebpf/flow.c
andrekassis: 3) Same

probe/ebpf/flow.c
```diff
@@ -357,6 +366,27 @@ int bpf_flow_table(struct __sk_buff *skb)
...
+		    if(prev->transport_layer.ab_syn==0){
+		        __sync_fetch_and_add(&prev->transport_layer.ab_syn, flow.transport_layer.ab_syn-prev->transport_layer.ab_syn);
+		    }
+		    if(prev->transport_layer.ab_fin==0){
+		        __sync_fetch_and_add(&prev->transport_layer.ab_fin, flow.transport_layer.ab_fin-prev->transport_layer.ab_fin);
```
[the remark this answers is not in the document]
andrekassis: 2)Same

probe/ebpf/flow.c
```diff
@@ -357,6 +366,27 @@ int bpf_flow_table(struct __sk_buff *skb)
 			__sync_fetch_and_add(&prev->rtt, tm - prev->start);
 			prev->_flags = RTT_SET;
 		}
+		if(prev->transport_layer.port_src==flow.transport_layer.port_src){
+		    if(prev->transport_layer.ab_syn==0){
+		        __sync_fetch_and_add(&prev->transport_layer.ab_syn, flow.transport_layer.ab_syn-prev->transport_layer.ab_syn);
```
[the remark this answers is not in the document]
andrekassis: Hi Sylvain,
Sorry I accidentally closed this PR. I will try to incorporate these changes and open it again. Meanwhile, I would appreciate it if you could answer a few questions regarding your comments:
1) Here I agree. Will change this

probe/ebpf/flow.c
andrekassis: Here I gave it some more thought and I don't think now that this statement is quite accurate. First the ab/ba order relates to the link layer while here we are concerned about the transport layer. Isn't this right?
+ I think there might be a bug in the current implementation so if you could help me clear a few things up I'd really appreciate it. Updatemetrics relies on the identities of the source and the destination in the link layer in order to determine the direction of the flow. but isn't this something that changes very frequently?
I mean if looking up a flow in the flow table relies on a 5-tuple representing the flow then the source and the destination in the link layer play no role in it. furthermore, if, say, A sends packets to B using different paths each time then this comparison in the link level would return false and the metrics would be updated incorrectly. Shouldn't we be using IPs instead?

safchain: The link layer is there in order to avoid conflict due to overlapping ips(tunneling). That said the ebpf implementation is really a naive version of the gopacket one. Tunneling is not supported and some other things. 

For the IPs vs Link there is a mode to select what are the fields to take into account. 
https://github.com/skydive-project/skydive/blob/master/flow/table.go#L56

I think we should pass the mode to the kernel side using this map https://github.com/skydive-project/skydive/blob/master/flow/probes/ebpf.go#L255

flow/probes/ebpf.go
safchain: The timestamp that you get from the kernel is different from the time that you get from userspace. It is the time since some unspecified starting point not the real time.

PR #1608, opened 2019-01-21
python:fix SSL context

SH118415:


SH118415:
run skydive-python-tests

PR #1609, opened 2019-01-21
[WIP] Adds openstack NFV parameters for host

Jaganathancse:
Displays openstack NFV OVS DPDK and SRIOV parameters.

### DIFF
--- a/agent/host_test.go
+++ b/agent/host_test.go
@@ -1,0 +1,42 @@
+// +build linux
+
+/*
+ * Copyright (C) 2018 Red Hat, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package agent
+
+import (
+        "github.com/Jaganathancse/openstack-nfv-params/openstacknfv"
+	"testing"
+)
+
+func TestOpenstackNfvParams(t *testing.T) {
+
+        numa, err := openstacknfv.GetOpenstackNfvParams()
+        if err != nil {
+                t.Fatal(err)
+        }
+
+        if numa == nil {
+                t.Fatalf("Unable to determine Openstack NFV parameters")
+        }
+}
--- a/agent/host.go
+++ b/agent/host.go
@@ -28,6 +28,8 @@ import (
 	"os"
 	"strings"
 
+        "github.com/Jaganathancse/openstack-nfv-params/openstacknfv"
+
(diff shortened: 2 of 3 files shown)


lebauce:
@Jaganathancse I have a few remarks on the approach: OpenStack configuration is pretty static and very much done at deployment time. In the agent configuration file, there is a `metadata` section where you can put parameters that will be directly be put in the metadata of the `host` node : https://github.com/skydive-project/skydive/blob/master/etc/skydive.yml.default#L237

As many parameters rely on the fact that OpenStack was deployed using TripleO (/etc/role.conf, /etc/hiera, ...), this could be done in https://github.com/openstack/tripleo-heat-templates/blob/master/extraconfig/services/skydive-analyzer.yaml where we have access to all of the parameters without having to read and parse files. By using the `skydive_extra_config` ansible variable, we can generate the appropriate `metadata` section.

Jaganathancse:
> @Jaganathancse I have a few remarks on the approach: OpenStack configuration is pretty static and very much done at deployment time. In the agent configuration file, there is a `metadata` section where you can put parameters that will be directly be put in the metadata of the `host` node : https://github.com/skydive-project/skydive/blob/master/etc/skydive.yml.default#L237
> 
> As many parameters rely on the fact that OpenStack was deployed using TripleO (/etc/role.conf, /etc/hiera, ...), this could be done in https://github.com/openstack/tripleo-heat-templates/blob/master/extraconfig/services/skydive-analyzer.yaml where we have access to all of the parameters without having to read and parse files. By using the `skydive_extra_config` ansible variable, we can generate the appropriate `metadata` section.

Openstack NFV parameters is not always static, if we use ease of deployment feature which generates ovs-dpdk and sriov parameters automatically and also operator can update the config values manually in post deployment. so i think, we need to code in agent host to get the live data for required parameters.

Jaganathancse:
> @Jaganathancse Does the operator can/allowed to change parameter by hand not updating the stack ?
> 
> With this approach, it depends on Tripleo deployment (I'm even not sure about the place of the files container/no container) and the files will be read only once, at the agent start which means that if the operator change something by hand he will have to restart the agents. I think that is really a deployment thing that should be handled by the appropriate tool/script (ansible, heat-template).

Yes, operator can change the parameters manually and restart the agents. currently no tool/script is available to update the config if any issues happened in post deployment and needs to do manually.

lebauce:
Your PR report different information :

- OVS config : this is already reported in the `Ovs.OtherConfig` section of bridge metadata
- Kernel command line : this is indeed useful and not only for OpenStack. I just submitted a PR that reports it in the metadata as a map, so that you can write query such as `G.V().Has("Type", "host", "KernelCmdLine.hugepages", true)`
- Isolated CPUs : this is already reported (in `agent/host.go`)
- /etc/role This is probably never modified after the deployment so this should be done in the heat templates (but passing arguments to our Ansible playbooks as we discussed above in the PR comments). Its probably the same for /etc/puppet/hieradata/service_names.json
- Nova parameters : a generic mechanism would be useful. @safchain just pushed a PR that makes use of the library we already for config parsing. It allows to specify a list of config files to parse.

agent/host_test.go
```diff
@@ -0,0 +1,42 @@
...
+package agent
+
+import (
+        "github.com/Jaganathancse/openstack-nfv-params/openstacknfv"
+	"testing"
+)
+
+func TestOpenstackNfvParams(t *testing.T) {
+
+        numa, err := openstacknfv.GetOpenstackNfvParams()
```
nplanel: I'm not against external library, but they should at least 

 1/ use mostly golang library/function
 2/ do not use shell command, as a host/program read/speak together with bytes/ABI/API ... not chars/string/quote   (I don't even talk about sudo call ...)
 3/ provide a simple API : not mixing kernel info, cpu isolations, nova info, ...
 4/ tests within Skydive should test (as much possible) your API call and implemented features

PR #1611, opened 2019-01-21
Istio: add link from destinationrule to service

tomerlange:
This PR creates link between destinationrule and service probes.
It also adds Istio test that checks this connection.

### DIFF
--- a/topology/probes/istio/destinationrule.go
+++ b/topology/probes/istio/destinationrule.go
@@ -22,7 +22,9 @@ import (
 
 	kiali "github.com/kiali/kiali/kubernetes"
 	"github.com/skydive-project/skydive/graffiti/graph"
+	"github.com/skydive-project/skydive/probe"
 	"github.com/skydive-project/skydive/topology/probes/k8s"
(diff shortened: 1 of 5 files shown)


topology/probes/istio/destinationrule.go
```diff
@@ -49,3 +52,24 @@ func (h *destinationRuleHandler) Dump(obj interface{}) string {
...
+
+type destinationRuleSpec struct {
+	App string `mapstructure:"host"`
+}
+
+func destinationRuleServiceAreLinked(a, b interface{}) bool {
+	dr := a.(*kiali.DestinationRule)
+	service := b.(*v1.Service)
+	drSpec := &destinationRuleSpec{}
+	if err := mapstructure.Decode(dr.Spec, drSpec); err != nil {
```
lebauce: Using mapstructure just to map a single attribute seems a bit overkill.

```
return dr.Spec["host"] == service.Labels["app"]
```
should be enough

tomerlange: DONE

Issue #1588, opened 2019-01-14
Customize Flows handled by subscribers


orozery:
We want to allow a general mechanism that flow subscribers will use that will allow them to:
1. Remove fields from the Flow structs being sent to the subscriber.
2. Filter flows being sent to the subscriber (e.g. subscriber will only get flows that have Flow.FinishType != NOT_FINISHED).

I see a couple of potential places where you could place this filter & struct transformation code:
1. On a specific subscriber, e.g. the object store subscriber. This will couple the mechanism to a specific subscriber, not allowing other subscribers to re-use it.
2. On the flow server, before calling SendFlows, apply transformation & filtering. Each subscriber should be able to customize the exact transformation & filtering. Can we make this part of the subscription protocol?
3. Write a general "transformation & filtering" subscriber that will apply those transformations, then send them (via a direct interface function call, not via websocket) to the "true" subscriber, such as object storage. The object store subscriber (or other subscribers that want to use this mechanism) should be re-implemented to match this interface, instead of subscribing to the flows server.

@eranra @safchain @andrekassis

eranra:
@orozery - thanks for adding this issue. I think that there is no doubt that from performance and efficiency point of view option 2. is most compelling but the flip of that is that this option limits the capabilities and complicates the subscription process .... 

@safchain @lebauce @nplanel waiting for your perspective on this issue

PR #1620, opened 2019-01-24
Add flow field exclusion in flow subscriber

orozery:
This PR addresses issue #1588.
It adds a general fieldexcluder package that allows removing fields (including subfields) from a general struct typed value (we use it to filter fields specifically from flow.Flow).
It is currently used only by the objectstore flow subscriber.
The objectstore flow subscriber was added with a new configuration parameters `excluded_fields`, to support this feature.

@eranra @safchain @andrekassis

### DIFF
--- a/contrib/objectstore/main.go
+++ b/contrib/objectstore/main.go
@@ -39,6 +39,12 @@ func main() {
 	subscriberUsername := cfg.GetString("subscriber_username")
 	subscriberPassword := cfg.GetString("subscriber_password")
 	maxSecondsPerStream := cfg.GetInt("max_seconds_per_stream")
+	excludedFieldsMap := cfg.GetStringMap("excluded_fields")
+
+	excludedFields := make([]string, 0, len(excludedFieldsMap))
+	for field := range excludedFieldsMap {
+		excludedFields = append(excludedFields, field)
+	}
 
 	authOptions := &shttp.AuthenticationOpts{
 		Username: subscriberUsername,
@@ -58,7 +64,7 @@ func main() {
 	}
 	structClient := wsClient.UpgradeToStructSpeaker()
 
-	s := subscriber.New(endpoint, region, bucket, accessKey, secretKey, objectPrefix, maxSecondsPerStream)
+	s := subscriber.New(endpoint, region, bucket, accessKey, secretKey, objectPrefix, maxSecondsPerStream, excludedFields)
 
 	// subscribe to the flow updates
 	structClient.AddStructMessageHandler(s, []string{"flow"})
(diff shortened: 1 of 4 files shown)


contrib/objectstore/main.go
```diff
@@ -39,6 +39,12 @@ func main() {
 	subscriberUsername := cfg.GetString("subscriber_username")
 	subscriberPassword := cfg.GetString("subscriber_password")
 	maxSecondsPerStream := cfg.GetInt("max_seconds_per_stream")
+	excludedFieldsMap := cfg.GetStringMap("excluded_fields")
```
safchain: GetStringSlice seems to be better

contrib/objectstore/subscriber/fieldexcluder/fieldexcluder.go
```diff
@@ -0,0 +1,61 @@
...
+	fieldFuncsMap := make(map[string]func(interface{}) map[string]interface{})
+	for field := range fieldMap {
+		excludedSubFields := matchExcludedFields(strings.ToLower(field), excludedFields)
+		if excludedSubFields != nil {
+			fieldFuncsMap[field] = generateExcludeFieldsFunction(fieldMap[field], excludedSubFields)
+		}
+	}
+
+	return func(val interface{}) map[string]interface{} {
+		v := reflect.ValueOf(val)
```
safchain: A general remark: usually reflect things are not good for performances. I think it would be better to write dedicated static marshallers.

orozery: @safchain do you mean that I should write an `if` clause for every possible field (including subfields) of `flow.Flow`?

contrib/objectstore/skydive-objectstore.yml.default
```diff
@@ -8,3 +8,22 @@
...
+    # TCPMetric:
+    # IPMetric:
+    # RTT:
+    # TrackingID:
+    # L3TrackingID:
+    # ParentUUID:
+    # NodeTID:
+    # LastRawPackets:
+    # RawPacketsCaptured:
+    # XXX_state:
```
safchain: this doesn't exists, it is not tranfered from agent to analyzer

contrib/objectstore/subscriber/fieldexcluder/fieldexcluder.go
safchain: yes It may be better for performances but I was more thinking about bunch of predefined marshaller like

```
MarshalOnlyLinkLayer(flow)
```

```
MarshalWithoutMetric(flow)
```

etc.

In the configuration you will specify the marshaller to use, so that no `if` at all. You specify the marshaller to use at the subscriber creation.

eranra: @orozery can we (maybe) tag the fields somehow and use the tags for the marsheler. Moving from hard coded names of marshelers to tags in the description of the structs?

safchain:
Can we close this one because of this one #1628  ?

orozery:
^^ yes

PR #1628, opened 2019-01-29
add objectstore marshaller

orozery:
This PR allows the object store flow subscriber to use a custom marshaller to transform the stored flows.
It includes one marshaller at the moment (named custom1) which is needed for a specific client using skydive.

This PR replaces the more generic (but perhaps less efficient) PR #1620.
As suggested here:
https://github.com/skydive-project/skydive/pull/1620#discussion_r250611550

@safchain @eranra

### DIFF
--- a/contrib/objectstore/main.go
+++ b/contrib/objectstore/main.go
@@ -9,6 +9,7 @@ import (
 	"github.com/skydive-project/skydive/common"
 	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/contrib/objectstore/subscriber"
+	"github.com/skydive-project/skydive/contrib/objectstore/subscriber/flowtransformer"
 	shttp "github.com/skydive-project/skydive/http"
 	"github.com/skydive-project/skydive/logging"
 	"github.com/skydive-project/skydive/websocket"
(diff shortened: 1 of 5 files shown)


contrib/objectstore/subscriber/marshaller/custom1/custom1.go
```diff
@@ -0,0 +1,42 @@
...
+	Last             int64
+	FinishType       flow.FlowFinishType
+}
+
+// Marshaller is a custom marshaller for flows
+type Marshaller struct {
+}
+
+// Marshal transforms a flow before being stored
+func (m *Marshaller) Marshal(f *flow.Flow) interface{} {
```
safchain: Marshaller/Marshal are probably not good names as it leads to think they are going to marshal while they don't in fact. What about FowTransformer ?

contrib/objectstore/subscriber/marshaller/marshaller.go
```diff
@@ -0,0 +1,25 @@
...
+	// Marshal transforms a flow before being stored
+	Marshal(flow *flow.Flow) interface{}
+}
+
+// New creates a new flow storage based on the backend
+func New(marshallerName string) (Marshaller, error) {
+	switch marshallerName {
+	case "custom1":
+		return custom1.New(), nil
+	case "":
```
safchain: do we need this case ?

orozery: We want the default to be no transformation at all.
The default is an empty string, so we use it to indicate no transformation.

safchain:
@orozery can you just rebase on top of master please, otherwise LGTM

PR #1631, opened 2019-01-29
agent: report kernel command line as host metadata

lebauce:


agent/host.go
```diff
@@ -129,5 +129,17 @@ func createRootNode(g *graph.Graph) (*graph.Node, error) {
 		m.SetField("VirtualizationRole", hostInfo.VirtualizationRole)
 	}
 
+	if cmdline, err := ioutil.ReadFile("/proc/cmdline"); err == nil {
+		kernelArgs := make(map[string]interface{})
+		for _, arg := range strings.Split(string(cmdline), " ") {
+			if splitted := strings.SplitN(arg, "=", 2); len(splitted) == 1 {
+				kernelArgs[splitted[0]] = true
```
nplanel: As we will store this map somewhere, would be better to have an fixed type, like map[string]string
I propose to set value to empty string like kernelArgs[] = ""

lebauce: What do you mean by "store this map somewhere" ?

nplanel: in the DB

lebauce: Not sure to understand what is the problem for the DB. It's always stored in the DB and it does work. So I would understand to use only strings for consistency, but I don't see the problem with the DB.

nplanel: yeah I agree for the DB, but consistency would be better as today in UI it's a little bit unusual to see "rw : true"
```

KernelCmdLine :
BOOT_IMAGE : /boot/vmlinuz-linux
btusb.enable_autosuspend : n
crashkernel : 128M
quiet : true
root : UUID=9474e5a2-0676-4090-be7d-7fb6cb1e879b
rw : true
```

safchain: In fact interface Features are already using boolean, I think it is more consistent to use boolean

nplanel: yeah but the result is quite unusual / ugly as "rw" is just a flag
User could interpret as  " quiet=true rw=true root=UUID=...."

nplanel:

PR #1635, opened 2019-01-30
ci: add mem profiling during functionals run

safchain:
skip skydive-compile-tests
skip skydive-go-fmt
skip skydive-k8s-tests
skip skydive-scale-tests
skip skydive-selenium-tests
skip skydive-unit-tests
skip skydive-cdd-overview-tests
skip skydive-ppc64-tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,9 @@ ifeq ($(COVERAGE), true)
   TEST_COVERPROFILE?=../functionals.cover
   EXTRA_ARGS+=-test.coverprofile=${TEST_COVERPROFILE}
 endif
+ifeq ($(WITH_PROF), true)
+  EXTRA_ARGS+=-profile
+endif
 TIMEOUT?=1m
 TEST_PATTERN?=
 UT_PACKAGES?=$(shell $(GOVENDOR) list -no-status +local | grep -Ev '/tests|/contrib')
--- a/common/profile.go
+++ b/common/profile.go
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2018 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy ofthe License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specificlanguage governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package common
+
+import (
+	"log"
+	"os"
+	"os/signal"
+	"runtime"
+	"runtime/pprof"
+	"syscall"
+)
+
+// Profile start profiling loop
+func Profile() {
+	cpu := make(chan os.Signal, 1)
+	signal.Notify(cpu, syscall.SIGUSR1)
+
+	memory := make(chan os.Signal, 1)
+	signal.Notify(memory, syscall.SIGUSR2)
+
+	for {
+		select {
+		case <-cpu:
+			f, err := os.Create("/tmp/skydive-cpu.prof")
+			if err != nil {
+				log.Fatal("could not create CPU profile: ", err)
+			}
+			if err := pprof.StartCPUProfile(f); err != nil {
+				log.Fatal("could not start CPU profile: ", err)
+			}
+			log.Println("CPU profiling Started")
+
+			<-cpu
+
+			pprof.StopCPUProfile()
+			f.Close()
+			log.Println("CPU profiling Stopped")
+		case <-memory:
+			// Memory Profile
+			runtime.GC()
+			memProfile, err := os.Create("/tmp/skydive-memory.prof")
+			if err != nil {
+				log.Fatal(err)
+			}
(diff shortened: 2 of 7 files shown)


safchain:

common/profile.go
```diff
@@ -0,0 +1,67 @@
...
+	memory := make(chan os.Signal, 1)
+	signal.Notify(memory, syscall.SIGUSR2)
+
+	for {
+		select {
+		case <-cpu:
+			f, err := os.Create("/tmp/skydive-cpu.prof")
+			if err != nil {
+				log.Fatal("could not create CPU profile: ", err)
+			}
```
hunchback: Missing:

```
defer f.Close()
```

common/profile.go
```diff
@@ -0,0 +1,67 @@
...
+	cpu := make(chan os.Signal, 1)
+	signal.Notify(cpu, syscall.SIGUSR1)
+
+	memory := make(chan os.Signal, 1)
+	signal.Notify(memory, syscall.SIGUSR2)
+
+	for {
+		select {
+		case <-cpu:
+			f, err := os.Create("/tmp/skydive-cpu.prof")
```
hunchback: rather use:
```
f, err := ioutil.TempFile("", "skydive-cpu.prof")
```

common/profile.go
```diff
@@ -0,0 +1,67 @@
...
+			log.Println("CPU profiling Started")
+
+			<-cpu
+
+			pprof.StopCPUProfile()
+			log.Println("CPU profiling Stopped")
+		case <-memory:
+			// Memory Profile
+			runtime.GC()
+			memProfile, err := os.Create("/tmp/skydive-memory.prof")
```
hunchback: rather use: `ioutilTempFile`

common/profile.go
```diff
@@ -0,0 +1,67 @@
...
+			f, err := os.Create("/tmp/skydive-cpu.prof")
+			if err != nil {
+				log.Fatal("could not create CPU profile: ", err)
+			}
+			if err := pprof.StartCPUProfile(f); err != nil {
+				log.Fatal("could not start CPU profile: ", err)
+			}
+			log.Println("CPU profiling Started")
+
+			<-cpu
```
hunchback: why is this needed?

common/profile.go
```diff
@@ -0,0 +1,67 @@
...
+
+			pprof.StopCPUProfile()
+			log.Println("CPU profiling Stopped")
+		case <-memory:
+			// Memory Profile
+			runtime.GC()
```
hunchback: this means sending a SIGUSR2 will change the behavior of the application, and the memory dump may not be representative of actual memory typically consumed; if calling `runtime.GC()` explicitly has a favorable effect on the application then it should be invoked periodically by the application regardless of SIGUSR2.

scripts/ci/cleanup.sh
```diff
@@ -69,6 +69,8 @@ EOF
   rm -rf /tmp/skydive_agent* /tmp/skydive-etcd
   rm -rf /var/lib/jenkins/.vagrant.d/tmp
 
+  rm /tmp/skydive-memory.prof*
```
hunchback: rather replace by:

```
rm -rf /tmp/skydive-memory.prof*
rm -rf /tmp/skydive-cpu.prof*
```

scripts/ci/run-tests-utils.sh
```diff
@@ -6,6 +6,16 @@ network_setup() {
         done
 }
 
+mem_prof() {
+        echo start memory profiling
+        while(true); do
+                echo trigger memory profiling snapshot
+                sudo pkill -USR2 functionals 2>/dev/null || true
+                sleep 10
+                sudo mv /tmp/skydive-memory.prof /tmp/skydive-memory.prof.$( date +%T ) 2>/dev/null || true
```
hunchback: if you create unique file names using `ioutil.TempFile` you can remove this line

hunchback:
@safchain - you probably want to run some checks on the output and *FAIL* the test if cpu/memory is more than should be - alternatively you can simply print the content of these files into the test execution log so that it is kept for later reference.

scripts/ci/run-tests-utils.sh
nplanel: no, he would like easily keep and found the memory profiling in time order (+%T)

hunchback: ok

common/profile.go
safchain: The goal here is to see real memory allocation in order to detect memory leak or unexpected memory allocation. If I do not call the GC I will get a lot of noise. In the future we could import this to select what we want to check.

common/profile.go
safchain: waiting a second kill to stop the profiling.

safchain:
@hunchback add check on the CI for the kind of thing is a tricky thing that should be address in another PR. The goal here is to add helper to help profiling and troubleshooting

common/profile.go
safchain: good catch: missing a close but not a defer

safchain:

safchain:

Makefile
```diff
@@ -75,6 +75,9 @@ ifeq ($(COVERAGE), true)
   TEST_COVERPROFILE?=../functionals.cover
   EXTRA_ARGS+=-test.coverprofile=${TEST_COVERPROFILE}
 endif
+ifeq ($(WITH_PROF), true)
+  EXTRA_ARGS+= -profile
```
lebauce: nit: remove leading space

PR #1637, opened 2019-01-31
ovs: allow user to disable ovs metrics via config

masco-enovance:


topology/probes/ovsdb/ovsdb.go
```diff
@@ -402,7 +402,7 @@ func (o *Probe) OnOvsInterfaceAdd(monitor *ovsdb.OvsMonitor, uuid string, row *l
 		}
 	}
 
-	if field, ok := row.New.Fields["statistics"]; ok {
+	if field, ok := row.New.Fields["statistics"]; ok && !config.GetBool("ovs.disable_metrics") {
```
safchain: I would prefer to not access to the configuration within the probe. You could add a bool as member of the probe definiton, setting it in the NewProbeFromConfig function ?

I would rather call the config option name `disable_stats` as it is an ovs option and it is called statistics in the ovs world.

safchain:

safchain:
rerun skydive-cdd-overview-tests

PR #1638, opened 2019-01-31
capture: broadcast capture update when capture.state change

safchain:


safchain:
run skydive-scale-tests

PR #1639, opened 2019-01-31
customize flow timeout by application

orozery:
We want to allow setting custom expire timeout for flow by their Application field.
For example, you may want to set DNS flows to timeout faster than the default 10 minutes.

I haven't yet tested this, but would like to get your feedback on the idea & implementation approach.
Thanks!

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -458,6 +458,12 @@ flow:
     udp:
       # 1194: OPENVPN
 
+  # application specific flow timeout, in seconds
+  # this timeout is enforced in addition to the general flow.expire timeout
+  application_timeout:
+    # - arp: 10
+    # - dns: 10
+
 ui:
   # Specify the extra assets folder. Javascript and CSS files present in this
   # folder will be added to the WebUI.
--- a/flow/table.go
+++ b/flow/table.go
@@ -18,6 +18,7 @@
 package flow
 
 import (
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -26,6 +27,7 @@ import (
 	"github.com/google/gopacket/layers"
 
 	"github.com/skydive-project/skydive/common"
+	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/filters"
 	"github.com/skydive-project/skydive/logging"
 )
@@ -81,6 +83,7 @@ type Table struct {
 	tcpAssembler      *TCPAssembler
 	flowOpts          Opts
 	appPortMap        *ApplicationPortMap
+	appTimeout        map[string]int64
 }
 
 // OperationType operation type of a Flow in a flow table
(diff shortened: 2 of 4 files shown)


hunchback:
@orozery - i suggest you add a default configuration with common application timeouts (such as dns).

orozery:
@safchain your thoughts?

safchain:
@orozery I need to check what's the performance impact of this

flow/table.go
```diff
@@ -107,6 +110,11 @@ type Operation struct {
 
 // NewTable creates a new flow table
 func NewTable(updateHandler *Handler, expireHandler *Handler, nodeTID string, opts ...TableOpts) *Table {
+	appTimeout := make(map[string]time.Duration)
+	for key := range config.GetConfig().GetStringMap("flow.table.app_timeout") {
+		appTimeout[strings.ToLower(key)] = time.Second * time.Duration(config.GetConfig().GetFloat64("flow.table.app_timeout." + key))
```
safchain: I would suggest to use upper case as it is like this for all the application and then no need to lower flow.Application in the update loop

safchain: Please provide an example in the default configuration file: etc/skydive.yml.default

flow/table.go
```diff
@@ -249,6 +258,7 @@ func (ft *Table) updateMetric(f *Flow, start, last int64) {
 func (ft *Table) update(updateFrom, updateTime int64) {
 	logging.GetLogger().Debugf("flow table update: %d, %d", updateFrom, updateTime)
 
+	now := time.Now()
```
safchain: please use updateTime instead of time.Now()

flow/table.go
```diff
@@ -257,6 +267,10 @@ func (ft *Table) update(updateFrom, updateTime int64) {
 			if f.FinishType != FlowFinishType_NOT_FINISHED {
 				delete(ft.table, k)
 			}
+		} else if now.Sub(time.Unix(0, f.Last * int64(time.Millisecond))) > ft.appTimeout[strings.ToLower(f.Application)] && ft.appTimeout[strings.ToLower(f.Application)] > 0 {
```
safchain: See my upper comment, if you set the map with upper case application no need to call ToLower here

orozery:
@safchain I updated the PR according to your comments

etc/skydive.yml.default
```diff
@@ -163,6 +163,14 @@ analyzer:
   replication:
     # debug: false
 
+  table:
```
safchain: I would prefer to use the flow section that is already existing:
https://github.com/skydive-project/skydive/blob/97921a7961f0bcead5d6fded82ce97cb501c7d2e/etc/skydive.yml.default#L445

the sub section could be `application_timeout` as we already have `application_ports`.

hunchback:
Please consider adding a unit test

orozery:
@hunchback added a unit test, and changed the default timeout for ARP/DSN to 10 seconds, as you suggested.
@safchain waiting for your performance evaluation & final review.
Thanks!

PR #1641, opened 2019-02-03
added DNS fields

andrekassis:
We add to extra fields for DNS flows to represent the query and the answer. The extra data we've added so far is not enough as it only displays the query and the answer, ignoring the authorities and other important fields. We would like to know what other fields you think are worth displaying.

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -905,6 +905,20 @@ func (f *Flow) newApplicationLayer(packet *Packet, opts Opts) error {
 				NSCount: d.NSCount,
 				QDCount: d.QDCount,
 			}
+			if len(d.Questions) > 0 {
+				questions := make([]string, len(d.Questions))
+				for i, v := range d.Questions {
+					questions[i] = string(v.Name)
+				}
+				f.DNS.DNSQuestions = questions
+			}
+			if len(d.Answers) > 0 {
+				answers := make([]string, len(d.Answers))
+				for i, v := range d.Answers {
+					answers[i] = string(v.IP)
+				}
+				f.DNS.DNSAnswers = answers
+			}
 			return nil
 		}
 	}
--- a/flow/layers/layers.go
+++ b/flow/layers/layers.go
@@ -29,6 +29,8 @@ type LayerDHCPv4 struct {
 //proteus:generate
 type LayerDNS struct {
 	*layers.DNS
+	DNSQuestions []string
+	DNSAnswers   []string
 }
 
 // LayerVRRPv2 wrapper to generate extra layer


flow/flow.go
```diff
@@ -893,17 +893,31 @@ func (f *Flow) newApplicationLayer(packet *Packet, opts Opts) error {
 	if (opts.ExtraLayers & DNSLayer) != 0 {
 		if layer := packet.Layer(layers.LayerTypeDNS); layer != nil {
 			d := layer.(*layers.DNS)
+			Answers := make([]string, 0, 10)
```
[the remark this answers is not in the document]
masco-enovance: can you change the variable name ```Answers``` to ```answers```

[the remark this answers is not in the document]

flow/flow.go
```diff
@@ -893,17 +893,31 @@ func (f *Flow) newApplicationLayer(packet *Packet, opts Opts) error {
...
+			Questions := make([]string, 0, 10)
```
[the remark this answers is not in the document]

flow/flow.go
```diff
@@ -893,17 +893,31 @@ func (f *Flow) newApplicationLayer(packet *Packet, opts Opts) error {
...
+			if d.Answers != nil {
+				for _, v := range d.Answers {
+					Answers = append(Answers, string(v.IP))
+				}
+			}
+			if d.Questions != nil {
```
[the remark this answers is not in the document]

flow/flow.go
```diff
@@ -893,17 +893,27 @@ func (f *Flow) newApplicationLayer(packet *Packet, opts Opts) error {
 	if (opts.ExtraLayers & DNSLayer) != 0 {
 		if layer := packet.Layer(layers.LayerTypeDNS); layer != nil {
 			d := layer.(*layers.DNS)
+			answers := make([]string, len(d.Answers))
+			questions := make([]string, len(d.Questions))
+			for _, v := range d.Answers {
```
safchain: it should be more something like this : 

```
                      if len(d.Answers) > 0 {
                               answers := make([]string, len(d.Answers))
                               for i, v := range d.Answers {
                                       answers[i] = string(v.IP)
                               }
                               f.DNS.Answer = answers
                        }
```

placed after `f.DNS = &fl.DNS{....`

hunchback: Since go 1.10.1 there should not be any significant difference in performance and using append is more idiomatic 

https://stackoverflow.com/questions/38654729/golang-slice-append-vs-assign-performance

safchain: Using append it appends an empty entry due to the make/len. That said my point was about adding only if needed (if statement).

flow/flow.go
```diff
@@ -893,17 +893,27 @@ func (f *Flow) newApplicationLayer(packet *Packet, opts Opts) error {
...
+				QR:       d.QR,
+				RA:       d.RA,
+				RD:       d.RD,
+				TC:       d.TC,
+				ANCount:  d.ANCount,
+				ARCount:  d.ARCount,
+				NSCount:  d.NSCount,
+				QDCount:  d.QDCount,
+				Question: questions,
+				Answer:   answers,
```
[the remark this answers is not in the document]

flow/flow.go

flow/flow.go
```diff
@@ -905,6 +905,18 @@ func (f *Flow) newApplicationLayer(packet *Packet, opts Opts) error {
 				NSCount: d.NSCount,
 				QDCount: d.QDCount,
 			}
+			questions := make([]string, len(d.Questions))
```
masco-enovance: as per the current implementation, this need to go inside the if block like answers.

andrekassis: why? Isn't there always a question in a DNS packet? even the response includes the question as far as I know so the "if" statement would be redundant..? isn't this correct?

masco-enovance: if some question present always no need for the ```if```

masco-enovance: in some special case the question may be ```zero``` as well. http://www-inf.int-evry.fr/~hennequi/CoursDNS/NOTES-COURS_eng/msg.html i refered this site, correct me if i am wrong.

andrekassis: You're probably right for there could be very special cases where no questions exist. changed this

PR #1662, opened 2019-02-16
Add advanced options for ovssflow capture type

dvandra:
Advanced options for "ovssflow" Capture Type : 
Options to add sampling rate and polling counters.

### DIFF
--- a/api/types/types.go
+++ b/api/types/types.go
@@ -70,22 +70,24 @@ func NewAlert() *Alert {
 
 // Capture describes a capture API
 type Capture struct {
-	BasicResource  `yaml:",inline"`
-	GremlinQuery   string           `json:"GremlinQuery,omitempty" valid:"isGremlinExpr" yaml:"GremlinQuery"`
-	BPFFilter      string           `json:"BPFFilter,omitempty" valid:"isBPFFilter" yaml:"BPFFilter"`
-	Name           string           `json:"Name,omitempty" yaml:"Name"`
-	Description    string           `json:"Description,omitempty" yaml:"Description"`
-	Type           string           `json:"Type,omitempty" valid:"isValidCaptureType" yaml:"Type"`
-	Count          int              `json:"Count" yaml:"Count"`
-	PCAPSocket     string           `json:"PCAPSocket,omitempty" yaml:"PCAPSocket"`
-	Port           int              `json:"Port,omitempty" yaml:"Port"`
-	RawPacketLimit int              `json:"RawPacketLimit,omitempty" valid:"isValidRawPacketLimit" yaml:"RawPacketLimit"`
-	HeaderSize     int              `json:"HeaderSize,omitempty" valid:"isValidCaptureHeaderSize" yaml:"HeaderSize"`
-	ExtraTCPMetric bool             `json:"ExtraTCPMetric" yaml:"ExtraTCPMetric"`
-	IPDefrag       bool             `json:"IPDefrag" yaml:"IPDefrag"`
-	ReassembleTCP  bool             `json:"ReassembleTCP" yaml:"ReassembleTCP"`
-	LayerKeyMode   string           `json:"LayerKeyMode,omitempty" valid:"isValidLayerKeyMode" yaml:"LayerKeyMode"`
-	ExtraLayers    flow.ExtraLayers `json:"ExtraLayers,omitempty" yaml:"ExtraLayers"`
+	BasicResource   `yaml:",inline"`
+	GremlinQuery    string           `json:"GremlinQuery,omitempty" valid:"isGremlinExpr" yaml:"GremlinQuery"`
+	BPFFilter       string           `json:"BPFFilter,omitempty" valid:"isBPFFilter" yaml:"BPFFilter"`
+	Name            string           `json:"Name,omitempty" yaml:"Name"`
+	Description     string           `json:"Description,omitempty" yaml:"Description"`
+	Type            string           `json:"Type,omitempty" valid:"isValidCaptureType" yaml:"Type"`
+	Count           int              `json:"Count" yaml:"Count"`
+	PCAPSocket      string           `json:"PCAPSocket,omitempty" yaml:"PCAPSocket"`
+	Port            int              `json:"Port,omitempty" yaml:"Port"`
+	SamplingRate    uint32           `json:"SamplingRate" yaml:"SamplingRate"`
+	PollingInterval uint32           `json:"PollingInterval" yaml:"PollingInterval"`
+	RawPacketLimit  int              `json:"RawPacketLimit,omitempty" valid:"isValidRawPacketLimit" yaml:"RawPacketLimit"`
+	HeaderSize      int              `json:"HeaderSize,omitempty" valid:"isValidCaptureHeaderSize" yaml:"HeaderSize"`
+	ExtraTCPMetric  bool             `json:"ExtraTCPMetric" yaml:"ExtraTCPMetric"`
+	IPDefrag        bool             `json:"IPDefrag" yaml:"IPDefrag"`
+	ReassembleTCP   bool             `json:"ReassembleTCP" yaml:"ReassembleTCP"`
+	LayerKeyMode    string           `json:"LayerKeyMode,omitempty" valid:"isValidLayerKeyMode" yaml:"LayerKeyMode"`
+	ExtraLayers     flow.ExtraLayers `json:"ExtraLayers,omitempty" yaml:"ExtraLayers"`
 }
 
 // NewCapture creates a new capture
--- a/cmd/client/capture.go
+++ b/cmd/client/capture.go
@@ -38,6 +38,8 @@ var (
 	captureType        string
 	nodeTID            string
 	port               int
+	samplingRate       uint32
+	pollingInterval    uint32
 	headerSize         int
 	rawPacketLimit     int
 	extraTCPMetric     bool
@@ -84,6 +86,8 @@ var CaptureCreate = &cobra.Command{
 		capture.Description = captureDescription
 		capture.Type = captureType
 		capture.Port = port
+		capture.SamplingRate = samplingRate
+		capture.PollingInterval = pollingInterval
 		capture.HeaderSize = headerSize
 		capture.ExtraTCPMetric = extraTCPMetric
 		capture.IPDefrag = ipDefrag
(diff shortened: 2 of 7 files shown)


dvandra:

dvandra:

common/capture.go
```diff
@@ -48,7 +48,7 @@ var (
 )
 
 func initCaptureTypes() {
-	CaptureTypes["ovsbridge"] = CaptureType{Allowed: []string{"ovssflow", "pcapsocket"}, Default: "ovssflow"}
+	CaptureTypes["ovsbridge"] = CaptureType{Allowed: []string{"ovssflow (Flow Samples)", "ovssflow (Counter Samples)", "ovssflow (Flow + Counter Samples)", "pcapsocket"}, Default: "ovssflow (Flow Samples)"}
```
safchain: I would prefer to keep only one probe but adding options to the API(capture) that will be passed/used when registering the probe (see https://github.com/skydive-project/skydive/blob/master/flow/probes/ovssflow.go#L241)

I think playing with sampling/polling parameter should be enough to have the expected behavior without modifying the sflow agent code ?
https://github.com/skydive-project/skydive/blob/master/flow/probes/ovssflow.go#L190

lebauce:
My bad, it seems the comments have not been addressed yet

dvandra:
Actually need to fix the "Packet decoding error : Invalid counter record type: 0"!

cmd/client/capture.go
```diff
@@ -38,6 +38,8 @@ var (
 	captureType        string
 	nodeTID            string
 	port               int
+	samplingrate       uint32
```
lebauce: s/samplingrate/samplingRate/

cmd/client/capture.go
```diff
@@ -38,6 +38,8 @@ var (
...
+	pollinginterval    uint32
```
lebauce: s/pollinginterval/pollingInterval/

api/types/types.go
```diff
@@ -70,22 +70,24 @@ func NewAlert() *Alert {
...
+	BasicResource   `yaml:",inline"`
+	GremlinQuery    string           `json:"GremlinQuery,omitempty" valid:"isGremlinExpr" yaml:"GremlinQuery"`
+	BPFFilter       string           `json:"BPFFilter,omitempty" valid:"isBPFFilter" yaml:"BPFFilter"`
+	Name            string           `json:"Name,omitempty" yaml:"Name"`
+	Description     string           `json:"Description,omitempty" yaml:"Description"`
+	Type            string           `json:"Type,omitempty" valid:"isValidCaptureType" yaml:"Type"`
+	Count           int              `json:"Count" yaml:"Count"`
+	PCAPSocket      string           `json:"PCAPSocket,omitempty" yaml:"PCAPSocket"`
+	Port            int              `json:"Port,omitempty" yaml:"Port"`
+	SamplingRate    uint32           `json:"SamplingRate" yaml:"SamplingRate"`
```
safchain: You should add a validator in order to check that the value is > 0. Moreover you should add something here
https://github.com/skydive-project/skydive/blob/master/api/server/capture.go#L104
in order to set the default values. That is probably why the functionnal tests are not passing. SamplingRate is set to 0 by default.

api/server/capture.go
```diff
@@ -118,6 +118,12 @@ func (c *CaptureAPIHandler) Create(r types.Resource) error {
 			}
 		}
 	}
+	if capture.SamplingRate == 0 {
+		capture.SamplingRate = 1
+	}
+	if capture.PollingInterval == 0 {
```
masco-enovance: we can check less than 10. if 10 is the minimum value.

statics/js/components/capture-form.js
```diff
@@ -68,9 +68,15 @@ Vue.component('capture-form', {
               <label for="port">Port</label>\
               <input id="port" type="number" class="form-control input-sm" v-model.number="port" min="0"/>\
             </div>\
+            <div class="form-group" v-if="captureType == \'ovssflow\'">\
+              <label for="samplingrate">Flow Sampling Rate</label>\
+              <input id="samplingrate" type="number" class="form-control input-sm" v-model.number="samplingrate" min="1"/>\
+              <label for="pollinginterval">Counter Polling Interval</label>\
+              <input id="pollinginterval" type="number" class="form-control input-sm" v-model.number="pollinginterval" min="0"/>\
```
masco-enovance: min is 10 right? and set the max too.

statics/js/components/capture-form.js
```diff
@@ -68,9 +68,15 @@ Vue.component('capture-form', {
               <label for="port">Port</label>\
               <input id="port" type="number" class="form-control input-sm" v-model.number="port" min="0"/>\
             </div>\
+            <div class="form-group" v-if="captureType == \'ovssflow\'">\
+              <label for="samplingrate">Flow Sampling Rate</label>\
+              <input id="samplingrate" type="number" class="form-control input-sm" v-model.number="samplingrate" min="1"/>\
```
masco-enovance: you can set the max value also similar to the validator.

api/server/capture.go
```diff
@@ -118,6 +118,12 @@ func (c *CaptureAPIHandler) Create(r types.Resource) error {
 			}
 		}
 	}
+	if capture.SamplingRate == 0 {
```
masco-enovance: check ```capture.SamplingRate < 1``` to avoid negative values also

tests/tests.go
```diff
@@ -454,6 +454,12 @@ func RunTest(t *testing.T, test *Test) {
 		capture.Type = tc.kind
 		capture.RawPacketLimit = tc.rawPackets
 		capture.Port = tc.port
+		if capture.SamplingRate == 0 {
```
masco-enovance: dito

dvandra:

cmd/client/capture.go
dvandra: +1

cmd/client/capture.go
dvandra: +1

api/server/capture.go
dvandra: Range of PollingInterval is 0 to 3600 seconds , default for SFlow is 2 seconds but with average 10 seconds will get better results! :)

statics/js/components/capture-form.js
dvandra: +1

statics/js/components/capture-form.js
dvandra: +1

api/server/capture.go
dvandra: +1

tests/tests.go
dvandra: +1

cmd/client/capture.go
dvandra: Thank you!

statics/js/components/capture-form.js
dvandra: Thank You!

api/types/types.go
dvandra: Done!
Added the default values inside the probe!
Thank You!

flow/probes/ovssflow.go
```diff
@@ -183,12 +183,19 @@ func (o *OvsSFlowProbesHandler) RegisterProbeOnBridge(bridgeUUID string, tid str
 	opts := tableOptsFromCapture(capture)
 	ft := o.fpta.Alloc(tid, opts)
 
+	if capture.SamplingRate < 1 {
```
safchain: With the current approach in order to get the counters we need to have samples(meaning flow) but I may want to not have flows. So I propose to do the following thing

```
if capture.SamplingRate < 1 {
   capture.SamplingRate = math.MaxUint32
}
```

In that case the number to so big that no sample will be sent and we will have the counters. What do you think ?

statics/js/components/capture-form.js
```diff
@@ -134,10 +140,12 @@ Vue.component('capture-form', {
       mode: "selection",
       visible: false,
       captureType: "",
-      captureLayerKeyMode: "",
+      layerKeyMode: "",
       nodeType: "",
       typeAllowed: false,
       port: 0,
+      samplingrate: 1,
```
safchain: nit: samplingRate
pollingInterval

statics/js/components/capture-form.js
```diff
@@ -315,12 +323,14 @@ Vue.component('capture-form', {
       capture.BPFFilter = this.bpf;
       capture.HeaderSize = this.headerSize;
       capture.RawPacketLimit = this.rawPackets;
-      capture.ExtraTCPMetric = this.tcpMetric;
+      capture.ExtraTCPMetric = this.extraTCPMetric;
```
safchain: good catch !

api/types/types.go
```diff
@@ -70,22 +70,24 @@ func NewAlert() *Alert {
...
+	BasicResource   `yaml:",inline"`
+	GremlinQuery    string           `json:"GremlinQuery,omitempty" valid:"isGremlinExpr" yaml:"GremlinQuery"`
+	BPFFilter       string           `json:"BPFFilter,omitempty" valid:"isBPFFilter" yaml:"BPFFilter"`
+	Name            string           `json:"Name,omitempty" yaml:"Name"`
+	Description     string           `json:"Description,omitempty" yaml:"Description"`
+	Type            string           `json:"Type,omitempty" valid:"isValidCaptureType" yaml:"Type"`
+	Count           int              `json:"Count" yaml:"Count"`
+	PCAPSocket      string           `json:"PCAPSocket,omitempty" yaml:"PCAPSocket"`
+	Port            int              `json:"Port,omitempty" yaml:"Port"`
+	SamplingRate    uint32           `json:"SamplingRate" valid:"min=0,max=4294967296" yaml:"SamplingRate"`
```
safchain: I don't think this is useful, the type definition avoid number outside of the range you defined

api/types/types.go
```diff
@@ -70,22 +70,24 @@ func NewAlert() *Alert {
...
+	PollingInterval uint32           `json:"PollingInterval" valid:"min=0,max=3600" yaml:"PollingInterval"`
```
safchain: nit: same here min is useless

flow/probes/ovssflow.go
dvandra: Yes, we can but then we will not able to capture any flows using default capture type (ovssflow) on ovs-bridge. (Will get counters only.)

And yes we can also take PollingInterval = 0, for no counters!

statics/js/components/capture-form.js
dvandra: +1

statics/js/components/capture-form.js
dvandra: Thanks!

api/types/types.go
dvandra: ok!

api/types/types.go
dvandra: Got it!

flow/probes/ovssflow.go
safchain: On the client side cli/ui you have set the default to 1 and 10 which fine as it was like this previously, meaning that without doing something specific the user will  get the same behavior. What I'm suggesting is to allow the user to set 0 flows or 0 counters. Since in order to get Counters we need to set `SamplingRate` to something > 0 I suggest to set it to something huge so that the counters will be sent without any sample packet.

dvandra:

api/types/types.go
```diff
@@ -70,22 +70,24 @@ func NewAlert() *Alert {
...
+	BasicResource   `yaml:",inline"`
+	GremlinQuery    string           `json:"GremlinQuery,omitempty" valid:"isGremlinExpr" yaml:"GremlinQuery"`
+	BPFFilter       string           `json:"BPFFilter,omitempty" valid:"isBPFFilter" yaml:"BPFFilter"`
+	Name            string           `json:"Name,omitempty" yaml:"Name"`
+	Description     string           `json:"Description,omitempty" yaml:"Description"`
+	Type            string           `json:"Type,omitempty" valid:"isValidCaptureType" yaml:"Type"`
+	Count           int              `json:"Count" yaml:"Count"`
+	PCAPSocket      string           `json:"PCAPSocket,omitempty" yaml:"PCAPSocket"`
+	Port            int              `json:"Port,omitempty" yaml:"Port"`
+	SamplingRate    uint32           `json:"SamplingRate" valid:"max=4294967295" yaml:"SamplingRate"`
```
lebauce: The "valid" part seems useless here

dvandra:

api/types/types.go
dvandra: Done!

PR #1664, opened 2019-02-19
python-api: improve python capture api

masco-enovance:


masco-enovance:
run skydive-python-tests

PR #1665, opened 2019-02-19
Runc guest hosts

hunchback:
This PR enhances runC with information extracted from guest `/etc/hosts` in cases that this file is mapped from the host machine by runC, as is the case when using k8s 1.12.x (as is now the default in IBM Kubernetes Services). 

To test:

```
sed -e %skydive/skydive%aidans/skydive% contrib/kubernetes/skydive.yaml > skydive2.yaml
kubectl apply -f skydive2.yaml
kubectl port-forward service/skydive-analyzer 8082:8082
```

Now open runC container and check that you see in Metadata `Runc.Hosts` field.

### DIFF
--- a/topology/probes/runc/hosts.go
+++ b/topology/probes/runc/hosts.go
@@ -1,0 +1,85 @@
+// +build linux
+
+/*
+ * Copyright (C) 2019 IBM, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy ofthe License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specificlanguage governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package runc
+
+import (
+	"bufio"
+	"net"
+	"os"
+	"strings"
+)
+
+type hosts struct {
+	IP       string              `json:"IP,omitempty"`
+	Hostname string              `json:"Hostname,omitempty"`
+	ByIP     map[string][]string `json:"ByIP,omitempty"`
+}
+
+func newHosts() *hosts {
+	hosts := new(hosts)
+	hosts.ByIP = make(map[string][]string)
+	return hosts
+}
+
+func readHosts(path string) (*hosts, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	hosts := newHosts()
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if err := scanner.Err(); err != nil {
+			return nil, err
+		}
+
+		if i := strings.IndexByte(line, '#'); i >= 0 {
+			// Discard comment.
+			line = line[0:i]
+		}
+
+		f := strings.Fields(line)
+		if len(f) < 2 {
+			continue
+		}
+
+		ip := net.ParseIP(f[0])
+		if ip == nil {
+			continue
+		}
+
+		if ip.IsLoopback() {
+			continue
+		}
+
+		hosts.IP = ip.String()
+		hosts.ByIP[hosts.IP] = make([]string, 0)
+		for i := 1; i < len(f); i++ {
+			hosts.Hostname = strings.ToLower(f[i])
+			hosts.ByIP[hosts.IP] = append(hosts.ByIP[hosts.IP], hosts.Hostname)
+		}
+	}
+
+	return hosts, nil
+}
--- a/topology/probes/runc/runc.go
+++ b/topology/probes/runc/runc.go
@@ -66,6 +66,12 @@ func (probe *Probe) containerNamespace(pid int) string {
 
 type initProcessStart int64
 
+// mount contains source (host machine) to destination (guest machine) mappings
+type mount struct {
+	Source      string `json:"source"`
+	Destination string `json:"destination"`
+}
+
 type containerState struct {
 	path             string
 	ID               string           `json:"id"`
@@ -73,6 +79,7 @@ type containerState struct {
 	InitProcessStart initProcessStart `json:"init_process_start"`
 	Config           struct {
 		Labels []string `json:"labels"`
+		Mounts []mount  `json:"mounts"`
 	} `json:"config"`
 }
 
@@ -172,6 +179,31 @@ func parseState(path string) (*containerState, error) {
 	return &state, nil
 }
 
+func getHostsFromState(state *containerState) (string, error) {
+	const path = "/etc/hosts"
+	for _, mount := range state.Config.Mounts {
+		if mount.Destination == path {
+			return mount.Source, nil
+		}
+	}
+	return "", fmt.Errorf("Unable to find binding of %s", path)
+}
+
+func parseHosts(state *containerState) *hosts {
+	path, err := getHostsFromState(state)
+	if err != nil {
+		return nil
+	}
+
+	hosts, err := readHosts(path)
+	if err != nil {
+		logging.GetLogger().Error(err)
+		return nil
+	}
+
+	return hosts
+}
+
 func getMetadata(state *containerState) graph.Metadata {
 	m := graph.Metadata{
 		"ContainerID": state.ID,
@@ -189,6 +221,10 @@ func getMetadata(state *containerState) graph.Metadata {
 				m["CreateConfig"] = cc
 			}
 		}
+	}
+
+	if hosts := parseHosts(state); hosts != nil {
+		m["Hosts"] = *hosts
 	}
 
 	return m
--- a/topology/probes/runc/runc_test.go
+++ b/topology/probes/runc/runc_test.go
@@ -18,10 +18,44 @@
 package runc
 
 import (
+	"io/ioutil"
+	"os"
 	"testing"
 
 	"github.com/skydive-project/skydive/common"
 )
+
+const stateBasename = "state.json"
+
+const stateContent = `{
+  "id": "c96f6a897a632970bd4d0aa34f94452a4ce236b1dc239bf010bcf12f98d5bc56",
+  "init_process_pid": 31046,
+  "init_process_start": 541882281,
+  "created": "2019-02-19T09:11:38.404438797Z",
+  "config": {
+    "mounts": [
+      {
+        "source": "/var/data/cripersistentstorage/io.containerd.grpc.v1.cri/sandboxes/87537844e40aaa9672ea2505600a94325221e901d04237356aff20c64849cbe0/resolv.conf",
+        "destination": "/etc/resolv.conf"
+      },
+      {
+        "source": "/var/data/kubelet/pods/5b9a9c62-3426-11e9-9f59-76f6e92e93c2/etc-hosts",
+        "destination": "/etc/hosts"
+      }
+    ]
+  }
+}`
+
+const hostsBasename = "etc-hosts"
+
+const hostsContent = `# Kubernetes-managed hosts file.
+127.0.0.1	localhost
+::1	localhost ip6-localhost ip6-loopback
+fe00::0	ip6-localnet # 1 space
+fe00::0	 ip6-mcastprefix # 2 spaces
+fe00::1	ip6-allnodes # 1 tab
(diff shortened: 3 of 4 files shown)


safchain:
I'm curious about the use case, can you elaborate on this ? 

Moreover if 30 containers bind mount /etc/hosts we will have the same information replicated lot of time. Maybe we should just report the bind mounts and add the /etc/hosts parsing at the host level ?

hunchback:
> I'm curious about the use case, can you elaborate on this ?

Without getting into too much details we want to translate the ephemeral IP with a string which remains the same between reboots, we use part of the hostname for that end.

> Moreover if 30 containers bind mount /etc/hosts we will have the same information replicated lot of time. Maybe we should just report the bind mounts and add the /etc/hosts parsing at the host level ?

Each container has its' own /etc/hosts, the host residing source is maintained by kubelet per each container - so if 30 containers means 30 pods - each will have a different /etc/hosts.

eranra:
@safchain We assume that this /etc/hosts file is in the pod level, so this is not 'host' level /etc/hosts bu the /etc/hosts/ that the k8s pod will 'see' ... on a single host there might be many of those. Each relats to specific pod. Now, multiple containers that share the same pod will be bound to the same file ... make sense? (what we are looking for is the DNS name that is associated with the IP of the POD. Remider containers in the same pod share IP and DNS).

safchain:
@hunchback @eranra thanks for the explanations. It could be nice to move the hosts file parse to the common package. I'll do a review ASAP

topology/probes/runc/runc_test.go
```diff
@@ -36,3 +36,47 @@ func TestLabels(t *testing.T) {
...
+	}
+
+	_, err = getHostsFromState(state)
+	if err != nil {
+		t.Errorf("unable to find /etc/hosts entry in file: %s", stateFilename)
+	}
+}
+
+func TestReadHosts(t *testing.T) {
+	const path = "etc-hosts"
```
nplanel: insert the testing content directly in the test file
Thanks to `

topology/probes/runc/etc-hosts
```diff
@@ -0,0 +1,8 @@
+# Kubernetes-managed hosts file.
```
nplanel: 1/ don't include this file here for the test only
2/ insert space,tab in front of the IPs and comment
3/ add comment at the end of the line like   ip host #comment

topology/probes/runc/state.json
```diff
@@ -0,0 +1,2958 @@
+{
```
nplanel: reduce the content to the minimum and include it in the test file please

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,137 @@
...
+	if last < len(s) {
+		a[n] = s[last:]
+		n++
+	}
+	return a[0:n]
+}
+
+// Split line on white space
+func getFields(s string) []string {
+	return splitAtBytes(s, " \r\t\n")
```
safchain: I propose the following split function

```
func split(str string, sep string) (result []string) {
	var s scanner.Scanner
	s.Init(strings.NewReader(str))

	for tok := s.Scan(); tok != scanner.EOF; tok = s.Scan() {
		result = append(result, s.TokenText())
	}
	
	return result
}
```

nplanel:
> @safchain We assume that this /etc/hosts file is in the pod level, so this is not 'host' level /etc/hosts bu the /etc/hosts/ that the k8s pod will 'see' ... on a single host there might be many of those. Each relats to specific pod. Now, multiple containers that share the same pod will be bound to the same file ... make sense? (what we are looking for is the DNS name that is associated with the IP of the POD. Remider containers in the same pod share IP and DNS).

@eranra Why not simply use the /etc/hostname for this purpose ? ... the IP is easy to retrieve from others way

hunchback:
> > @safchain We assume that this /etc/hosts file is in the pod level, so this is not 'host' level /etc/hosts bu the /etc/hosts/ that the k8s pod will 'see' ... on a single host there might be many of those. Each relats to specific pod. Now, multiple containers that share the same pod will be bound to the same file ... make sense? (what we are looking for is the DNS name that is associated with the IP of the POD. Remider containers in the same pod share IP and DNS).
> 
> @eranra Why not simply use the /etc/hostname for this purpose ? ... the IP is easy to retrieve from others way

@nplanel - because we have not direct access from within the container to `/etc/hostname` - the reason we can access `/etc/hosts` is because it has a mapping to the **host** and the runC has access to this host path (runC is actually being run within a pod/container but we have mapped the host directly into this pod/container).

topology/probes/runc/hosts.go
hunchback: `strings.Fields` does what is needed

topology/probes/runc/runc.go
```diff
@@ -191,6 +222,10 @@ func getMetadata(state *containerState) graph.Metadata {
 		}
 	}
 
+	if hosts != nil {
```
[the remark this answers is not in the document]
hunchback: ok

topology/probes/runc/runc.go
```diff
@@ -172,7 +179,31 @@ func parseState(path string) (*containerState, error) {
...
+	return "", fmt.Errorf("Unable to find binding of %s", path)
+}
+
+func parseHosts(state *containerState) (*hosts, error) {
+	path, err := getHostsFromState(state)
+	if err != nil {
+		return nil, err
+	}
+
+	hosts, err := readHosts(path)
```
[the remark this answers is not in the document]
hunchback: `parseHosts` after my last change only returns one parameter

topology/probes/runc/state.json
hunchback: I have reduced context

regarding inlining the file as `parseHosts()` accepts a filename should we not leave `state.json` as a separate file?

topology/probes/runc/etc-hosts
hunchback: 1/ - `readHosts` accepts a file anyway
2/ - handled

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, err
+	}
+	defer file.Close()
+
+	hosts := newHosts()
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if i := strings.IndexByte(line, '#'); i >= 0 {
```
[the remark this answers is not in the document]
hunchback: you are only handling a line which includes whitespace + comment, while my current implementation eliminates everything after the comment and later on will dispose of lines w/o any context (or invalid content).

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,137 @@
...
+func readHosts(path string) (*hosts, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	hosts := newHosts()
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
```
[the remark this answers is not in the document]
hunchback: but here I want to get line by line (only later do I split each line into tokens)

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		if !ip.IsLoopback() {
+			// The last non-loopback sticks.
+			hosts.IP = ip.String()
+			hosts.Hostname = hostname
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
```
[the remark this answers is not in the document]
hunchback: ok

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,137 @@
...
+		f := getFields(line)
+		if len(f) < 2 {
+			continue
+		}
+
+		ip := net.ParseIP(f[0])
+		if ip == nil {
+			continue
+		}
+
```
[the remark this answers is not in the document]
hunchback: ok

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,137 @@
...
+	return a[0:n]
+}
+
+// Split line on white space
+func getFields(s string) []string {
+	return splitAtBytes(s, " \r\t\n")
+}
+
+// toLower makes ASCII lowercase.
+func toLower(s string) string {
```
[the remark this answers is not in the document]
hunchback: ok - using `strings.ToLower()`

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,137 @@
...
+	for i := 0; i < len(s); i++ {
+		if strings.IndexByte(t, s[i]) >= 0 {
+			n++
+		}
+	}
+	return n
+}
+
+// Split s at any bytes in t.
+func splitAtBytes(s string, t string) []string {
```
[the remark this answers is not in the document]
hunchback: using `strings.Fields()`

hunchback:

hunchback:

topology/probes/runc/state.json
nplanel: in the test file you can create a temp file ioutil.TempFile, it make no sense to pollute the repo with testing text files, specially not in tests/ subfolders

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,97 @@
...
+		ip := net.ParseIP(f[0])
+		if ip == nil {
+			continue
+		}
+
+		if ip.IsLoopback() {
+			continue
+		}
+
+		hosts.IP = ip.String()
```
eranra: We need to handle cases in which there is more than a single IP in the /etc/hosts (that is not loopback) ... the current solution is not guarantied to work in such cases in any consistent way.

You might take the largest or smallest IP ... or sort lexicographic .... or something ... in any event taking the 'last' IP in the table might not be consistent over reboots or other operations done by k8s

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,97 @@
...
+		hosts.ByIP[hosts.IP] = make([]string, 0)
+		for i := 1; i < len(f); i++ {
+			hosts.Hostname = strings.ToLower(f[i])
```
eranra: We need to handle cases in which there is more than a single IP in the /etc/hosts (that is not loopback) ... the current solution is not guarantied to work in such cases in any consistent way.

You might take the largest or smallest IP ... or sort lexicographic .... or something ... in any event taking the 'last' IP in the table might not be consistent over reboots or other operations done by k8s

hunchback:
@nplanel - I have embedded the test files - please review

topology/probes/runc/runc.go
```diff
@@ -172,6 +179,30 @@ func parseState(path string) (*containerState, error) {
...
+
+func parseHosts(state *containerState) *hosts {
+	path, err := getHostsFromState(state)
+	if err != nil {
+		return nil
+	}
+
+	hosts, err := readHosts(path)
+	if err != nil {
+		return nil
```
nplanel: would be good to log the error

topology/probes/runc/runc_test.go
```diff
@@ -36,3 +70,79 @@ func TestLabels(t *testing.T) {
...
 }
+
+func tempFile(filename, content string) (string, error) {
+	tmpfile, err := ioutil.TempFile("", filename)
+	if err != nil {
+		return "", err
+	}
+
+	if _, err := tmpfile.Write([]byte(content)); err != nil {
+		return "", err
```
nplanel: tmpfile.Close();

hunchback:

topology/probes/runc/hosts.go
hunchback: moved to: https://github.com/skydive-project/skydive/issues/1679

topology/probes/runc/hosts.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func newHosts() *hosts {
+	hosts := new(hosts)
+	hosts.ByIP = make(map[string][]string)
+	return hosts
+}
+
+// Count occurrences in s of any bytes in t.
+func countAnyByte(s string, t string) int {
```
nplanel: could you remove it, unused

topology/probes/runc/runc_test.go
hunchback: ok

topology/probes/runc/runc.go
hunchback: ok

topology/probes/runc/hosts.go
hunchback: ok

hunchback:

nplanel:
run tests

hunchback:
run skydive-vpp-tests

hunchback:
run skydive-scale-tests

hunchback:

hunchback:
run skydive-vpp-tests

PR #1666, opened 2019-02-19
workflow: move workflow execution to server side

masco-enovance:


rbac/policy.csv
```diff
@@ -21,6 +21,8 @@ p, admin, noderule, read, allow
 p, admin, noderule, write, allow
 p, admin, edgerule, read, allow
 p, admin, edgerule, write, allow
+p, admin, workflowcall, read, allow
```
lebauce: this one doesn't seem to be used

rbac/policy.csv
```diff
@@ -21,6 +21,8 @@ p, admin, noderule, read, allow
...
+p, admin, workflowcall, write, allow
```
lebauce: workflow.call according to the suggest route

api/server/workflow_call.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return workflow.(*types.Workflow), nil
+}
+
+func (wc *WorkflowCallAPIHandler) registerEndPoints(s *shttp.Server, authBackend shttp.AuthenticationBackend) {
+	routes := []shttp.Route{
+		{
+			Name:        "WorkflowCall",
+			Method:      "POST",
+			Path:        "/api/workflowcall",
```
lebauce: I would prefer that we use /api/workflow/<id>/call

masco-enovance: done.

rbac/policy.csv
masco-enovance: removed

rbac/policy.csv
masco-enovance: done

api/server/runtime.go
```diff
@@ -0,0 +1,138 @@
...
+		data := []byte(call.Argument(2).String())
+
+		subs := strings.Split(url, "/") // filepath.Base(url)
+		if len(subs) < 3 {
+			return r.MakeCustomError("WrongArgument", fmt.Sprintf("Malformed URL %s", url))
+		}
+		resource := subs[2]
+
+		// For topology query, we directly call the Gremlin engine
+		if resource == "topology" {
```
safchain: this is not really related to this patch but I don't understand the purpose of this :)
@lebauce any idea ?

safchain: ok got it , tricky thing :)

PR #1667, opened 2019-02-20
hold ended flows

orozery:
Recently (in #1532), we started deleting flows immediately once we detect a TCP FIN/RST flags.
However, such flows can still see new packets, and thus we propose to add a "hold" duration to keep these flows in the flows table a bit longer.

Another change I made here is not report flows as "expired" if they are already reported as ended by other means (such as TCP FIN/RST).

### DIFF
--- a/flow/flow.go
+++ b/flow/flow.go
@@ -528,6 +528,8 @@ func (f *Flow) Init(now int64, nodeTID string, uuids UUIDs) {
 
 	f.NodeTID = nodeTID
 	f.ParentUUID = uuids.ParentUUID
+
+	f.FinishType = FlowFinishType_NOT_FINISHED
 }
 
 // initFromPacket initializes the flow based on packet data, flow key and ids
--- a/flow/table.go
+++ b/flow/table.go
@@ -32,6 +32,9 @@ import (
 	"github.com/skydive-project/skydive/logging"
 )
 
+// HoldTimeoutMilliseconds is the number of milliseconds for holding ended flows before they get deleted from the flow table
+const HoldTimeoutMilliseconds = int64(time.Duration(15) * time.Second / time.Millisecond)
+
 // ExpireUpdateFunc defines expire and updates callback
 type ExpireUpdateFunc func(f *FlowArray)
 
@@ -205,8 +208,10 @@ func (ft *Table) expire(expireBefore int64) {
 			}
 
 			logging.GetLogger().Debugf("Expire flow %s Duration %v", f.UUID, duration)
-			f.FinishType = FlowFinishType_TIMEOUT
-			expiredFlows = append(expiredFlows, f)
+			if f.FinishType == FlowFinishType_NOT_FINISHED {
+				f.FinishType = FlowFinishType_TIMEOUT
+				expiredFlows = append(expiredFlows, f)
+			}
 
 			// need to use the key as the key could be not equal to the UUID
 			delete(ft.table, k)
@@ -259,9 +264,6 @@ func (ft *Table) update(updateFrom, updateTime int64) {
 		if f.XXX_state.updateVersion > ft.updateVersion {
 			ft.updateMetric(f, updateFrom, updateTime)
 			updatedFlows = append(updatedFlows, f)
-			if f.FinishType != FlowFinishType_NOT_FINISHED {
-				delete(ft.table, k)
-			}
 		} else if updateTime-f.Last > ft.appTimeout[f.Application] && ft.appTimeout[f.Application] > 0 {
 			updatedFlows = append(updatedFlows, f)
 			f.FinishType = FlowFinishType_TIMEOUT
@@ -271,6 +273,10 @@ func (ft *Table) update(updateFrom, updateTime int64) {
 		}
 
 		f.XXX_state.lastMetric = f.Metric.Copy()
+
+		if f.FinishType != FlowFinishType_NOT_FINISHED && updateTime-f.Last >= HoldTimeoutMilliseconds {
+			delete(ft.table, k)
+		}
 	}
 
 	if len(updatedFlows) != 0 {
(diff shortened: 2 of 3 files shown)


hunchback:
How can you change be verified? Please provide a sequence that can prove its' working - preferably add a unit test.

orozery:
@safchain It was suggested by @hunchback that we add another ticker that will run every `flow.hold` seconds (default 15), and will delete those flows that were detected as finished (TCP FIN/RST) on the last `update` iteration (they will be added to some dedicated "finishedFlows" array, instead of going over the entire flow table).

Your view on that suggestion?

safchain:
@orozery the approach suggested by @hunchback seems to be good. Indeed it will limit the amount of flow.

orozery:
@safchain I've implemented @hunchback 's suggestion.
Also added a unit test.

orozery:
I will shortly fix the test.

flow/flow.go
```diff
@@ -845,6 +845,8 @@ func (f *Flow) newTransportLayer(packet *Packet, opts Opts) error {
 			f.TCPMetric = &TCPMetric{}
 		}
 
+		f.FinishType = FlowFinishType_NOT_FINISHED
```
nplanel: could you put this line to flow.Init() line 530

flow/table.go
```diff
@@ -126,6 +128,8 @@ func NewTable(updateHandler *Handler, expireHandler *Handler, nodeTID string, op
 		tcpAssembler:      NewTCPAssembler(),
 		appPortMap:        NewApplicationPortMapFromConfig(),
 		appTimeout:        appTimeout,
+		holdTime:          int64(config.GetConfig().GetInt("flow.hold") * 1000),
```
nplanel: please remove * 1000

flow/table.go
```diff
@@ -437,6 +464,9 @@ func (ft *Table) Run() {
 	expireTicker := time.NewTicker(ft.expireHandler.every)
 	defer expireTicker.Stop()
 
+	holdTicker := time.NewTicker(time.Duration(ft.holdTime) * time.Millisecond)
```
nplanel: time.Second

flow/table.go
```diff
@@ -297,6 +308,22 @@ func (ft *Table) expireAt(now time.Time) {
 	ft.lastExpire = common.UnixMillis(now)
 }
 
+func (ft *Table) unholdAt(now time.Time) {
+	nowMilliseconds := common.UnixMillis(now)
+	for k := range ft.holdedFlows {
+		f, ok := ft.table[k]
+		if ok {
+			if f.FinishType != FlowFinishType_NOT_FINISHED {
+				if nowMilliseconds-f.Last < ft.holdTime {
```
nplanel: (ft.holdTime * 1000)

config/config.go
```diff
@@ -118,6 +118,7 @@ func init() {
 
 	cfg.SetDefault("flow.expire", 600)
 	cfg.SetDefault("flow.update", 60)
+	cfg.SetDefault("flow.hold", 15)
```
nplanel: could you set flow.how as a constant in flow.go like const HoldTimeout
don't need to be a parameter as it's not to be attend to change

flow/table.go
```diff
@@ -271,6 +274,14 @@ func (ft *Table) update(updateFrom, updateTime int64) {
 		}
 
 		f.XXX_state.lastMetric = f.Metric.Copy()
+
+		if f.FinishType != FlowFinishType_NOT_FINISHED {
+			if updateTime-f.Last >= ft.holdTime {
```
nplanel: (ft.holdTime * 1000)

orozery:
@nplanel thanks for your comments! I made code changes.

flow/table.go
```diff
@@ -297,6 +309,22 @@ func (ft *Table) expireAt(now time.Time) {
...
+			if f.FinishType != FlowFinishType_NOT_FINISHED {
+				if nowMilliseconds-f.Last < HoldTimeoutMilliseconds {
+					continue
+				}
+				delete(ft.table, k)
+			}
+		}
+		delete(ft.holdedFlows, k)
+	}
+}
```
nplanel: to have the same behavior you should advise clients like in expire()

if len(expiredFlows) > 0 {
	ft.expireHandler.callback(&FlowArray{Flows: expiredFlows})
}

flow/table.go
```diff
@@ -271,6 +275,14 @@ func (ft *Table) update(updateFrom, updateTime int64) {
 		}
 
 		f.XXX_state.lastMetric = f.Metric.Copy()
+
+		if f.FinishType != FlowFinishType_NOT_FINISHED {
+			if updateTime-f.Last >= HoldTimeoutMilliseconds {
```
nplanel: We just need to do that

```
			if updateTime-f.Last >= HoldTimeoutMilliseconds {
				delete(ft.table, k)
			}

```
We don't need to check if any packet arrvied in the flow 15sec after, as we tested already line 280 that the flow didn't received any packets.
So you can remove the HoldTimeout ticker mechanism and holdFlows map

flow/table.go
orozery: you suggest adding another handler?
I do not see the point in that. Unlike expired flows, flows that were holded have already been reported as finished to the user, so he should not expect to get a second update.

flow/table.go
orozery: I don't understand.
If I remove the holdticker then flows will only get deleted after "flow.update" - 60 seconds, instead of 15 seconds.
This is what we had before,  but I added the holdticker as @hunchback and @safchain suggested.

nplanel: @orozery It's a bit confusing I understand, but as discuss with @safchain ; The whole purpose of this feature didn't require a ticker, basically you just need to expire the flow (with timeout flag and no packets arrived last HoldTimeout).
This will have same behavior and will avoid over engineering.

orozery:
@safchain so should I revert the change suggested by @hunchback to add the new ticker?

safchain:
@orozery I discussed with @nplanel and I think he is right. As we push the flow in the hold list only in the update function I think that is better to remove the ticker to keep it simple. With this approach we will keep a bit longer the flow in the flow table (per update ticker) but it was potentially the same with the ticker.

safchain:

PR #1671, opened 2019-02-21
added libvirt to list of probes in yaml file

KalmanMeth:
The libvirt probe did not appear as a possibility in the etc/skydive.yaml file.
I added libvirt (commented) to the list of probes available.


KalmanMeth:
Yes, one of the admin could/should review this patch.

PR #1675, opened 2019-02-22
ansible: handle analyzer and agent docker tag independently

safchain:


safchain:
rerun skydive-scale-tests

PR #1687, opened 2019-03-08
workflow: make 'finally' work in otto engine

masco-enovance:


masco-enovance:

hunchback:
When running the following (using online engine: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve):

```
var promise1 = Promise.resolve("resolve").finally(function() { console.log("finally") });

promise1.then(function(value) {
  console.log(value);
  throw "error"
}).catch(function(err) {
  console.log(err);
});
```

I got output:

```
**> "finally"**
> "resolve"
> "error"
```

While I was expecting:

```
> "resolve"
> "error"
**> "finally"**
```

masco-enovance:
@hunchback thanks for the review!
The output what you got is the correct one. because you added the prints in that order.

and this patch is for otto engine not for the browser engine.
please check your test in otto environment.

PR #1689, opened 2019-03-10
flow object store: batch flows together to improve performance

orozery:
Prior to this PR, flows sent from the analyzer to the object storage (every `flow.update` seconds) were immediately flushed to the object store.
We change this behavior to allow the user to batch the flows into larger objects.
To achieve that, we introduce 2 new configuration parameters: `max_seconds_per_object`, `max_flows_per_object`.
Furthermore, we tag flows by their "network flow direction", either internal cluster traffic, inbound or outbound.
The flows are now in objects by their tags.

@safchain Note this change is not affecting skydive, but only the object store subscriber.
I've added extensive unit testing + I did a run test to verify everything's working as expected.


contrib/objectstore/subscriber/flowclassifier/flowclassifier.go
```diff
@@ -0,0 +1,91 @@
+package flowclassifier
```
hunchback: missing license preamble

hunchback:
@orozery - I suggest you change the file layout of  `contrib/objectstore` and move all files directly under `contrib/objectstore` as I see little justification of having many subdirectories each with 1-2 files.

orozery:
@hunchback I added the license headers + moved all non-main files to the same package.

contrib/objectstore/subscriber/flowclassifier.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	if isDstInCluster {
+		return tagIngress
+	}
+	return tagOther
+}
+
+// isClusterIP check if IP is in defined subnet
+func (fc *FlowDirectionClassifier) isClusterIP(ip string) (bool, error) {
```
hunchback: can be simplified to prototype  `isClusterIP(ip string) bool` returning true iff ip is both valid and within at-least one cluster-netmask.

contrib/objectstore/subscriber/objectstore.go
```diff
@@ -54,45 +76,104 @@ func (s *Subscriber) OnStructMessage(c ws.Speaker, msg *ws.StructMessage) {
...
 func (s *Subscriber) StoreFlows(flows []*flow.Flow) error {
-	if len(flows) == 0 {
-		return nil
+	endTime := time.Now()
+
+	s.flowsMutex.Lock()
+	defer s.flowsMutex.Unlock()
+
+	// transform flows and save to in-memory arrays, based on tag
+	if flows != nil {
```
hunchback: is there any execution path which may yield `flows == nil` in this line? if not then you can eliminate this if.

contrib/objectstore/subscriber/objectstore.go
```diff
@@ -54,45 +76,104 @@ func (s *Subscriber) OnStructMessage(c ws.Speaker, msg *ws.StructMessage) {
...
+			if s.flowTransformer != nil {
+				transformedFlow = s.flowTransformer.Transform(fl)
+			} else {
+				transformedFlow = fl
+			}
+			if transformedFlow != nil {
+				flowTag := s.flowClassifier.GetFlowTag(fl)
+				s.flows[flowTag] = append(s.flows[flowTag], transformedFlow)
+			}
+		}
```
hunchback:
```
                for _, fl := range flows {
                        	tag := s.flowClassifier.GetFlowTag(fl)
                                if s.flowTransformer != nil {
                                        fl = s.flowTransformer.Transform(fl)
                                        if fl == nil {
                                                continue
                                        }
                                }
				s.flows[tag] = append(s.flows[tag], fl)
                  }
```

contrib/objectstore/subscriber/objectstore.go
```diff
@@ -54,45 +76,104 @@ func (s *Subscriber) OnStructMessage(c ws.Speaker, msg *ws.StructMessage) {
...
 }
 
-// StoreFlows writes a set of flows to the object storage service
+// StoreFlows store flows in memory, before being written to the object store
 func (s *Subscriber) StoreFlows(flows []*flow.Flow) error {
-	if len(flows) == 0 {
-		return nil
+	endTime := time.Now()
+
+	s.flowsMutex.Lock()
```
hunchback: can you explain why is the mutex needed?

contrib/objectstore/subscriber/flowclassifier.go
orozery: This will change the semantics of `GetFlowTag`, so I don't think I can follow your suggestion.

contrib/objectstore/subscriber/objectstore.go
orozery: We call it with a `nil` argument by a `time.AfterFunc` invocation.

contrib/objectstore/subscriber/objectstore.go
orozery: that won't work:
`subscriber/objectstore.go:91:8: cannot use s.flowTransformer.Transform(fl) (type interface {}) as type *flow.Flow in assignment: need type assertion`

contrib/objectstore/subscriber/objectstore.go
orozery: because I call `StoreFlows` asynchronously, using a `time.AfterFunc`.

orozery:
I think the failures of `skydive-functional-hw-tests`, `skydive-k8s-tests` are unrelated to this PR.

contrib/objectstore/subscriber/secadvisor.go
```diff
@@ -0,0 +1,59 @@
...
+		LayersPath:       f.LayersPath,
+		Network:          f.Network,
+		Transport:        f.Transport,
+		LastUpdateMetric: f.LastUpdateMetric,
+		Metric:           f.Metric,
+		Start:            f.Start,
+		Last:             f.Last,
+		FinishType:       f.FinishType,
+	}
+}
```
hunchback: I would much rather that we had a single transformer which could be setup for this use-case (of security-advisor) using configuration - such as:

```
transformer:
  fields:
    # transformed-field-name: original-field-name 
    UUID: UUID
    LayersPath: LayersPath
    Network: Network
    Transport: Transport
    LastUpdateMetric: LastUpdateMetric
    Metric: Metric
    Start: Start
    Last: Last
    FinishType: FinishType
```

Or alternatively:

```
transformer:
  include:
  - UUID
  - LayersPath
  - Network
  - Transport
  - LastUpdateMetric
  - Metric
  - Start
  - Last
  - FinishType
```

orozery: This approach was already implemented a few months ago, but was rejected by @safchain:
https://github.com/skydive-project/skydive/pull/1620#discussion_r250611550

orozery: @hunchback In retrospect, it was a good decision to do it the way we did (with a fixed dedicated transformer for security advisor).
The requirements raised by security-advisor architects, other than the plain removal of fields, are pretty bizarre. I cannot envision a way for them to be generalized.

contrib/objectstore/subscriber/objectstore.go
hunchback: What about:

```
                  for _, fl := range flows {
                        	tag := s.flowClassifier.GetFlowTag(fl)
                                newfl := interface{}(fl)
                                if s.flowTransformer != nil {
                                        newfl = s.flowTransformer.Transform(fl)
                                }
				s.flows[tag] = append(s.flows[tag], newfl)
                  }
```

contrib/objectstore/subscriber/objectstore.go
hunchback: ok

contrib/objectstore/subscriber/objectstore.go
hunchback: ok

contrib/objectstore/subscriber/flowclassifier.go
hunchback: are you ever making use of the returned `error`? if not then maybe its' redundant.

contrib/objectstore/subscriber/flowclassifier/flowclassifier.go
hunchback: ok

contrib/objectstore/subscriber/objectstore.go
orozery: I intentionally allowed `Transform` to return nil, to indicate that a flow should be discarded.
This will allow security-advisor to discard the "STARTED" flows.

contrib/objectstore/subscriber/secadvisor.go
hunchback: @safchain - what was the reason you rejected the approach of setting in configuration which fields making it through to the target object store (vs. using target specific structures such as the one used for security-advisor)?

safchain: @hunchback because of performances. It is better to have a predefined serializer instead of having to look at what field to serialize for each flow. But I'm open to any other solution that will keep good performance.

safchain:
need to be rebased

PR #1690, opened 2019-03-10
K8s: fix linker for non-namespace links

hunchback:
this PR fixes a bug causing linker links to resources outside namespace to fail sometimes, specifically:
- storageclass --> namespace/pvc
- namespace/pvc --> pv

the fix removes automatic filtering by namespace in linker GetABLinks()/GetBALinks() and now in cases where namespace filtering is needed it the responsibility of implementer of XXXAreLinked() call back.

to test this works:

```
tests/k8s/storage.sh start
```

now check WebUI that links exists:
- [pod] default/task-pv-pod --> [pvc] default/task-pv-claim
- [sc] standard --> [pvc] default/task-pv-claim
- [sc] standard --> [pv] default/task-pv-volume

PR #1696, opened 2019-03-11
K8s empty service links

hunchback:
this fixes the case in which a service (such as default/kubernetes) will point to all pods in case of an empty label selector (which is incorrect).

to check you can create a pod in the default namespace and make sure default/kubernetes service does **not** point to it.

to create the pod you can use:

```
kubectl apply -f tests/k8s/pod.yaml
```

NOTE: depends to https://github.com/skydive-project/skydive/pull/1690


safchain:
need to be rebased

safchain:
@hunchback this PR conflict since #1690 merged, can you please rebase it

hunchback:

PR #1697, opened 2019-03-11
sriov and libvirt

pierrecregut:
This is a very initial support for SRIOV with focus on the VM case (libvirt).
Principle:
* we use netlink to get initial information on the "virtual functions" (VF), the interface being the "physical function" (PF)
* we use ethtool to get the pci address of the PF,
* we explore /sys/bus/pci/devices/<address> to get the sriov offset and stride used to compute the pci address of the VF.
* libvirt will see the VF as a <hostdev> with driver vfio. The source address will be the pci address of the VF.
* We should be able to use peering to associate the mac address discovered in the VF to the mac address of the interface on the guest.

### DIFF
--- a/graffiti/graph/processor.go
+++ b/graffiti/graph/processor.go
@@ -57,8 +57,6 @@ func NewProcessor(g *Graph, listenerHandler ListenerHandler, m ElementMatcher, i
 
 // DoAction will perform the action for nodes matching values.
 func (processor *Processor) DoAction(action NodeAction, values ...interface{}) {
-	processor.Lock()
-	defer processor.Unlock()
 	nodes, _ := processor.MetadataIndexer.Get(values...)
 	kont := true
 	for _, node := range nodes {
@@ -70,12 +68,14 @@ func (processor *Processor) DoAction(action NodeAction, values ...interface{}) {
 	if kont {
 		act := deferred{action: action}
 		hash := Hash(values...)
+		processor.Lock()
 		if actions, ok := processor.actions[hash]; ok {
-			actions = append(actions, act)
+			processor.actions[hash] = append(actions, act)
 		} else {
 			actions := []deferred{act}
 			processor.actions[hash] = actions
 		}
+		processor.Unlock()
 	}
 }
 
@@ -96,25 +96,33 @@ func (processor *Processor) Cancel(values ...interface{}) {
 	processor.Unlock()
 }
 
+// OnNodeUpdated event
+func (processor *Processor) OnNodeUpdated(n *Node) {
+	processor.OnNodeAdded(n)
+}
+
 // OnNodeAdded event
 func (processor *Processor) OnNodeAdded(n *Node) {
 	if vValues, err := getFieldsAsArray(n, processor.MetadataIndexer.indexes); err == nil {
 		for _, values := range vValues {
 			hash := Hash(values...)
-			processor.Lock()
-			defer processor.Unlock()
-			if actions, ok := processor.actions[hash]; ok {
+			processor.RLock()
+			actions, ok := processor.actions[hash]
+			processor.RUnlock()
+			if ok {
 				var keep []deferred
 				for _, action := range actions {
 					if action.action.ProcessNode(processor.MetadataIndexer.graph, n) {
 						keep = append(keep, action)
 					}
 				}
+				processor.Lock()
 				if len(keep) == 0 {
 					delete(processor.actions, hash)
 				} else {
 					processor.actions[hash] = keep
 				}
+				processor.Unlock()
 			}
 		}
 	}
--- a/tests/sriov_test.go
+++ b/tests/sriov_test.go
@@ -22,32 +22,60 @@ package tests
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
+	"path"
+	"strings"
 	"testing"
 )
 
+const sysNet = "/sys/class/net"
+
+func findSRIOV() string {
+	files, err := ioutil.ReadDir(sysNet)
+	if err != nil {
+		return ""
+	}
+	for _, info := range files {
+		device := info.Name()
+		sriovFile := path.Join(sysNet, device, "device", "sriov_numvfs")
+		if _, err = os.Stat(sriovFile); os.IsNotExist(err) {
+			continue
+		}
+		operstateFile := path.Join(sysNet, device, "operstate")
+		contents, err := ioutil.ReadFile(operstateFile)
+		if err == nil && strings.Trim(string(contents), "\r\n") == "up" {
+			return device
+		}
+	}
+	return ""
+}
+
 func TestSRIOV(t *testing.T) {
+	device := findSRIOV()
+	if device == "" {
+		t.Skip("No SRIOV device found.")
+	}
+	sriovFile := path.Join(sysNet, device, "device", "sriov_numvfs")
 	test := &Test{
 		setupFunction: func(c *TestContext) error {
-			return ioutil.WriteFile("/sys/class/net/enp5s0f0/device/sriov_numvfs", []byte("2"), 0664)
+			return ioutil.WriteFile(sriovFile, []byte("2"), 0664)
 		},
 
 		tearDownFunction: func(c *TestContext) error {
-			return ioutil.WriteFile("/sys/class/net/enp5s0f0/device/sriov_numvfs", []byte("0"), 0664)
+			return ioutil.WriteFile(sriovFile, []byte("0"), 0664)
 		},
 
 		mode: Replay,
 
 		checks: []CheckFunction{func(c *CheckContext) error {
-			gremlin := c.gremlin.V().Has("Name", "enp5s0f0").OutE("RelationType", "vf").OutV()
+			gremlin := c.gremlin.V().Has("Name", device).OutE("RelationType", "vf").OutV()
 			nodes, err := c.gh.GetNodes(gremlin)
 			if err != nil {
 				return err
 			}
-
 			if len(nodes) != 2 {
-				return fmt.Errorf("Expected 2 nodes, got %+v", nodes)
+				return fmt.Errorf("Expected 2 nodes, got %+v (sriov intf: %s)", nodes, device)
 			}
-
 			return nil
 		}},
 	}
--- a/topology/probes/libvirt/libvirt.go
+++ b/topology/probes/libvirt/libvirt.go
@@ -22,7 +22,9 @@ package libvirt
 import (
 	"context"
 	"encoding/xml"
+	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/skydive-project/skydive/config"
@@ -44,20 +46,24 @@ type Probe struct {
 	uri          string                // uri of the libvirt connection
 	cancel       context.CancelFunc    // cancel function
 	root         *graph.Node           // root node for ownership
-	tunProcessor *graph.Processor      // metadata indexer
+	tunProcessor *graph.Processor      // metadata indexer for regular interfaces
 }
 
 // Address describes the XML coding of the pci addres of an interface in libvirt
 type Address struct {
-	Type     string `xml:"type,attr"`
+	Type     string `xml:"type,attr,omitempty"`
 	Domain   string `xml:"domain,attr"`
 	Bus      string `xml:"bus,attr"`
 	Slot     string `xml:"slot,attr"`
 	Function string `xml:"function,attr"`
 }
 
+type Source struct {
+	Address *Address `xml:"address"`
+}
+
 // DomainStateMap stringifies the state of a domain
-var DomainStateMap map[libvirtgo.DomainState]string = map[libvirtgo.DomainState]string{
+var DomainStateMap = map[libvirtgo.DomainState]string{
 	libvirtgo.DOMAIN_NOSTATE:     "UNDEFINED",
 	libvirtgo.DOMAIN_RUNNING:     "UP",
 	libvirtgo.DOMAIN_BLOCKED:     "BLOCKED",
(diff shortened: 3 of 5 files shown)


pierrecregut:
The current version does not even create the VF nodes. Everything is put as metadata in the node of the interface. We need VF nodes to distinguish what is associated to each libvirt domain.

lebauce:
> The current version does not even create the VF nodes.

The VF nodes should reported as regular nodes by the netlink probes. The only thing that we do for SRIO-V right now is to create a `vf` link between the PV and its VFs.

pierrecregut:
@lebauce I report the PCI address for all interfaces. I will probably change the name BusInfo to something less misleading and yes I use an indexer. 
I must distinguish the VF control as seen from the PF and the interface itself. I could try to push the VF control info in the interface but it will not work with libvirt. If you use it with libvirt, you do not even load the device driver for the virtual interface (should be blacklisted).

pierrecregut:
I have added the libvirt part. I treat the two cases I have tried: declaring a hostdev or declaring an interface of type hostdev. The second provides access to the control function of the virtual interface and so is the preferred approach. I have had limited success with the vf declared as a pool in a libvirt network. It may change what is available in libvirt.

pierrecregut:
This is an alternative design.The old one is available on Orange-opensource as sriov-old. I am back to the old design where there is no node per VF but the virtual interface is annotated with a VfID. For libvirt I create a device for each card using the passthrough. This is the card that will be referenced by the peering probe. Note that this design should also work for physical cards handled to a VM.

pierrecregut:
Here is a screenshot of the new design. Note the VfID but the parameters of the VF are in enp9s0f2 node (the physical device)
sriov-new
And the old design:
sriov-old

pierrecregut:
run tests

pierrecregut:

pierrecregut:
@lebauce  @safchain : So this new version fails the sriov test. The reason is that it does not only test that it reads correctly a sriov configuration but that it can UPDATE a sriov config as the vfs are created on the fly. On the other hand, it does not test that the Vfs array does exist in the physical interface, otherwise it would fail even with the old implementation. The point is that when sriov_numvfs is changed a netlink message is sent immediately, but the Vfs structure is not updated and will not be updated before the VF are created. If we are in a classical setting for libvirt, the module for VF will be blacklisted in the host. There will be no interface to capture, so no message to read and the VF structure will remain empty.

A quick patch would be to add back the old way to create the link. The test pass, but the problem remains. I have found only one solution to solve it but not necessarily pretty and a little hard to implement. The only value up to date is sriov_numvfs in the sys/bus/pci hierarchy. As long as the number of vfs in link.Attrs is different from the sriov_numvfs value, we should poll link.Attrs() and may be the sriov_numvfs file until they give back the same value.

What do you think ?

pierrecregut:
run skydive-functional-tests-ebpf

pierrecregut:

topology/probes/libvirt/libvirt.go
```diff
@@ -167,13 +283,15 @@ func (probe *Probe) createOrUpdateDomain(d *libvirtgo.Domain) *graph.Node {
 	g := probe.graph
 	g.Lock()
 	defer g.Unlock()
-	domainName, _ := d.GetName()
+	domainName, err := d.GetName()
+	if err != nil {
+		return nil
```
nplanel: same here

nplanel:

nplanel:

pierrecregut:
run skydive-functional-tests-ebpf

pierrecregut:

pierrecregut:
This new implementation implements all sriov related functions in a separate goroutine. The first step is to ensure that we are in a stable state. When numvfs is changed, its value is not immediately reported in the /sys filesystem although a netlink message is immediately  triggered (it is not clear why there is no lock in the kernel on the pci device when numvfs is read in the /sys filesystem). Furthermore there may be a difference between vf information retrieved by ethtool and numvfs. As the node handling the physical function knows the number of virtual functions, it can register actions to create links with the nodes of the virtual functions.

pierrecregut:

lebauce:

lebauce:

lebauce:
run skydive-k8s-tests

lebauce:

pierrecregut:
@lebauce We need to merge #1710 and rebase this PR first.

pierrecregut:

pierrecregut:

pierrecregut:

pierrecregut:

pierrecregut:
I have just rebase to solve the conflict introduced by #1725 but in fact I just eliminated the code because the solution proposed creates the link from the Phy function not from the virtual ones. I am not surprised that you need to retry a few times before the structure stabilize. There is a similar problem with the value of sriov_numvfs. I submitted a bugreport and proposed a patch so eventually we will be able to simplify the code for new kernels : https://bugzilla.kernel.org/show_bug.cgi?id=202991

pierrecregut:

pierrecregut:

pierrecregut:
run skydive-k8s-tests

topology/probes/netlink/sriov.go
```diff
@@ -0,0 +1,208 @@
...
+	// Yes we have to wait. May be considered as a race bug in Linux kernel
+	numVfsFile := fmt.Sprintf(
+		"/sys/bus/pci/devices/%s/sriov_numvfs", businfo)
+
+	if _, err = os.Stat(numVfsFile); os.IsNotExist(err) {
+		// There is no VFS to monitor.
+		return
+	}
+	count, numVfs := 0, 0
+	for numVfs == 0 && count < 5 {
```
lebauce: nit: use common.Retry

topology/probes/netlink/sriov.go
```diff
@@ -0,0 +1,208 @@
...
+		if err != nil {
+			logging.GetLogger().Errorf(
+				"SR-IOV: cannot get numvfs of PCI - %s", err)
+			return
+		}
+		count = count + 1
+	}
+
+	count = 0
+	for len(attrsVfs) != numVfs && count < 10 {
```
lebauce: nit: use common.Retry

topology/probes/netlink/sriov.go
pierrecregut: I could in both case but it is a common.Retry with an additional predicate P. So I need to transform ~P in an error and then filter out at the end this ~P because it is not exactly an error but I have finally found what I was looking for. I may be wrong but I do not think I know how to have simpler code with common.Retry. May be I missed something.

topology/probes/netlink/sriov.go
pierrecregut: Same as above but you may say that if both numbers do not converge there is a real problem and may be we should abort. In that case we are in a common retry pattern with `len(attrVfs) != numVfs` considered as an error. Not sure about this one.

pierrecregut:
I am pushing an updated set of commits. The main changes are:
* Try to get some consistent names for hostdev interfaces
* Check better that we had something in the XML (tests for nil) for some attributes
* A new commit that solves a general netlink problem (not linked to sriov): interfaces can be renamed especially with "consistent device naming". The solution is to use the ifindex instead of names. Note that even with this, if you create 64 or 128 VF as you can on an X710, you will probably have netlink errors because the message buffers are too small.

There is still a small UI problem, you must reload the UI to see the name change in the graph (node names). The change is visible in the metadata window without reload.

topology/probes/netlink/sriov.go
pierrecregut: @lebauce May be a common function `RetryUnless(fnc func() (bool, error), try int, delay time.Duration) (bool, error)` ?

pierrecregut:

pierrecregut:

safchain:

tests/sriov_test.go
```diff
@@ -23,32 +23,37 @@ import (
 	"fmt"
 	"io/ioutil"
 	"testing"
+	"time"
+
+	"github.com/skydive-project/skydive/common"
 )
 
+var device = "enp5s0f0"
```
safchain: It seems we need to detect NIC being UP and supporting SRIO-V otherwise the test will fail

topology/probes/libvirt/libvirt.go
```diff
@@ -153,9 +243,35 @@ func (itf *Interface) ProcessNode(g *graph.Graph, node *graph.Node) bool {
...
+	}
+	return false
+}
+
 // getDomain access the graph node representing a libvirt domain
 func (probe *Probe) getDomain(d *libvirtgo.Domain) *graph.Node {
-	domainName, _ := d.GetName()
+	domainName, err := d.GetName()
+	if err != nil {
+		return nil
```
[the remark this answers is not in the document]
pierrecregut: done

topology/probes/netlink/sriov.go
pierrecregut: Finally used the common.Retry pattern and introduced a local error that is catched for numVfs=0

topology/probes/netlink/sriov.go
pierrecregut: The reason was that I was afraid numVfs was updated continuously as vfs were created (the reason of the asynchronous behaviour was not clear). This cannot happen (see the patch in the kernel to change the order of events, in any case (with or without) there is a single assignment of numVfs.

graffiti/graph/processor.go
```diff
@@ -96,25 +96,33 @@ func (processor *Processor) Cancel(values ...interface{}) {
...
+
 // OnNodeAdded event
 func (processor *Processor) OnNodeAdded(n *Node) {
 	if vValues, err := getFieldsAsArray(n, processor.MetadataIndexer.indexes); err == nil {
 		for _, values := range vValues {
 			hash := Hash(values...)
 			processor.Lock()
-			defer processor.Unlock()
-			if actions, ok := processor.actions[hash]; ok {
+			actions, ok := processor.actions[hash]
```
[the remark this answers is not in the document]
pierrecregut: done.

tests/sriov_test.go
```diff
@@ -23,32 +23,37 @@ import (
...
-				return err
-			}
-
-			if len(nodes) != 2 {
-				return fmt.Errorf("Expected 2 nodes, got %+v", nodes)
-			}
-
-			return nil
+			gremlin := c.gremlin.V().Has("Name", device).OutE("RelationType", "vf").OutV()
+			return common.Retry(func() error {
```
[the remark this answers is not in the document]
pierrecregut: done.

tests/sriov_test.go
pierrecregut: implemented as findSRIOV as discussed on irc: check for sriov_numvfs as a proof it is an sriov interface and check for operstate == up.

pierrecregut:

safchain:
I tested the libvirt part, I had few issues:

- shutdown / start domain recreate the `hostdev` interface each time, so duplicate interfaces
- I had once two difference interfaces for the `VF` interface, the `hostdev` attached only to the VM and and an `itf-1` interface attached to the `PF`. I had this one time, haven't been able to reproduce.

pierrecregut:
@safchain yes indeed. And there is only one good solution I see to both problems: rely more on pci address as unique identifier of the hostdev. It will also capture the case where a libvirt domain is declared with an host pci device but the device is currently bound in the host.

pierrecregut:
@safchain The new solution use the BDF part of the pci address as name. The node is only created if it is new. I try not to create a new node when there is an interface already in the host. It should be removed by libvirt and replaced when the domain is started (if the interface is declared as managed).

pierrecregut:

topology/probes/libvirt/libvirt.go
```diff
@@ -106,45 +131,122 @@ func (probe *Probe) getDomainInterfaces(
...
+		err = errors.New("hostdev without PCI address")
+		return
+	}
+	address := formatPciAddress(source.Address)
+	node = probe.graph.LookupFirstChild(probe.root, graph.Metadata{
+		"BusInfo": address,
+	})
+	if node == nil {
+		m := graph.Metadata{
+			"Name":    fmt.Sprintf("%s-%s", name, strings.TrimPrefix(address, "0000:")),
```
safchain: It could be nice to ass the Alias if present

topology/probes/libvirt/libvirt.go
```diff
@@ -106,45 +131,122 @@ func (probe *Probe) getDomainInterfaces(
...
+			"BusInfo": address,
+			"Driver":  "pci-passthrough",
```
safchain: You added the driver field in the XML parser, could we use it here ?

topology/probes/libvirt/libvirt.go
pierrecregut: The problem is that libvirt creates some alias on the fly when the domain is really started. You have a name that changes... Note that we should fix the UI at some point as it does not reflect dynamically changing names (true also for interfaces).

topology/probes/libvirt/libvirt.go
pierrecregut: no: the driver field is used when the libvirt presents an abstraction of the device to the VM. Here it just gives a pci descriptor after eventual translation of the BDF (that is why you still have a pci address that is often different from the source address).  In my case the driver would be i40evf but libvirt does not even know this.

topology/probes/libvirt/libvirt.go
safchain: My point was to add the Alias if present not to use it as name, but we can address this in a following PR.

safchain:
@pierrecregut thanks for addressing the comments and the patch

topology/probes/libvirt/libvirt.go
pierrecregut: makeHostDev is the core for HostDev nodes. Then as other libvirt interfaces they are enriched either by interface.ProcessNode if they were declared as interface or by enrichHostdev for raw host device. In both cases, if Alias is defined, it will be added. Did you find a path where it was not used ?

safchain: my bad I didn't see the Alias thing in the enrichHostdev function. Thx

PR #1709, opened 2019-03-18
ci: bug in vpp_test

pierrecregut:
Typo in vpp_tests breaks ci (and useless package import).

### DIFF
--- a/tests/vpp_tests.go
+++ b/tests/vpp_tests.go
@@ -25,8 +25,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pierrecregut:
@lebauce @nplanel  Could you have a look at this quick patch ? I think it is necessary if we want to run functional-hw-tests on the ci (needed by my sriov pull request).

Not sure why VPPPollingTime is used as a constant here anyway. May be the dependency could be removed with another constant defined in the test. If the goal is to have at least an iteration, this is not hard real time and a little more time would be safer.

PR #1715, opened 2019-03-19
ui: fix bandwidth not showing for links on selected nodes

eranra:


statics/js/components/topology.js
```diff
@@ -896,17 +896,18 @@ var TopologyComponent = {
     },
 
     normalizeMetric: function(metric) {
-      if (metric.Start && metric.Last && (metric.Last - metric.Start) > 0) {
-        bps = Math.floor(1000 * 8 * ((metric.RxBytes || 0) + (metric.TxBytes || 0)) / (metric.Last - metric.Start));
-        metric["Bandwidth"] = bandwidthToString(bps);
+      var metriccopy = JSON.parse(JSON.stringify(metric));
```
masco-enovance: s/metriccopy/metricCopy

statics/js/components/topology.js
```diff
@@ -896,17 +896,18 @@ var TopologyComponent = {
...
 
       ['Start', 'Last'].forEach(function(k) {
-        if (metric[k] && typeof metric[k] === 'number') {
-          metric[k] = new Date(metric[k]).toLocaleTimeString();
+        if (metricCopy[k] && typeof metricCopy[k] === 'number') {
+          metricCopy[k] = new Date(metricCopy[k]).toLocaleTimeString();
         }
       });
-      return metric;
+      return metricCopy;
```
hunchback: Just return copy reverting all other changes:

```
return JSON.parse(JSON.stringify(metric));
```

A more performant option can be found here:

```
https://github.com/kolodny/immutability-helper/blob/master/README.md
```

eranra: @hunchback thanks for looking on this. metricCopy includes JSON.parse(JSON.stringify(metric)) and all changes, we need to return the version of metric that includes the changes.

 JSON.parse(JSON.stringify) is the common way to do deep copy in js (look on extractMetadata #line 913  and on https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript). You can change and enhance as you like post acceptance of the PR, with pleasure. One request if you go that path, make sure you verify that there is value in that change, that there is extra performance, this is less common so will be harder to maintain.

PR #1720, opened 2019-03-20
k8s: ehnance configmap and secret

hunchback:
Enhanced secret / configMap:
- adding links from pod
- adding metadata
- extending the wordpress.sh demo
- adding support script to restart dependent (on a configuration object) pods

You can see all working together here: https://youtu.be/9lfOYM_-S2o


lebauce:
@hunchback Why are the values changed to booleans ?

hunchback:
https://github.com/skydive-project/skydive/pull/1720#issuecomment-475562487

The data map is converted into  string to bool map as:
- in the case of secrets the value is an obfuscated byte array (so no point indexing)
- in the case of configmap the value may be a very long string - such as the contents of a bash script - this is simply too much infomation to index.

hunchback:

topology/probes/k8s/configmap.go
```diff
@@ -37,6 +37,7 @@ func (h *configMapHandler) Dump(obj interface{}) string {
 func (h *configMapHandler) Map(obj interface{}) (graph.Identifier, graph.Metadata) {
 	cm := obj.(*v1.ConfigMap)
 	m := NewMetadataFields(&cm.ObjectMeta)
+	m.SetField("Data", mapStringToBool(cm.Data))
```
safchain: Just a question, why a map with all the value set as `true` ? why not a list in that case ?

hunchback: so the reason for the map (vs. list) is that when using a make you can query:

```
g.V().HasKey("K8s.Data.<keyname>") 
```

and reason for the map pointing to `true` is that the alternative would be pointing to an empty string ... but this may be misleading in the case the the actual value could be an empty string.

lebauce: @hunchback If it's a list, you can also do g.V().Has("K8s.Data.<keyname>") so a list is probably better than a map of true values

hunchback: Ok, will change to a list, and use following expression:

```
g.V().Has("K8s.Data", "<keyname>")
```

hunchback: Fixed:

image

hunchback:

contrib/kubernetes/skydive.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+PROG=skydive.sh
```
safchain: This should be in a `samples` folder with a README or any explanation

vendor/vendor.json
```diff
@@ -2387,7 +2387,7 @@
 			"revision": "604eaf189ee867d8c147fafc28def2394e878d25"
 		},
 		{
-			"checksumSHA1": "rFSw9cpg4vxtZD7dlOtKx25h4kI=",
+			"checksumSHA1": "2jmj7l5rSw0yVb/vlWAYkK/YBwk=",
```
lebauce: With the latest version of govendor, I get a `checksumSHA1` of `rFSw9cpg4vxtZD7dlOtKx25h4kI=`, not `2jmj7l5rSw0yVb/vlWAYkK/YBwk=`

hunchback: done

lebauce:

lebauce:

lebauce:
run skydive-k8s-tests

hunchback:
Can this be merged ?

hunchback:
run tests

hunchback:

hunchback:
run skydive-ppc64le-tests

PR #1722, opened 2019-03-21
tests: fix functional tests in static mode

lebauce:
skip skydive-go-fmt
skip skydive-functional-tests-backend-elasticsearch
skip skydive-scale-tests
skip skydive-cdd-overview-tests
skip skydive-unit-tests
skip skydive-compile-tests
skip skydive-k8s-tests
skip skydive-ppc64le-tests
skip skydive-functional-hw-tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -419,9 +419,9 @@ test.functionals.compile: govendor genlocalfiles
 .PHONY: test.functionals.static
 test.functionals.static: govendor genlocalfiles
 	$(GOVENDOR) test -tags "netgo ${BUILD_TAGS} test" \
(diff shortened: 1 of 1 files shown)


lebauce:

lebauce:
@nlewo It seems we are missing the `libnl` library. Where this dependency should be added ?

nlewo:
@lebauce I think there: 
- https://github.com/skydive-project/skydive/blob/master/Makefile#L105
- https://github.com/skydive-project/skydive/blob/master/Makefile#L114
If this doesn't work, I could give a try on next week.

lebauce:
@nlewo I thought that libnl had to be installed using NixOS but I don't know how to do this

nlewo:
No, I don't think so. If I remember correctly, the binary is statically compiled on the Jenkins host, without relying on any Nix stuffs.

PR #1758, opened 2019-04-03
flow: eBPF IP/GRE support

nplanel:
skip skydive-k8s-tests
skip skydive-scale-tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -414,9 +414,9 @@ dpdk.clean:
 ebpf.build:
 ifeq ($(WITH_EBPF), true)
 ifeq ($(WITH_EBPF_DOCKER_BUILDER), true)
-	$(MAKE) -C probe/ebpf docker-ebpf-build
+	$(MAKE) -C probe/ebpf clean docker-ebpf-build
 else
-	$(MAKE) -C probe/ebpf ebpf-build
+	$(MAKE) -C probe/ebpf clean ebpf-build
 endif
 endif
 
--- a/flow/probes/ebpf.go
+++ b/flow/probes/ebpf.go
@@ -33,8 +33,6 @@ import (
 	"time"
 	"unsafe"
 
-	"github.com/google/gopacket"
-	"github.com/google/gopacket/layers"
 	"github.com/iovisor/gobpf/elf"
 
 	"github.com/skydive-project/skydive/api/types"
@@ -51,6 +49,7 @@ import (
 import "C"
 
 const (
+	BPF_ANY    = 0
 	ebpfUpdate = 2 * time.Second
 )
 
@@ -81,6 +80,12 @@ type EBPFProbesHandler struct {
 	wg         sync.WaitGroup
 }
 
+func kernFlowKeyOuter(kernFlow *C.struct_flow) string {
+	hasher := sha1.New()
+	hasher.Write(C.GoBytes(unsafe.Pointer(&kernFlow.key_outer), C.sizeof___u64))
+	return hex.EncodeToString(hasher.Sum(nil))
+}
+
 func kernFlowKey(kernFlow *C.struct_flow) string {
 	hasher := sha1.New()
 	hasher.Write(C.GoBytes(unsafe.Pointer(&kernFlow.key), C.sizeof___u64))
@@ -94,35 +99,119 @@ func tcpFlagTime(currFlagTime C.__u64, startKTimeNs int64, start time.Time) int6
 	return common.UnixMillis(start.Add(time.Duration(int64(currFlagTime) - startKTimeNs)))
 }
 
-func (p *EBPFProbe) newFlowOperation(ebpfFlow *EBPFFlow, kernFlow *C.struct_flow, startKTimeNs int64, start time.Time) *flow.Operation {
+func kernLayersPath(kernFlow *C.struct_flow) (layersPath string, hasGRE bool) {
+	notFirst := false
+	path := uint64(kernFlow.layers_path)
+	for i := C.LAYERS_PATH_LEN - 1; i >= 0; i-- {
+		layer := (path & (uint64(C.LAYERS_PATH_MASK) << uint(i*C.LAYERS_PATH_SHIFT))) >> uint(i*C.LAYERS_PATH_SHIFT)
+		if layer == 0 {
+			continue
+		}
+		if notFirst {
+			layersPath += "/"
+		}
+		notFirst = true
+		switch layer {
+		case C.ETH_LAYER:
+			layersPath += "Ethernet"
+		case C.ARP_LAYER:
+			layersPath += "ARP"
+		case C.DOT1Q_LAYER:
+			layersPath += "Dot1Q"
+		case C.IP4_LAYER:
+			layersPath += "IPv4"
+		case C.IP6_LAYER:
+			layersPath += "IPv6"
+		case C.ICMP4_LAYER:
+			layersPath += "ICMPv4"
+		case C.ICMP6_LAYER:
+			layersPath += "ICMPv6"
+		case C.UDP_LAYER:
+			layersPath += "UDP"
+		case C.TCP_LAYER:
+			layersPath += "TCP"
+		case C.SCTP_LAYER:
+			layersPath += "SCTP"
+		case C.GRE_LAYER:
+			hasGRE = true
+			layersPath += "GRE"
+		default:
+			layersPath += "Unknown"
+		}
+	}
+	return layersPath, hasGRE
+}
+
+func (p *EBPFProbe) newFlowOperation(ebpfFlow *EBPFFlow, kernFlow *C.struct_flow, startKTimeNs int64, start time.Time) []*flow.Operation {
+
+	var fops []*flow.Operation
 	f := flow.NewFlow()
 	f.Init(common.UnixMillis(ebpfFlow.start), p.probeNodeTID, flow.UUIDs{})
 	f.Last = common.UnixMillis(ebpfFlow.last)
 
-	layersFlag := uint8(kernFlow.layers)
+	layersInfo := uint8(kernFlow.layers_info)
 
 	// LINK
-	linkA := C.GoBytes(unsafe.Pointer(&kernFlow.link_layer.mac_src[0]), C.ETH_ALEN)
-	linkB := C.GoBytes(unsafe.Pointer(&kernFlow.link_layer.mac_dst[0]), C.ETH_ALEN)
-	f.Link = &flow.FlowLayer{
-		Protocol: flow.FlowProtocol_ETHERNET,
-		A:        net.HardwareAddr(linkA).String(),
-		B:        net.HardwareAddr(linkB).String(),
-		ID:       int64(kernFlow.link_layer.id),
-	}
-	f.LayersPath = "Ethernet"
-
-	nVLAN := uint8(kernFlow.link_layer.vlans)
-	for i := uint8(0); i < nVLAN; i++ {
-		f.LayersPath += "/Dot1Q"
+	if layersInfo&uint8(C.LINK_LAYER_INFO) > 0 {
+		linkA := C.GoBytes(unsafe.Pointer(&kernFlow.link_layer.mac_src[0]), C.ETH_ALEN)
+		linkB := C.GoBytes(unsafe.Pointer(&kernFlow.link_layer.mac_dst[0]), C.ETH_ALEN)
+		f.Link = &flow.FlowLayer{
+			Protocol: flow.FlowProtocol_ETHERNET,
+			A:        net.HardwareAddr(linkA).String(),
+			B:        net.HardwareAddr(linkB).String(),
+			ID:       int64(kernFlow.link_layer.id),
+		}
 	}
 
-	if layersFlag&uint8(C.ARP_LAYER) > 0 {
-		f.LayersPath += "/ARP"
+	var hasGRE bool
+	f.LayersPath, hasGRE = kernLayersPath(kernFlow)
+
+	if hasGRE {
+		pathSplitAt := strings.Index(f.LayersPath, "/GRE/")
+		innerLayerPath := f.LayersPath[pathSplitAt+len("/GRE/"):]
+
+		parent := f
+		parent.LayersPath = f.LayersPath[:pathSplitAt] + "/GRE"
+
+		// NETWORK
+		if layersInfo&uint8(C.NETWORK_LAYER_INFO) > 0 {
+			protocol := uint16(kernFlow.network_layer_outer.protocol)
+
+			switch protocol {
+			case syscall.ETH_P_IP:
+				netA := C.GoBytes(unsafe.Pointer(&kernFlow.network_layer_outer.ip_src[12]), net.IPv4len)
+				netB := C.GoBytes(unsafe.Pointer(&kernFlow.network_layer_outer.ip_dst[12]), net.IPv4len)
+				parent.Network = &flow.FlowLayer{
+					Protocol: flow.FlowProtocol_IPV4,
+					A:        net.IP(netA).To4().String(),
+					B:        net.IP(netB).To4().String(),
+				}
+			case syscall.ETH_P_IPV6:
+				netA := C.GoBytes(unsafe.Pointer(&kernFlow.network_layer_outer.ip_src[0]), net.IPv6len)
+				netB := C.GoBytes(unsafe.Pointer(&kernFlow.network_layer_outer.ip_dst[0]), net.IPv6len)
+				parent.Network = &flow.FlowLayer{
+					Protocol: flow.FlowProtocol_IPV6,
+					A:        net.IP(netA).String(),
+					B:        net.IP(netB).String(),
+				}
+			}
+		}
+		key := kernFlowKeyOuter(kernFlow)
+		parent.UpdateUUID(key, flow.Opts{LayerKeyMode: flow.L3PreferedKeyMode})
+		fops = append(fops, &flow.Operation{
+			Type: flow.ReplaceOperation,
+			Flow: parent,
+			Key:  key,
+		})
+		// upper layer
+		f = flow.NewFlow()
+		f.Init(common.UnixMillis(ebpfFlow.start), p.probeNodeTID, flow.UUIDs{ParentUUID: parent.UUID})
+		f.Last = common.UnixMillis(ebpfFlow.last)
+		f.LayersPath = innerLayerPath
 	}
 
 	// NETWORK
-	if layersFlag&uint8(C.NETWORK_LAYER) > 0 {
+	if layersInfo&uint8(C.NETWORK_LAYER_INFO) > 0 {
 		protocol := uint16(kernFlow.network_layer.protocol)
 
 		switch protocol {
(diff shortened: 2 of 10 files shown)


nplanel:

nplanel:

nplanel:

nplanel:

nplanel:
run skydive-compile-tests

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

safchain:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:
run tests

nplanel:

nplanel:

nplanel:

nplanel:

nplanel:

probe/ebpf/flow.c
```diff
@@ -48,46 +78,117 @@ MAP(u64_config_values) {
...
-	//fill_payload_bucket(skb, offset, flow->payload, 30);
+//	bpf_skb_load_bytes(skb, offset, flow->payload, sizeof(flow->payload));
 }
 
-static inline void fill_transport(struct __sk_buff *skb, __u8 protocol, int offset,
-	struct flow *flow, int len, __u64 tm)
+static inline __u16 fill_gre(struct __sk_buff *skb, size_t *offset, struct flow *flow)
+{
+	__u8 config = load_byte(skb, *offset);
+	__u8 cfgChecksum = config & 0x80;
```
safchain: nit: let's keep everything in snake case

probe/ebpf/flow.c
```diff
@@ -48,46 +78,117 @@ MAP(u64_config_values) {
...
+	}
+	if (cfgKey > 0) {
+		(*offset) += 4;
+	}
+	if (cfgSeq > 0) {
+		(*offset) += 4;
+	}
+	if (cfgRouting > 0) {
+#pragma unroll
+		for(int i=0;i<MAX_GRE_ROUTING_INFO;i++ ) {
```
safchain: nit: format 

```
for(int i = 0; i < MAX_GRE_ROUTING_INFO; i++) {
```

probe/ebpf/flow.c
```diff
@@ -211,111 +322,69 @@ static inline void fill_ipv6(struct __sk_buff *skb, int offset, __u8 *dst, __u64
 	update_hash_word(hash, w);
 }
 
-static inline void fill_network(struct __sk_buff *skb, __u16 netproto, int offset,
-	struct flow *flow, __u64 tm)
+static inline void fill_network(struct __sk_buff *skb, __u16 netproto, size_t offset, int len,
+				struct flow *flow)
```
safchain: nit: fix indent

Makefile
```diff
@@ -414,9 +414,9 @@ dpdk.clean:
 ebpf.build:
 ifeq ($(WITH_EBPF), true)
 ifeq ($(WITH_EBPF_DOCKER_BUILDER), true)
-	$(MAKE) -C probe/ebpf docker-ebpf-build
+	$(MAKE) -C probe/ebpf clean docker-ebpf-build check
```
lebauce: Why is the `clean` required ?

flow/probes/ebpf.go
```diff
@@ -110,19 +162,52 @@ func (p *EBPFProbe) newFlowOperation(ebpfFlow *EBPFFlow, kernFlow *C.struct_flow
...
-	for i := uint8(0); i < nVLAN; i++ {
-		f.LayersPath += "/Dot1Q"
-	}
-
-	if layersFlag&uint8(C.ARP_LAYER) > 0 {
-		f.LayersPath += "/ARP"
+	f.LayersPath = kernLayersPath(kernFlow)
+
+	if strings.Contains(f.LayersPath, "GRE") {
+		layersPaths := strings.Split(f.LayersPath, "/GRE/")
```
lebauce: nit: use strings.Index(f.Layers.Path, "/GRE/") as it's faster and would handle the (very improbable) case where there are multiple GRE layers

flow/probes/ebpf.go
```diff
@@ -439,6 +519,7 @@ func (p *EBPFProbesHandler) registerProbe(n *graph.Node, capture *types.Capture,
 		if err := elf.DetachSocketFilter(socketFilter, fd); err != nil {
 			logging.GetLogger().Errorf("Unable to detach eBPF probe: %s", err)
 		}
+		rs.Close()
```
lebauce: This should also be done line 492 (for return fmt.Errorf....)

probe/ebpf-check/check_load_ebpf.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// #cgo CFLAGS: -I../../probe/ebpf
+// #include "flow.h"
+import "C"
+
+// jmpTable { JMP_NETWORK_LAYER }
+var jmpTable = []string{"socket_network_layer"}
+
+func loadJumpMap(module *elf.Module) {
```
lebauce: Why not simply export the loadJumpMap function in ebpf.go and call it from here instead of duplicating the function ?

probe/ebpf-check/check_load_ebpf.go
```diff
@@ -0,0 +1,134 @@
+/*
```
lebauce: Why not moving this to tests/flow_ebpf_test.go or probe/ebpf/ebpf_test.go ?

Makefile
lebauce: If I read correctly, the `check` uses `sudo`. I'm not sure it's a good idea to require root privileges during the build process

probe/ebpf/flow_network.c
```diff
@@ -0,0 +1,72 @@
+
```
lebauce: I understand this file is `#include`d but it probably still needs the Apache header

probe/ebpf/flow_network.c
```diff
@@ -0,0 +1,72 @@
...
+		flow->network_layer_1st = flow->network_layer;
+
+		layer->_hash = FNV_BASIS;
+		flow->key_1st = flow->key;
+		flow->key_1st ^= flow->network_layer._hash ^ IPPROTO_GRE;
+
+		add_layer(flow, GRE_LAYER);
+		netproto = fill_gre(skb, &offset, flow);
+		{
+#include "flow_network.c"
```
lebauce: Love this trick :-)

probe/ebpf/flow.h
```diff
@@ -82,17 +110,29 @@ struct flow_metrics {
...
-	__u8                   layers;
+	__u64                  layers_path;
 
 	// ifIndex of the interface seing the packet
 	__u32                  ifindex;
 
 	// layers information
+	__u8                   layers_info;
 	struct link_layer      link_layer;
+	struct network_layer   network_layer_1st;
```
lebauce: Wondering if `_outer` would be clearer that `_1st`

flow/probes/ebpf.go
```diff
@@ -94,12 +101,57 @@ func tcpFlagTime(currFlagTime C.__u64, startKTimeNs int64, start time.Time) int6
...
+
+func (p *EBPFProbe) newFlowOperation(ebpfFlow *EBPFFlow, kernFlow *C.struct_flow, startKTimeNs int64, start time.Time) []*flow.Operation {
+
+	var fops []*flow.Operation
 	f := flow.NewFlow()
 	f.Init(common.UnixMillis(ebpfFlow.start), p.probeNodeTID, flow.UUIDs{})
 	f.Last = common.UnixMillis(ebpfFlow.last)
 
-	layersFlag := uint8(kernFlow.layers)
+	layersFlag := uint8(kernFlow.layers_info)
```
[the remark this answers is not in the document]
nplanel: I use layers_info when as flags, when the associated {link,network,icmp}_layer fields in flow struct is set

flow/probes/ebpf.go
```diff
@@ -110,19 +162,52 @@ func (p *EBPFProbe) newFlowOperation(ebpfFlow *EBPFFlow, kernFlow *C.struct_flow
...
-	nVLAN := uint8(kernFlow.link_layer.vlans)
-	for i := uint8(0); i < nVLAN; i++ {
-		f.LayersPath += "/Dot1Q"
-	}
-
-	if layersFlag&uint8(C.ARP_LAYER) > 0 {
-		f.LayersPath += "/ARP"
+	f.LayersPath = kernLayersPath(kernFlow)
+
+	if strings.Contains(f.LayersPath, "GRE") {
```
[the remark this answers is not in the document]
nplanel: not really as we don't extract info from gre, but I will check the layers path

Makefile
nplanel: clean is used because (for example) if you compile once in normal and after docker-ebpf-build you the target will be done the first time, do docker build will not be done.
It's a specific case, but may we should remove the docker build

nplanel: yes I used sudo because the checker use raw packets socket
not really ideal but you will win lots of time with the checker if you can compile, load and run the ebpf code (sudo is need only for running)
I will do it in 2 pass (compile and test target)

flow/probes/ebpf.go
nplanel: golang implementation : 

func Contains(s, substr string) bool {
	return Index(s, substr) >= 0
}

but ok

lebauce: I was talking about the `Split`, not the `Contain`

flow/probes/ebpf.go
nplanel: good catch

Makefile
lebauce: This is also the case for the protobuf and easyjson autogenerated files. The problem fixed by using docker was much more severe : the module was successfully compiled but failed to load, even on a clean environment.

lebauce:
run skydive-packaging-tests

probe/ebpf-check/check_load_ebpf.go
nplanel: because it was written independently ... and ebpf.go need gotags +ebpf ...

probe/ebpf-check/check_load_ebpf.go
nplanel: you mean probe/ebpf-check/check_load_ebpf.go in probe/ebpf/check_load_ebpf.go ?  because we can't have C and golang (main) for 2 different purpose (don't remember exactly the behavior, but try you will see golang restriction :) )

Otherwise : 
 tests/flow_ebpf_test.go  is equivalent to  tests/flow_test.go

probe/ebpf/flow_network.c
nplanel: eBPF too :)

probe/ebpf/flow.h
nplanel: yeah the idea of 1st, it's the 1st layer on the path, but outer may be _cleaner_

lebauce: CleaNer, I don't know. But CleaRer is important :-)

nplanel:

nplanel:
run skydive-packaging-tests

nplanel:

nplanel:
run skydive-packaging-tests

probe/ebpf/flow_network.c
safchain: so this implementation if GRE will only work(which already great) with gre(gre/l3) interface and not gretap (gre/l2). The question is do we want to support all the layers in this PR, or do we want to address this later (new issue) ?

probe/ebpf/flow.c
```diff
@@ -397,47 +470,95 @@ int bpf_flow_table(struct __sk_buff *skb)
...
+		new->layers_info |= LINK_LAYER_INFO;
+	}
+
+	/* L3/L4/... */
+	fill_network(skb, protocol, l2->next_layer_offset, skb->len - l2->next_layer_offset, new);
+
+	flow = bpf_map_lookup_element(&flow_table, &new->key);
+	if (flow == NULL) {
+		/* New flow */
+		new->ifindex = skb->ifindex;
```
safchain: not related to this patch but doesn't seem to be used

Makefile
```diff
@@ -394,11 +394,13 @@ static: skydive.clean govendor genlocalfiles
 contribs.clean:
 	$(MAKE) -C contrib/snort clean
 	$(MAKE) -C contrib/objectstore clean
+	$(MAKE) -C contrib/ebpf-check clean
 
 .PHONY: contribs
 contribs:
 	$(MAKE) -C contrib/snort
 	$(MAKE) -C contrib/objectstore
+	$(MAKE) -C contrib/ebpf-check
```
safchain: It is a bit unusual to have "functional test" running while just building. Moreover we have to wait 2sec because of this each time we build. The test itself is not enough robust there is no guaranty that we will have packet within 2sec. I really think we should remove this. make test or test.functional does already the job

probe/ebpf/flow.c
```diff
@@ -342,15 +414,15 @@ static inline void fill_link(struct __sk_buff *skb, int offset, struct flow *flo
 	update_hash_half(&hash_dst, layer->mac_dst[2] << 8 | layer->mac_dst[3]);
 	update_hash_half(&hash_dst, layer->mac_dst[4] << 8 | layer->mac_dst[5]);
 
+	layer->id = load_half(skb, offsetof(struct ethhdr, h_proto));
```
safchain: seems to be incorrect, should be filled with the vlan id.

flow/probes/ebpf.go
```diff
@@ -94,12 +101,59 @@ func tcpFlagTime(currFlagTime C.__u64, startKTimeNs int64, start time.Time) int6
...
+		}
+		if notFirst {
+			layersPath += "/"
+		}
+		notFirst = true
+		switch layer {
+		case C.ETH_LAYER:
+			layersPath += "Ethernet"
+		case C.ARP_LAYER:
+			layersPath += "Arp"
```
safchain: ARP

We should use the gopacket string definition otherwise we will introduce difference between layer spelling (lower case)

probe/ebpf/flow.c
safchain: Also I'm seeing flows with only ethernet(no ARP, just broadcast) layer which seems to be incorrect.

contrib/ebpf-check/check_load_ebpf.go
```diff
@@ -0,0 +1,122 @@
...
+	"runtime"
+	"time"
+	"unsafe"
+
+	"github.com/iovisor/gobpf/elf"
+	"github.com/skydive-project/skydive/common"
+	"github.com/skydive-project/skydive/flow/probes"
+)
+
+// #cgo CFLAGS: -I../../probe/ebpf
```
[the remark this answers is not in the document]
nplanel: it's not a functional test, it's a test helper when develop ebpf
current functional test will test the load, and running ebpf probe already

flow/probes/ebpf.go
```diff
@@ -94,12 +101,59 @@ func tcpFlagTime(currFlagTime C.__u64, startKTimeNs int64, start time.Time) int6
 	return common.UnixMillis(start.Add(time.Duration(int64(currFlagTime) - startKTimeNs)))
 }
 
-func (p *EBPFProbe) newFlowOperation(ebpfFlow *EBPFFlow, kernFlow *C.struct_flow, startKTimeNs int64, start time.Time) *flow.Operation {
+func kernLayersPath(kernFlow *C.struct_flow) (layersPath string, hasGRE bool) {
+	hasGRE = false
+	notFirst := false
+	layersPath = ""
```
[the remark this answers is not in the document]
nplanel: yep old code

flow/probes/ebpf.go
```diff
@@ -110,19 +164,56 @@ func (p *EBPFProbe) newFlowOperation(ebpfFlow *EBPFFlow, kernFlow *C.struct_flow
...
-	}
-
-	if layersFlag&uint8(C.ARP_LAYER) > 0 {
-		f.LayersPath += "/ARP"
+	var hasGRE bool
+	f.LayersPath, hasGRE = kernLayersPath(kernFlow)
+
+	if hasGRE {
+		pathSplitAt := strings.Index(f.LayersPath, "/GRE/")
+		innerLayerPath := f.LayersPath
```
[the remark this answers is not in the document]
nplanel: nope f.LayersPath[pathSplitAt+len("/GRE/"):] will modify f.LayersPath, so can't be used to split each side ... that's why I using a copy

probe/ebpf/flow.c
nplanel: yes it's not used and it's always accessible via skb->ifindex for debug purpose .. removing it

contrib/ebpf-check/check_load_ebpf.go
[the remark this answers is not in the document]
safchain: mmm, you are testing that the ebpf is working no? looks like a functional test :) I would suggest to make this a real functional test. Then make; make test.functional TEST_PATTERN=EbpfLoad

See another comment, it is called each time we build and take 2sec at least

probe/ebpf/flow.c
nplanel: for GRE L2 it's another game (or any other L2 tunnel layer), it's a little bit more tricky and I think will be address by a tail_call of the inner packet

Makefile
nplanel: my attention was to build it only ... fixed

probe/ebpf/flow.c
nplanel: link->id is filled by ethernet layer and vlan layers

flow/probes/ebpf.go
nplanel: good catch

flow/probes/ebpf.go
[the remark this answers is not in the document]
safchain: slices don't modify orginal arrays. Anyway that is a nit

probe/ebpf/flow.c
safchain: Seeing wrong link->id as well on some flows

flow/probes/ebpf.go
```diff
@@ -489,6 +577,29 @@ func (p *EBPFProbesHandler) Stop() {
...
+	}
+	for i, sym := range jmpTable {
+		entry := module.SocketFilter(sym)
+		if entry == nil {
+			return fmt.Errorf("Symbol %s not found", sym)
+		}
+
+		index := uint32(i)
+		fd := uint32(entry.Fd())
+		err := module.UpdateElement(jmpTableMap, unsafe.Pointer(&index), unsafe.Pointer(&fd), BPF_ANY)
```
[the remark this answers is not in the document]
lebauce: I can't believe I missed this one. Shame on me !

Makefile
safchain: I don't see the point to build it. If it is a dev tool it should not be there, we all have our own dev tools. Functional tests make sense. Building a binary not being used I'm not sure.

flow/probes/ebpf.go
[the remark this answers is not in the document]
nplanel: agreed with your last comment, but got a strange behavior

nplanel:

nplanel:

PR #1765, opened 2019-04-08
[WIP] docker/runc: emit warnings (not error) in case does not exist

hunchback:
Mute logs (warning instead of errors) when docker or runC probes are created back capabilities are missing from host.

To test create a system which is missing either runc or docker and then run:

```
export SKYDIVE_AGENT_TOPOLOGY_PROBES="runc docker"
sudo -E $(which skydive) allinone
```

And verify that repeated log messages are a warning every 1s

### DIFF
--- a/topology/probes/docker/docker.go
+++ b/topology/probes/docker/docker.go
@@ -192,7 +192,8 @@ func (probe *Probe) connect() error {
 	defer probe.client.Close()
 
 	if _, err := probe.client.ServerVersion(context.Background()); err != nil {
-		logging.GetLogger().Errorf("Failed to connect to Docker daemon: %s", err)
+		// in case docker does not exist we emit warnings not and errors
+		logging.GetLogger().Warningf("Failed to connect to Docker daemon: %s", err)
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


lebauce:
run skydive-k8s-tests

lebauce:
@hunchback I'm not sure about this. The `usual` scenario (Kubernetes, ...) is that docker is started before Skydive. If the operator specified it wants the `docker` probe enabled, and if for some reason `docker` is not running or not reachable due to a wrong configuration in the Skydive configuration file, I think it makes sense to display an error. What probably doesn't make much sense is to display the error message every second though

lebauce:
Just to recap our discussion during the community meeting:
If Skydive is started when docker/runc is not started yet, we should display an "ERROR" message like "failed to connect to docker. We'll retry in the background every n seconds". When Docker is later available, we should display an "INFO" message like "successfully connected to Docker". If the connection to Docker is lost, we should do the same than above : print an error message and retry in the background.

hunchback:
@lebauce - I will close this for now - and create a new PR if/when handled

PR #1776, opened 2019-04-12
istio: fix virtualservice weight calculation

tomerlange:
This PR fixes the way virtualservices weights are being calculated.
Note that only the default weights (which appear in an unconditional routing rules) are taken into account.

### DIFF
--- a/topology/probes/istio/virtualservice.go
+++ b/topology/probes/istio/virtualservice.go
@@ -48,14 +48,17 @@ func newVirtualServiceProbe(client interface{}, g *graph.Graph) k8s.Subprobe {
 	return k8s.NewResourceCache(client.(*kiali.IstioClient).GetIstioNetworkingApi(), &kiali.VirtualService{}, "virtualservices", g, &virtualServiceHandler{})
 }
 
+type route struct {
+	Destination struct {
+		App     string `mapstructure:"host"`
+		Version string `mapstructure:"subset"`
+	} `mapstructure:"destination"`
+	Weight int `mapstructure:"weight"`
+}
+
 type rule struct {
-	Routes []struct {
-		Destination struct {
-			App     string `mapstructure:"host"`
-			Version string `mapstructure:"subset"`
-		} `mapstructure:"destination"`
-		Weight int `mapstructure:"weight"`
-	} `mapstructure:"route"`
+	Match  interface{} `mapstructure:"match"`
+	Routes []route     `mapstructure:"route"`
 }
 
 type virtualServiceSpec struct {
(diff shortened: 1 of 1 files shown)


tomerlange:

topology/probes/istio/virtualservice.go
```diff
@@ -88,20 +89,20 @@ func virtualServicePodMetadata(a, b interface{}, typeA, typeB, manager string) g
...
+		if http.Match != nil {
+			continue
+		}
 		for _, route := range http.Routes {
 			if route.Destination.App != pod.Labels["app"] {
 				continue
 			}
 			if route.Destination.Version == pod.Labels["version"] || route.Destination.Version == "" {
-				weight += route.Weight
+				m["weight"] = route.Weight
```
hunchback: I found the below paragraph here https://istio.io/docs/reference/config/networking/v1alpha3/virtual-service/:

```
REQUIRED. The proportion of traffic to be forwarded to the service version. (0-100). Sum of weights across destinations SHOULD BE == 100. If there is only one destination in a rule, the weight value is assumed to be 100.
```

Indicating you should set `m["weight"]` to `100` in the above case

topology/probes/istio/virtualservice.go
```diff
@@ -88,20 +89,20 @@ func virtualServicePodMetadata(a, b interface{}, typeA, typeB, manager string) g
 	if pod.Labels["app"] == "" {
 		return m
 	}
-	weight := 0
 	for _, http := range vsSpec.HTTP {
+		if http.Match != nil {
```
hunchback: Why do you skip when `http.Match` is set? I found examples in which both `match` and `routes` are set:

https://istio.io/docs/reference/config/networking/v1alpha3/virtual-service/

```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: reviews-route
  namespace: foo
spec:
  hosts:
  - reviews # interpreted as reviews.foo.svc.cluster.local
  http:
  - match:
    - uri:
        prefix: "/wpcatalog"
    - uri:
        prefix: "/consumercatalog"
    rewrite:
      uri: "/newcatalog"
    route:
    - destination:
        host: reviews # interpreted as reviews.foo.svc.cluster.local
        subset: v2
  - route:
    - destination:
        host: reviews # interpreted as reviews.foo.svc.cluster.local
        subset: v1
```

topology/probes/istio/virtualservice.go
tomerlange: DONE

topology/probes/istio/virtualservice.go
tomerlange: A virtualservice that is configured properly is ending with a default rule, which is intended for a traffic that doesn't match any former conditional rule. As we discussed before, presenting weights for conditional rules doesn't make sense as we don't know the traffic origin beforehand. Thus, we should verify that a rule is the default one before adding its weights to the metadata field.

hunchback: Ok - please add a comment stating the above (as this is not apparent from reading the code).

tomerlange: DONE

hunchback:
@tomerlange - please rebase

hunchback:

PR #1778, opened 2019-04-16
flow: move RTT to Metric in order to keep track of it.

safchain:
This patch moves the RTT value to the Metric struct
in order to keep multiple version of it in the datastore.
The goal is to be able to get its value with only one
packet injection. For example create a long ICMP packet
injection with the same ID.

It is also improve the algorithm to rely on protocols
to get the 2 first packets. It avoid to calculate
the RTT in the middle of the flow.


safchain:

safchain:

safchain:
run  skydive-functional-tests-backend-orientdb

safchain:

safchain:

PR #1798, opened 2019-05-10
make: fix proto clean

hunchback:
to test:

```
rm flow/layers/generated.proto
make clean
```

failed before will succeed now


masco-enovance:
The file is deleting for without this fix by just running ```make clean```

hunchback:
> The file is deleting for me without this fix by just running `make clean`

The problem is that if the file does not exist - in the current implementation the `make clean` will fail with exit 1 - as the `rm flow/layers/generated.proto` command within the make will fail if the file does not exist - solved by `rm -f flow/layers/generated.proto`

masco-enovance:
got it!

PR #1804, opened 2019-05-13
flow : use xxHash and eBPF fixies

andrekassis:
Hi guys,
as promised this commit includes a few changes. 
Flow table is now bounded - no memory leaks
Hash function was replaced - no collisions 
Only one copy of the flow table is kept in ebpf now

This is still the first commit. Still need to report the number of deleted flows so this is on our todo list.
Cheers

### DIFF
--- a/flow/ebpf.go
+++ b/flow/ebpf.go
@@ -1,0 +1,309 @@
+// +build linux
+
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy ofthe License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specificlanguage governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package flow
+
+import (
+	"encoding/hex"
+	"net"
+	"strings"
+	"syscall"
+	"time"
+	"unsafe"
+
+	"github.com/skydive-project/skydive/common"
+)
+
+// #cgo CFLAGS: -I../probe/ebpf
+// #include "flow.h"
+import "C"
+
+// EBPFFlow Wrapper type used for passing flows from probe to main agent routine
+type EBPFFlow struct {
+	start        time.Time
+	last         time.Time
+	kernFlow     *C.struct_flow
+	startKTimeNs int64
+	probeNodeTID string
+}
+
+// SetEBPFFlow initializes pre-allocated ebpfflow
+func SetEBPFFlow(ebpfFlow *EBPFFlow, start time.Time, last time.Time, kernFlow unsafe.Pointer, startKTimeNs int64, probeNodeTID string) {
+	ebpfFlow.start = start
+	ebpfFlow.last = last
+	ebpfFlow.kernFlow = (*C.struct_flow)(kernFlow)
+	ebpfFlow.startKTimeNs = startKTimeNs
+	ebpfFlow.probeNodeTID = probeNodeTID
+}
+
+//key has already been hashed by the module, just convert to string
+func kernFlowKey(kernFlow *C.struct_flow) string {
+	return hex.EncodeToString(C.GoBytes(unsafe.Pointer(&kernFlow.key), C.sizeof___u64))
+}
+
+func kernFlowKeyOuter(kernFlow *C.struct_flow) string {
+	return hex.EncodeToString(C.GoBytes(unsafe.Pointer(&kernFlow.key_outer), C.sizeof___u64))
+}
+
+func tcpFlagTime(currFlagTime C.__u64, startKTimeNs int64, start time.Time) int64 {
+	if currFlagTime == 0 {
+		return 0
+	}
+	return common.UnixMillis(start.Add(time.Duration(int64(currFlagTime) - startKTimeNs)))
+}
+
+func kernLayersPath(kernFlow *C.struct_flow) (layersPath string, hasGRE bool) {
+	notFirst := false
+	path := uint64(kernFlow.layers_path)
+	for i := C.LAYERS_PATH_LEN - 1; i >= 0; i-- {
+		layer := (path & (uint64(C.LAYERS_PATH_MASK) << uint(i*C.LAYERS_PATH_SHIFT))) >> uint(i*C.LAYERS_PATH_SHIFT)
+		if layer == 0 {
+			continue
+		}
+		if notFirst {
+			layersPath += "/"
+		}
+		notFirst = true
+		switch layer {
+		case C.ETH_LAYER:
+			layersPath += "Ethernet"
+		case C.ARP_LAYER:
+			layersPath += "ARP"
+		case C.DOT1Q_LAYER:
+			layersPath += "Dot1Q"
+		case C.IP4_LAYER:
+			layersPath += "IPv4"
+		case C.IP6_LAYER:
+			layersPath += "IPv6"
+		case C.ICMP4_LAYER:
+			layersPath += "ICMPv4"
+		case C.ICMP6_LAYER:
+			layersPath += "ICMPv6"
+		case C.UDP_LAYER:
+			layersPath += "UDP"
+		case C.TCP_LAYER:
+			layersPath += "TCP"
+		case C.SCTP_LAYER:
+			layersPath += "SCTP"
+		case C.GRE_LAYER:
+			hasGRE = true
+			layersPath += "GRE"
+		default:
+			layersPath += "Unknown"
+		}
+	}
+	return layersPath, hasGRE
+}
+
+func (ft *Table) newFlowFromEBPF(ebpfFlow *EBPFFlow, key string) ([]string, []*Flow) {
+	var flows []*Flow
+	var keys []string
+	f := NewFlow()
+	f.Init(common.UnixMillis(ebpfFlow.start), ebpfFlow.probeNodeTID, UUIDs{})
+	f.Last = common.UnixMillis(ebpfFlow.last)
+
+	layersInfo := uint8(ebpfFlow.kernFlow.layers_info)
+
+	// LINK
+	if layersInfo&uint8(C.LINK_LAYER_INFO) > 0 {
+		linkA := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.link_layer.mac_src[0]), C.ETH_ALEN)
+		linkB := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.link_layer.mac_dst[0]), C.ETH_ALEN)
+		f.Link = &FlowLayer{
+			Protocol: FlowProtocol_ETHERNET,
+			A:        net.HardwareAddr(linkA).String(),
+			B:        net.HardwareAddr(linkB).String(),
+			ID:       int64(ebpfFlow.kernFlow.link_layer.id),
+		}
+	}
+
+	var hasGRE bool
+	f.LayersPath, hasGRE = kernLayersPath(ebpfFlow.kernFlow)
+
+	if hasGRE {
+		pathSplitAt := strings.Index(f.LayersPath, "/GRE/")
+		innerLayerPath := f.LayersPath[pathSplitAt+len("/GRE/"):]
+
+		parent := f
+		parent.LayersPath = f.LayersPath[:pathSplitAt] + "/GRE"
+
+		// NETWORK
+		if layersInfo&uint8(C.NETWORK_LAYER_INFO) > 0 {
+			protocol := uint16(ebpfFlow.kernFlow.network_layer_outer.protocol)
+
+			switch protocol {
+			case syscall.ETH_P_IP:
+				netA := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.network_layer_outer.ip_src[12]), net.IPv4len)
+				netB := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.network_layer_outer.ip_dst[12]), net.IPv4len)
+				parent.Network = &FlowLayer{
+					Protocol: FlowProtocol_IPV4,
+					A:        net.IP(netA).To4().String(),
+					B:        net.IP(netB).To4().String(),
+				}
+			case syscall.ETH_P_IPV6:
+				netA := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.network_layer_outer.ip_src[0]), net.IPv6len)
+				netB := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.network_layer_outer.ip_dst[0]), net.IPv6len)
+				parent.Network = &FlowLayer{
+					Protocol: FlowProtocol_IPV6,
+					A:        net.IP(netA).String(),
+					B:        net.IP(netB).String(),
+				}
+			}
+		}
+		pkey := kernFlowKeyOuter(ebpfFlow.kernFlow)
+		parent.UpdateUUID(pkey, Opts{LayerKeyMode: L3PreferedKeyMode})
+		flows = append(flows, parent)
+		keys = append(keys, pkey)
+		// upper layer
+		f = NewFlow()
+		f.Init(common.UnixMillis(ebpfFlow.start), ebpfFlow.probeNodeTID, UUIDs{ParentUUID: parent.UUID})
+		f.Last = common.UnixMillis(ebpfFlow.last)
+		f.LayersPath = innerLayerPath
+	}
+
+	// NETWORK
+	if layersInfo&uint8(C.NETWORK_LAYER_INFO) > 0 {
+		protocol := uint16(ebpfFlow.kernFlow.network_layer.protocol)
+
+		switch protocol {
+		case syscall.ETH_P_IP:
+			netA := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.network_layer.ip_src[12]), net.IPv4len)
+			netB := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.network_layer.ip_dst[12]), net.IPv4len)
+			f.Network = &FlowLayer{
+				Protocol: FlowProtocol_IPV4,
+				A:        net.IP(netA).To4().String(),
+				B:        net.IP(netB).To4().String(),
+			}
+		case syscall.ETH_P_IPV6:
+			netA := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.network_layer.ip_src[0]), net.IPv6len)
+			netB := C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.network_layer.ip_dst[0]), net.IPv6len)
+			f.Network = &FlowLayer{
+				Protocol: FlowProtocol_IPV6,
+				A:        net.IP(netA).String(),
+				B:        net.IP(netB).String(),
+			}
+		}
+	}
+
+	// TRANSPORT
+	if layersInfo&uint8(C.TRANSPORT_LAYER_INFO) > 0 {
+		portA := int64(ebpfFlow.kernFlow.transport_layer.port_src)
+		portB := int64(ebpfFlow.kernFlow.transport_layer.port_dst)
+		protocol := uint8(ebpfFlow.kernFlow.transport_layer.protocol)
+
+		switch protocol {
+		case syscall.IPPROTO_UDP:
+			f.Transport = &TransportLayer{
+				Protocol: FlowProtocol_UDP,
+				A:        portA,
+				B:        portB,
+			}
+
+			/* disabled for now as no payload is sent
+			p := gopacket.NewPacket(C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.payload[0]), C.PAYLOAD_LENGTH), layers.LayerTypeUDP, gopacket.DecodeOptions{})
+			if p.Layer(gopacket.LayerTypeDecodeFailure) == nil {
+				path, app := LayersPath(p.Layers())
+				f.LayersPath += "/" + path
+				f.Application = app
+			}
+			*/
+		case syscall.IPPROTO_TCP:
+			f.Transport = &TransportLayer{
+				Protocol: FlowProtocol_TCP,
+				A:        portA,
+				B:        portB,
+			}
+			f.TCPMetric = &TCPMetric{
+				ABSynStart: tcpFlagTime(ebpfFlow.kernFlow.transport_layer.ab_syn, ebpfFlow.startKTimeNs, ebpfFlow.start),
+				BASynStart: tcpFlagTime(ebpfFlow.kernFlow.transport_layer.ba_syn, ebpfFlow.startKTimeNs, ebpfFlow.start),
+				ABFinStart: tcpFlagTime(ebpfFlow.kernFlow.transport_layer.ab_fin, ebpfFlow.startKTimeNs, ebpfFlow.start),
+				BAFinStart: tcpFlagTime(ebpfFlow.kernFlow.transport_layer.ba_fin, ebpfFlow.startKTimeNs, ebpfFlow.start),
+				ABRstStart: tcpFlagTime(ebpfFlow.kernFlow.transport_layer.ab_rst, ebpfFlow.startKTimeNs, ebpfFlow.start),
+				BARstStart: tcpFlagTime(ebpfFlow.kernFlow.transport_layer.ba_rst, ebpfFlow.startKTimeNs, ebpfFlow.start),
+			}
+			/* disabled for now as no payload is sent
+			p := gopacket.NewPacket(C.GoBytes(unsafe.Pointer(&ebpfFlow.kernFlow.payload[0]), C.PAYLOAD_LENGTH), layers.LayerTypeTCP, gopacket.DecodeOptions{})
+			if p.Layer(gopacket.LayerTypeDecodeFailure) == nil {
+				path, app := LayersPath(p.Layers())
+				f.LayersPath += "/" + path
+				f.Application = app
+			}
+			*/
+		}
+	}
+
+	// ICMP
+	if layersInfo&uint8(C.ICMP_LAYER_INFO) > 0 {
+		kind := uint8(ebpfFlow.kernFlow.icmp_layer.kind)
+		code := uint8(ebpfFlow.kernFlow.icmp_layer.code)
+		id := uint16(ebpfFlow.kernFlow.icmp_layer.id)
+
+		f.ICMP = &ICMPLayer{
+			Code: uint32(code),
+			ID:   uint32(id),
+		}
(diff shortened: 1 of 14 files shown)


andrekassis:
Oh, I added xxhash vi go get. need to also add this to the makefile
will do tomorrow

flow/flow.go
```diff
@@ -309,28 +310,43 @@ func (p *Packet) TransportFlow() (gopacket.Flow, error) {
 	return layer.TransportFlow(), nil
 }
 
+//Hash flow with custom func
+func Hash(f gopacket.Flow, hasher hash.Hash) {
+	if strings.Compare(string(f.Src().Raw()), string(f.Dst().Raw())) > 0 {
```
nplanel: may retrieve the the endpoints only once with src, dst := f.Endpoints()

flow/flow.go
```diff
@@ -309,28 +310,43 @@ func (p *Packet) TransportFlow() (gopacket.Flow, error) {
 	return layer.TransportFlow(), nil
 }
 
+//Hash flow with custom func
+func Hash(f gopacket.Flow, hasher hash.Hash) {
```
nplanel: on side note, you should propose a PR to gopacket to use xxHash  : github.com/google/gopacket/flows.go

flow/flow.go
```diff
@@ -309,28 +310,43 @@ func (p *Packet) TransportFlow() (gopacket.Flow, error) {
...
 func (p *Packet) Key(parentUUID string, opts Opts) string {
-	var uuid uint64
-
+	hasher := xxHash64.New(0)
 	// uses L2 is requested or if there is no network layer
 	if opts.LayerKeyMode == L2KeyMode || p.NetworkLayer() == nil {
 		if layer := p.LinkLayer(); layer != nil {
-			uuid ^= layer.LinkFlow().FastHash()
+			Hash(layer.LinkFlow(), hasher)
+			hasher.Write([]byte(strconv.FormatInt(linkID(p), 10)))
```
nplanel: use binary package to convert int64 to bytes
     b := make([]byte, 8)
    binary.BigEndian.PutUint64(b, linkID(p))

flow/flow.go
```diff
@@ -309,28 +310,43 @@ func (p *Packet) TransportFlow() (gopacket.Flow, error) {
...
+	if layer := p.TransportLayer(); layer != nil {
+		Hash(layer.TransportFlow(), hasher)
 	}
 	if af, err := p.ApplicationFlow(); err == nil {
-		uuid ^= af.FastHash()
+		Hash(af, hasher)
 	}
-
-	return parentUUID + strconv.FormatUint(uuid, 10)
+	return hex.EncodeToString(hasher.Sum(nil))
```
nplanel: you should keep parentUUID logic or you will break tunnel

flow/probes/ebpf.go
```diff
@@ -394,50 +109,28 @@ func (p *EBPFProbe) run() {
...
 				}
-			}
-
-			for k, v := range ebpfFlows {
-				if time.Now().Sub(v.last).Seconds() > p.expire.Seconds() {
-					delete(ebpfFlows, k)
+				flow.SetEBPFFlow(&ebpfFlows[i], us, last, unsafe.Pointer(&kernFlows[i]), startKTimeNs, p.probeNodeTID)
+				flowChanOperation <- &ebpfFlows[i]
+				nextAvailablePtr++
+				if nextAvailablePtr == 2001 {
```
nplanel: please set a constant like FLOW_POOL_SIZE
nextAvailablePtr is not used, should be used line 400
nextAvailablePtr = nextAvailablePtr++ % FLOW_POOL_SIZE

flow/table.go
```diff
@@ -113,29 +120,280 @@ func updateTCPFlagTime(prevFlagTime int64, currFlagTime int64) int64 {
 	return currFlagTime
 }
 
+func kernFlowKeyOuter(kernFlow *C.struct_flow) string {
```
nplanel: please keep eBPF probe code in the probe itself

flow/probes/ebpf.go
```diff
@@ -394,50 +109,28 @@ func (p *EBPFProbe) run() {
...
+				if us.After(now) {
+					us = now
 				}
-			}
-
-			for k, v := range ebpfFlows {
-				if time.Now().Sub(v.last).Seconds() > p.expire.Seconds() {
-					delete(ebpfFlows, k)
+				flow.SetEBPFFlow(&ebpfFlows[i], us, last, unsafe.Pointer(&kernFlows[i]), startKTimeNs, p.probeNodeTID)
+				flowChanOperation <- &ebpfFlows[i]
```
nplanel: not sure the allocation pool will really enhance something here or it need to match the chan size

flow/probes/ebpf.go
nplanel: you should test you got an potential overlap ( (nextAvailablePtr+1) % FLOW_POOL_SIZE == i)
meaning your pool is full

nplanel:
@andrekassis could you re-run your tests with latest fix PR #1803

flow/flow.go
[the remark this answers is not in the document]

flow/flow.go
andrekassis: will do, thanks;)

flow/flow.go

flow/flow.go

flow/probes/ebpf.go
andrekassis: done for the first issue.
As for the second, there shouldn't be any overlaps since the channel is FIFO and blocks when it's full (after 1000) the 2001 is merely for safety. Am I correct?

flow/table.go
andrekassis: EBPFFlow is now part of the table isn't it?

flow/probes/ebpf.go
andrekassis: extra size is for safety. Althouth the channel is guaranteed to be fifi, when using a size matching that of the channel (i.e. 1000) there was indeed an overlap. maybe because reads happen in bulks, so that when the channel is full it is drained on a lockstep (perhaps for performance gain). so using 2001 prevents that from happening. (tested)

gremlin/traversal/cflows.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"math/rand"
+	"net"
+	"time"
+	"unsafe"
+
+	"github.com/skydive-project/skydive/flow"
+)
+
+func newICMPFlow(id uint32, nodeTID string, linkSrc string, linkDst string) *flow.EBPFFlow {
```
safchain: can you move this in the test file as it is only used for the tests

flow/table.go
```diff
@@ -268,22 +531,24 @@ func (ft *Table) update(updateFrom, updateTime int64) {
 	logging.GetLogger().Debugf("flow table update: %d, %d", updateFrom, updateTime)
 
 	var updatedFlows []*Flow
-	for k, f := range ft.table {
+	for _, k := range ft.table.Keys() {
```
safchain: This will create a copy of all the keys. It would be nice to have an implementation allowing to iterate over the values/keys

flow/table.go
```diff
@@ -113,29 +120,280 @@ func updateTCPFlagTime(prevFlagTime int64, currFlagTime int64) int64 {
...
+			layersPath += "GRE"
+		default:
+			layersPath += "Unknown"
+		}
+	}
+	return layersPath, hasGRE
+}
+
+func (ft *Table) newFlowFromEBPF(ebpfFlow *EBPFFlow, key string) *Flow {
+
```
safchain: nit: remove extra line

flow/table.go
```diff
@@ -113,29 +120,280 @@ func updateTCPFlagTime(prevFlagTime int64, currFlagTime int64) int64 {
...
+				parent.Network = &FlowLayer{
+					Protocol: FlowProtocol_IPV6,
+					A:        net.IP(netA).String(),
+					B:        net.IP(netB).String(),
+				}
+			}
+		}
+		pkey := kernFlowKeyOuter(ebpfFlow.kernFlow)
+		parent.UpdateUUID(pkey, Opts{LayerKeyMode: L3PreferedKeyMode})
+		ft.table.Add(pkey, parent)
```
safchain: I would prefer to have either both flows Added in this function or both returned, but here the parent is inserted here and the other flow in the caller

flow/probes/ebpf.go
```diff
@@ -368,10 +81,13 @@ func (p *EBPFProbe) run() {
 	var startKTimeNs int64
 	var start time.Time
 
-	ebpfFlows := make(map[C.__u64]*EBPFFlow)
 	updateTicker := time.NewTicker(ebpfUpdate)
 	defer updateTicker.Stop()
 
+	kernFlows := [2001]C.struct_flow{}
```
safchain: this size should be related to the chan size. I'm afraid that one day we will chan the chan size of this one without changing the other

flow/table.go
```diff
@@ -43,6 +52,21 @@ type Handler struct {
 	callback ExpireUpdateFunc
 	every    time.Duration
 }
+type EBPFFlow struct {
```
safchain: can you move all the ebpf related functions into a dedicated file, it will be easier to read

gremlin/traversal/cflows.go
andrekassis: The reason this funtion is in a seperate file is that cgo is not allowed in test files. i.e. files ending with _test.go
so this is my work around for this. If you have a better idea would love to hear it;)

andrekassis:
run 
skydive-compile-tests

flow/table.go
```diff
@@ -493,8 +755,8 @@ func (ft *Table) Run() {
 			}
 		case ps := <-ft.packetSeqChan:
 			ft.processPacketSeq(ps)
-		case op := <-ft.flowChanOperation:
-			ft.processFlowOP(op)
+		case ebpfFlow := <-ft.flowEBPFChan:
```
safchain: please do not remove the flowOpChan as it will be used for other probes

gremlin/traversal/cflows.go
safchain: That is indeed a good reason :)

gremlin/traversal/flow_test.go
```diff
@@ -132,8 +122,8 @@ func TestHasStep(t *testing.T) {
 		time.Sleep(100 * time.Millisecond)
 	}
 
-	flowChan <- &flow.Operation{Type: flow.ReplaceOperation, Flow: newICMPFlow(222), Key: strconv.Itoa(rand.Int())}
-	flowChan <- &flow.Operation{Type: flow.ReplaceOperation, Flow: newICMPFlow(444), Key: strconv.Itoa(rand.Int())}
+	flowChan <- newICMPFlow(222, "node1", "", "")
```
safchain: please do not change the test, please add epbf specific tests

flow/probes/ebpf.go
nplanel: old comment :
replace 2001 by 
p.flow_pool_size = cap(flowChanOperation)+1

flow/table.go
andrekassis: right now this library doesn't expose such a functionality, would you like me to override it? is it ok to do so (there are the licence issues or whatever). do you want us to open a PR?

flow/table.go
andrekassis: done

flow/table.go
andrekassis: done

flow/probes/ebpf.go
andrekassis: done

flow/table.go
andrekassis: done

flow/table.go
andrekassis: done

gremlin/traversal/flow_test.go
andrekassis: done

flow/ebpf.go
```diff
@@ -0,0 +1,290 @@
+package flow
```
nplanel: please add the license header

flow/probes/ebpf.go
```diff
@@ -53,22 +50,16 @@ const (
...
-	expire       time.Duration
-	quit         chan bool
+	probeNodeTID   string
+	fd             int
+	flowTable      *flow.Table
+	module         *elf.Module
+	fmap           *elf.Map
+	expire         time.Duration
+	quit           chan bool
+	flow_pool_size int
```
nplanel: keep the camelCase style flowPoolSize
Sorry, my fault, on my last comment ;)

flow/probes/ebpf.go
```diff
@@ -80,298 +71,24 @@ type EBPFProbesHandler struct {
...
 	defer p.flowTable.Stop()
 
 	var startKTimeNs int64
 	var start time.Time
 
-	ebpfFlows := make(map[C.__u64]*EBPFFlow)
 	updateTicker := time.NewTicker(ebpfUpdate)
 	defer updateTicker.Stop()
 
+	p.flow_pool_size = 2*cap(flowChanOperation) + 1
```
nplanel: why 2*cap ? the chan will block at cap so pool = cap+1 is enough

flow/ebpf.go
```diff
@@ -0,0 +1,290 @@
...
+	"syscall"
+	"time"
+	"unsafe"
+
+	"github.com/skydive-project/skydive/common"
+)
+
+// #cgo CFLAGS: -I../probe/ebpf
+// #include "flow.h"
+import "C"
```
nplanel: this is breaking windows cross compiling target
generate a dummy implentation of EBFFlow in no_ebpf.go with header // +build !linux

andrekassis: done

flow/probes/ebpf.go
andrekassis: done

flow/ebpf.go
andrekassis: done

flow/probes/ebpf.go
andrekassis: this one is a bit tricky. It seems that go keeps local read/write buffers to use with channels so that when the channel blocks after cap entries and the reader performs a read operation the entire content of the channel is copied to its local buffer (for performance) and then reusing the pointers by the writer results in an overlap (I myself experienced this in one of the stress tests). therefore the pool size should be twice the size of the channel ( I agree that +1 here is redundant, can remove it if you'd like)

andrekassis:

andrekassis:

flow/probes/ebpf.go
nplanel: golang channel use lockfree mechanism, but this should not be the double ... or I missing something ... do you have a bench test reproducing your issue (in case of cap+1) ?

nplanel: I found your issue and 2*cap()+1 will not fix it

Imaging you can push 3 elements in the chan
len(chan) is 3
next is 3
 ... now the chan reader side read the 3 elements
len(chan) is 0
next is still 3

so you are out of sync between the chan pointer and pointer that could be already referenced in another element in the chan.

so you should remove next and have another channel to send back, the freed pointers; at start the allocated pointers should be injected in the _freed_ channel.

i := getFreePointer()   // <- freeEbpfFlows
write in &ebpfFlows[i]
flowChanOperation <- &ebpfFlows[i]

flow/flow.go
```diff
@@ -309,28 +312,66 @@ func (p *Packet) TransportFlow() (gopacket.Flow, error) {
 	return layer.TransportFlow(), nil
 }
 
+//Hash flow with custom func, ignore ports in transport level for VXLAN and Geneve
+func Hash(f gopacket.Flow, hasher hash.Hash, isTransportUDP bool) {
+	src, dst := f.Endpoints()
+	if isTransportUDP {
+		srcPort, _ := strconv.ParseInt(src.String(), 10, 64)
+		dstPort, _ := strconv.ParseInt(dst.String(), 10, 64)
+		if srcPort == 4789 || srcPort == 6081 {
```
safchain: If you don't change the code line 368-369, you don't have to handle geneve nor vxlan here, IMHO

flow/flow.go
```diff
@@ -309,28 +312,66 @@ func (p *Packet) TransportFlow() (gopacket.Flow, error) {
...
+			b := make([]byte, 8)
+			binary.BigEndian.PutUint64(b, uint64(linkID(p)))
+			hasher.Write(b)
 		}
 	}
 	if layer := p.NetworkLayer(); layer != nil {
-		uuid ^= layer.NetworkFlow().FastHash()
+		Hash(layer.NetworkFlow(), hasher, false)
+		b := make([]byte, 8)
+		binary.BigEndian.PutUint64(b, uint64(networkID(p)))
```
safchain: Why is this needed ?

flow/no_ebpf.go
```diff
@@ -0,0 +1,47 @@
...
+package flow
+
+import (
+	"time"
+	"unsafe"
+)
+
+// EBPFFlow Wrapper type used for passing flows from probe to main agent routine
+type EBPFFlow struct {
+	start        time.Time
```
safchain: All the struct member are useless

flow/probes/ebpf.go
```diff
@@ -80,298 +71,24 @@ type EBPFProbesHandler struct {
...
 	defer p.flowTable.Stop()
 
 	var startKTimeNs int64
 	var start time.Time
 
-	ebpfFlows := make(map[C.__u64]*EBPFFlow)
 	updateTicker := time.NewTicker(ebpfUpdate)
 	defer updateTicker.Stop()
 
+	p.flowPoolSize = 2*cap(flowChanOperation) + 1
```
safchain: p := 2*cap....
not need to have it as struct member

flow/table.go
safchain: I need to check what the cost of this. Yes maybe opening a PR.

flow/table.go
```diff
@@ -445,6 +458,42 @@ func (ft *Table) processFlowOP(op *Operation) {
...
+			ABRstStart: updateTCPFlagTime(fl.TCPMetric.ABRstStart, nfl.TCPMetric.ABRstStart),
+			BARstStart: updateTCPFlagTime(fl.TCPMetric.BARstStart, nfl.TCPMetric.BARstStart),
+		}
+	}
+	// TODO(safchain) remove this should be provided by the sender
+	// with a good time resolution
+	if fl.Metric.RTT == 0 && fl.Metric.ABPackets > 0 && fl.Metric.BAPackets > 0 {
+		fl.Metric.RTT = fl.Last - fl.Start
+	}
+
```
safchain: nit: remove extra line

flow/probes/ebpf.go
```diff
@@ -80,298 +71,24 @@ type EBPFProbesHandler struct {
...
-		Key:  key,
-	}
-}
-
 func (p *EBPFProbe) run() {
 	var info syscall.Sysinfo_t
 	syscall.Sysinfo(&info)
 
-	_, flowChanOperation := p.flowTable.Start()
+	_, flowChanOperation, _ := p.flowTable.Start()
```
safchain: please change flowChanOperation to flowEBPFChan

flow/probes/ebpf.go
eranra: @nplanel @andrekassis is using circular buffer that the size of the buffer is twice the size of the channel he is writing to . You @nplanel are correct that next = 3 (in your example) but this is ok ... 3 is free to use ... and so is 4  and 5 (3*2 size) ... when new flows are written to 3,4,5 the reader will continue to read and this is ok because those are new flows ... when reader ends reading 3,4,5 next goes to 6 ... so circular goes to zero --- i.e. next = 0 and this is again ok, because we know that the reader already read 0,1 and 2 ... I think that 2 * capacity is ok (assuming the implementation of chan in golan has local buffer up to the size of the channel).

What you suggest will work ... I am sitting here with @andrekassis trying to review and I think this is fine. If we are missing something can you help by adding more elaborated example ... sorry for the long remark ... but we are trying to think and understand what is going on here.

flow/flow.go
```diff
@@ -309,28 +312,66 @@ func (p *Packet) TransportFlow() (gopacket.Flow, error) {
 	return layer.TransportFlow(), nil
 }
 
+//Hash flow with custom func, ignore ports in transport level for VXLAN and Geneve
```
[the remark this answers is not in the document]

flow/flow.go
```diff
@@ -309,28 +312,66 @@ func (p *Packet) TransportFlow() (gopacket.Flow, error) {
...
+func Hash(f gopacket.Flow, hasher hash.Hash, isTransportUDP bool) {
```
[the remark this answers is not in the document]

flow/flow.go
andrekassis: maybe, but the whole reason I changed those lines was that the hash function that was being used wasn't working correctly, right?
one more thing though, if you take a look at the code of the hash function for this layer that was in use originally (which is in flows.go in gopacket's code) you won't see any special treatmen for those protocols, and yet the ports are always being hashed, which begs the question, how come this works?

flow/flow.go
andrekassis: if you drop the last line, it would result in mismatching flows (those that have Geneve application layer, because we need the VNI in order to distinguish between them)
** you cant test it yourself -- make test will fail

flow/no_ebpf.go
andrekassis: done

flow/probes/ebpf.go
andrekassis: +1 is redundant so I removed it;)
this is no longer a struct member as well.
As for your comment @nplanel -- I can indeed incorporate this change you are suggesting, but I still don't understand why it's needed. Eran also wrote a comment for u on this. Would be much appreciated if you could take a look, because both of us don't see why the current implementation is wrong.
+ adding a channel in the opposite direction is blocking as well and harms performance so would be smart to avoid this if the code is working as is, right?

flow/table.go
andrekassis: done

flow/probes/ebpf.go
andrekassis: done

flow/flow.go
```diff
@@ -312,25 +312,30 @@ func (p *Packet) TransportFlow() (gopacket.Flow, error) {
...
-		uuid ^= layer.NetworkFlow().FastHash()
+		Hash(layer.NetworkFlow(), hasher, false)
+		b := make([]byte, 8)
+		binary.BigEndian.PutUint64(b, uint64(networkID(p)))
+		hasher.Write(b)
+
 	}
-	if tf, err := p.TransportFlow(); err == nil {
-		uuid ^= tf.FastHash()
+	if layer := p.TransportLayer(); layer != nil {
```
safchain: To follow my comment on the Hash function, previously TransportFlow was used and that were the geneve/vxlan thing was handled. I think you just need to use it instead of layer.TransportFlow so that you should be able to remove the geneve/vxlan thing from your hash function. You will see that the VNI is also used there.

flow/probes/ebpf.go
```diff
@@ -49,17 +46,11 @@ import (
 import "C"
 
 const (
-	BPF_ANY    = 0
-	ebpfUpdate = 2 * time.Second
+	BPF_ANY       = 0
+	FLOW_TABLE_SZ = 500000
+	ebpfUpdate    = 10 * time.Second
```
andrekassis: under large stress, i.e. hundreds of thousands of different flows or more, 2 seconds become just illogical (empirically) since the agent needs much longer than that to go over the entire table.
Changing the number to 10 make much more sense, it's much cheeper in CPU since the agent can rest (unlike 2 where the scans happen all the time and kill the system). 
When the number of connections is low we don't lose anything since we will finally read all the flows.
As for a large number of connections, it seems that the average HTTP server for example handles around 30K connections per second, so since we have a table of 500K that we scan every 10 s, the average number of flows per second we can track is 50K which is bigger.

flow/probes/ebpf.go
```diff
@@ -394,50 +111,31 @@ func (p *EBPFProbe) run() {
...
+				found, err := p.module.LookupNextElement(p.fmap, unsafe.Pointer(&key), unsafe.Pointer(&nextKey), unsafe.Pointer(&kernFlows[nextAvailablePtr]))
 				if !found || err != nil {
+					p.module.DeleteElement(p.fmap, unsafe.Pointer(&key))
+					key = 0
 					break
 				}
-				key = nextKey
 				// delete every entry after we read the entry value
 				p.module.DeleteElement(p.fmap, unsafe.Pointer(&key))
+				key = nextKey
```
andrekassis: we need to delete the previous key and not the new one, since deleting the new one will result in looking for a non existant key next time which will never fail but will always take us back to the beginning of the table (according to the bpf man) and therefore the loop will never end in some scenarios and also we will be starving flows

flow/probes/ebpf.go
```diff
@@ -394,50 +111,31 @@ func (p *EBPFProbe) run() {
...
-			}
-
-			for k, v := range ebpfFlows {
-				if time.Now().Sub(v.last).Seconds() > p.expire.Seconds() {
-					delete(ebpfFlows, k)
+				flow.SetEBPFFlow(&ebpfFlows[nextAvailablePtr], us, last, unsafe.Pointer(&kernFlows[nextAvailablePtr]), startKTimeNs, p.probeNodeTID)
+				flowEBPFChan <- &ebpfFlows[nextAvailablePtr]
+				nextAvailablePtr = (nextAvailablePtr + 1) % flowPoolSize
+				flowsRead++
+				if flowsRead == FLOW_TABLE_SZ {
```
andrekassis: don't scan more than 500K flows everytime (avoid infinite loop -- however, to avoid starvation, we now keep the index of the current flow and start from the same point next time)

andrekassis:

flow/flow.go
andrekassis: you are correct again;)
done

andrekassis:
skydive-functional-tests-backend-orientdb

andrekassis:

andrekassis:
run skydive-k8s-tests

flow/table.go
```diff
@@ -120,11 +122,12 @@ func NewTable(updateHandler *Handler, expireHandler *Handler, nodeTID string, op
 		// convert seconds to milleseconds
 		appTimeout[strings.ToUpper(key)] = int64(1000 * config.GetConfig().GetInt("flow.application_timeout."+key))
 	}
-
+	LRU, _ := simplelru.NewLRU(500000, nil)
```
safchain: This is now a strong limitation, while I'm fine with it I would add something in the config file to set it, default of 500k seems ok:

https://github.com/skydive-project/skydive/blob/master/etc/skydive.yml.default#L455

safchain: I'm not sure about what advantage this implementation about LRU brings as internally it is a map, there is just a check that limit the amount of data and as there is a evictList it should take even more memory space. Probably limiting the amount of flows is better. Moreover there is no event when a flow is evicted so we can't mark it as timeout.

andrekassis: Well, if you remember the previous implementation had no limit on the table's size which was causing it to inflate in size and kill the system. LRU implementation as we discussed enables you to ensure this doesn't happen and alliws you to keep the most recent flows as well. Also we can add a functionality to handle evictions if you'd like...

andrekassis: Well, if you remember the previous implementation had no limit on the table's size which was causing it to inflate in size and kill the system. LRU implementation as we discussed enables you to ensure this doesn't happen and alliws you to keep the most recent flows as well. Also we can add a functionality to handle evictions if you'd like...

safchain:
run skydive-k8s-tests

PR #1836, opened 2019-05-27
topology_manager: do not delete rules in case of an error

lebauce:


masco-enovance:
run skydive-scale-tests

PR #1857, opened 2019-06-10
topology: Added support for VPP running in docker container

fgschwan:
This pull request extends docker topology by information from [VPP](https://wiki.fd.io/view/VPP) running inside docker container:
1. adding/removing VPP node into docker container namespace (if running VPP is detected inside docker container)
2. adding/removing edge between VPP and veth interface in docker container (if it is usable/listed in interface list in VPP)
3. adding/removing [memif](https://docs.fd.io/vpp/19.04/libmemif_doc.html) interface node in docker container (if created by VPP)
4. adding/removing edge between memif intefaces (only one edge is supported by memif socket file) 

Motivation for given topology extensions is this topology case: 
host/node -> veth tunnel -> VPP (first docker container) -> memif tunnel -> VPP (second docker container) -> memif tunnel -> VPP (third docker container) -> veth tunnel -> host/node

Relation to existing VPP probe:
The existing VPP probe is communicating with VPP installed on host/node. My PR is detecting VPP inside docker containers. The existing VPP probe is using govpp, but i use for simplicity VPP CLI due to additional docker client layer.


lebauce:
run tests

lebauce:

fgschwan:
Code compilation is working for me (using "make") , "make fmt" is also success (i'm using VM with ubuntu 18.04, go 1.11) and elastic search failure is some test timeout problem (not sure about the cause).
So what can i do to resolve/help to resolve the failing tests?

lebauce:
@fgschwan First, thanks a lot for this pull request. Our CI is using Go 1.10.3 - which has different `fmt` policies than 1.11 that could explain the issues with `skydive-go-fmt`.

`skydive-compile-tests` fails on Windows compilation because of :
```
agent/probes.go:92:39: too many arguments in call to docker.NewProbe
	have (*netns.Probe, string, string, []subprobes.Subprobe)
	want (*netns.Probe, string, string)
```

lebauce:

fgschwan:
> 
> 
> @fgschwan First, thanks a lot for this pull request. Our CI is using Go 1.10.3 - which has different `fmt` policies than 1.11 that could explain the issues with `skydive-go-fmt`.
> 
I downgraded my go installation to match your CI, reapplied "make fmt" and created fix commit https://github.com/skydive-project/skydive/pull/1857/commits/761ff264864e22836e6afab7414b58a1a27cd670 (basically i amended previous fmt commit).
> `skydive-compile-tests` fails on Windows compilation because of :
> 
> ```
> agent/probes.go:92:39: too many arguments in call to docker.NewProbe
> 	have (*netns.Probe, string, string, []subprobes.Subprobe)
> 	want (*netns.Probe, string, string)
> ```
I didn't properly check all possible architecture/custom build tags combinations. I fixed the code in https://github.com/skydive-project/skydive/pull/1857/commits/8f7829be2ccbfd58cd837c02d194570ed837119b .

matrohon:
great @fgschwan ! this is the code used for this demo? :
https://www.youtube.com/watch?v=wl96kwRmS5I

fgschwan:
> 
> 
> great @fgschwan ! this is the code used for this demo? :
> https://www.youtube.com/watch?v=wl96kwRmS5I

@matrohon, yes, but code used in that demo was before heavy refactoring that was needed to do this PR. So do code is different, but the functionality is exactly the same.

matrohon:
I wonder how we could make this probe more generic:
- what if another runtine than docker is used (runc?...)
- what if the container are running in a kubernetes context?
- what if we want to see other vpp interfaces than container/memif , like container/vpp-vxlan or container/vpp-physical-nics(veth)?


I'm going to play with your probe in the context of the [NSM](https://github.com/networkservicemesh/networkservicemesh) project.

fgschwan:
> 
> I wonder how we could make this probe more generic:
> 
>     * what if another runtine than docker is used (runc?...)
VPP subprobe must be decoupled from docker probe. That means that the API between docker probe and VPP subprobe have to be more generic and the VPP data retrieving also have to be changed to support other containerization technologies.
> 
>     * what if the container are running in a kubernetes context?
I didn't check the kubernetes part of Skydive, so no idea what additional kubernetes-related functionality must be supported.
> 
>     * what if we want to see other vpp interfaces than container/memif , like container/vpp-vxlan or container/vpp-physical-nics(veth)?
I think that we are hitting some complexity threshold here. The code in this PR is using the VPP CLI. It works. It can be extended also to get information about vxlan or veth from VPP just by extending info retrieving capabilities. The problem is that VPP CLI was never meant for heavier data grabbing. So the proper solution in compare to using more and more VPP CLI would be the binary API in VPP. This is actually used in VPP probe (the non-container VPP topology) by usage of [govpp](https://github.com/FDio/govpp). The binary API gives us structured data without the need to parse text from CLI. There is also possibility to make some common code with VPP probe(non-container VPP), but there containers can have different VPP versions and that is problem. The binary API probably can also solve some performance problems (i never checked that with govpp). I tried to make my code more efficient in case of many docker containers and in the end the docker client calls to container were the problem. I minimized their usage by getting more information in on docker client call, but that is all i can do. If i used govpp, i could communicate to VPP directly using socket file and bypass docker. The only problem is to get VPP socket file, but that can be done by using docker inspect and grab "MergedDir" to get path to docker filesystem and get socket file there. Don't know if something similar can be done for [OCI](https://github.com/opencontainers).

matrohon:
@fgschwan thanks for your reply.

An interesting alternative would be to receive notifications from vpp, by using vpp-agent for instance:
https://github.com/ligato/vpp-agent/tree/master/examples/grpc_vpp/notifications

It would avoid polling. But it also implies that the container running vpp is also running vppagent.

It seems pantheon is working on those technologies too: https://pantheon.tech/vpp-grpc-rest/

fgschwan:
> It would avoid polling. But it also implies that the container running vpp is also running vppagent.
> 
> It seems pantheon is working on those technologies too: https://pantheon.tech/vpp-grpc-rest/
I'm familiar with vpp-agent, but the need to have it running inside container breaks skydive's ability to detect VPP in ANY docker container. Of course if you are some kind of owner of docker containers and can enforce running vpp-agent inside docker containers then you can use vpp-agent. IMHO this could be used only when also running vpp-agent is detected in container (together with running VPP) and in cases without vpp-agent fallback to non-vpp-agent solution.

fgschwan:

fgschwan:
@lebauce
I did some functional tests for this PR (https://github.com/skydive-project/skydive/pull/1857/commits/718cf0a7d6a6eca569d8e7657dee058822ca291d), but to properly run my tests with your CI i needed to set env var WITH_DOCKER_VPP to true. I checked how it is done in other PRs or other tests and it seems to be either in scripts/ci/jobs/jobs.yml or in CI scripts (package scripts/ci). I was able to run my tests using the scripts, but they are running 3 times (skydive-functional-tests-backend-orientdb([log](http://ci-logs.skydive.network/skydive-functional-tests-backend-orientdb/builds/3819/log.html)), skydive-functional-tests-backend-elasticsearch([log](http://ci-logs.skydive.network/skydive-functional-tests-backend-elasticsearch/builds/3772/log.html)), skydive-functional-hw-tests) and one test is always failing in skydive-functional-hw-tests (i guess that something is colliding in setup with docker container with VPP because in that test the container must be in privileged mode due to veth interfaces setup). So the solution is to reduce my tests (TestRunningVPPInDocker, TestDockerVPPConnectingToVeth, TestTwoVPPsConnectedUsingMemifTunnel) to only one test run, i.e skydive-functional-tests-backend-orientdb. To do that i had to move setting of env var WITH_DOCKER_VPP to jobs.yml. The problem is that i can't really test such launch configuration (i have it now commited in such state, see currently last commit https://github.com/skydive-project/skydive/pull/1857/commits/d9a03ea3e457ebe83b90be18d3ab5079da45045a), because my changed jobs.yml seems to not be used by your CI to run the tests. 

I guess that someone from your side needs to update the (jenkins?) jobs (using scripts/ci/jobs/update-jobs.py ?). Or do you prefer other way how to do it?

fgschwan:

lebauce:
@fgschwan Indeed. I just updated our Jenkins with your change.

Sorry for the delay, we were a bit busy those 2 last weeks. We will give a proper review of your PR this week for sure.

lebauce:

fgschwan:

fgschwan:
@lebauce thank you, tests are passing in skydive-functional-tests-backend-orientdb test suite (see [log](http://ci-logs.skydive.network/skydive-functional-tests-backend-orientdb/builds/3897/log.html) for tests TestRunningVPPInDocker, TestDockerVPPConnectingToVeth, TestTwoVPPsConnectedUsingMemifTunnel)

lebauce:
@fgschwan I took a pretty long to review your PR, sorry about that.

I actually spend quite a long time trying things.

My concern was that we would kinda have 2 different VPP probes : one that uses govpp, the other one using the CLI.
I therefore tried to run the existing VPP probe - that uses govpp - inside the container. The problem is that govpp uses /dev/shm to talk with VPP so we have to enter the PID and most importantly the mount namespace. I created a simple mode for Skydive (in addition to `agent` and `analyzer`) that is basically a very stripped down version of the agent but that runs only a single probe. Before starting the probe, it simply jumps to the right namespaces (so that we can use /dev/shm). I faced many issues (regarding the fact that the VPP libs don't support changing namespaces). Then I found that new versions of `govpp` allow talking to VPP using dead simple UNIX sockets. With this method, my trick worked nicely and it brings also a nice thing : the Skydive executable do not rely on VPP libs (libvppinfra.so, libvppapiclient.so, .....), so we don't care about the version of VPP that is installed on the host or inside the container.

I also tried to simplify the code that handles the links between the VPP node and the interfaces, and the links between the memifs (using Linker and MetadataIndexerLinker https://github.com/skydive-project/skydive/blob/master/graffiti/graph/linker.go and https://github.com/skydive-project/skydive/blob/master/graffiti/graph/indexer.go).

But I don't see the point in delaying the merging of your PR, I'll push a PR with the refactoring later and add you as a reviewer, if it's ok for you.

That being said, I have one remark with the PR. It's a bit odd to have a `VPP` folder in the `docker` probe, with a `docker.go` inside this `VPP` folder. So I suggest something :
In the `NewProbe` of the existing `vpp` probe, we can register a graph listener using `g.AddEventListener(listener)` so that we are notified when a container has been added. To do so, we can react on new edge in `OnEdgeAdded`. If the parent is a network namespace (Type: "netns") and the children a container ("Type": "container"), then we start the subprobe for this particular namespace.

With this change, all the related VPP code would be in `topology/probes/vpp` and not in `topology/probes/docker`. For the VPP probe to be able to use the Docker probe (for the `docker exec` thing), the docker probe could be passed to `vpp.NewProbe` (like how the Docker probe is passed the Netlink probe in https://github.com/skydive-project/skydive/blob/master/topology/probes/docker/docker.go#L289) and a public `Exec` function could be added to the Docker probe. This way, we don't have a instantiate a new Docker client and could use the existing one.

fgschwan:
@lebauce Thank you for your review.

> the Skydive executable do not rely on VPP libs (libvppinfra.so, libvppapiclient.so, .....), so we don't care about the version of VPP that is installed on the host or inside the container.

I don't think that we can ignore VPP version, because govpp is using binary API of VPP and that can change from version to version. You probably can compile it, but actual communication with VPP for those binary API that changes may fail.

>But I don't see the point in delaying the merging of your PR, I'll push a PR with the refactoring later and add you as a reviewer, if it's ok for you.

Yes, please add me as reviewer to your refactoring.

>It's a bit odd to have a VPP folder in the docker probe, with a docker.go inside this VPP folder

The docker.go was meant as library for all subprobes, but there was only one subprobe (the VPP subprobe). So i left it inside the VPP subprobe. The only other place that i can put this file without getting cyclic dependency is topology/probes/docker and that seems odd to because it is for subprobes only.

> So I suggest something :
In the NewProbe of the existing vpp probe, we can register a graph listener using g.AddEventListener(listener) so that we are notified when a container has been added. To do so, we can react on new edge in OnEdgeAdded. If the parent is a network namespace (Type: "netns") and the children a container ("Type": "container"), then we start the subprobe for this particular namespace.

This seems like a nice idea. Maybe easily breakable if someone changes docker probe code in some way and not knowing dependency in vpp probe, but that could be manageable by using tests. Also i checked the subprobe API if this is possible. All thing except https://github.com/skydive-project/skydive/pull/1857/files#diff-01d2cbebb5fe6a69afd9cd490e4791d6R39 are graph related and this one thing that is not is used to grab mounted volumes from docker to see whether memif interfaces are referring to the same physical socket file. This can be managed i.e. by docker probe exposing volume mount information in metadata (don't know if it does right know or not).

> This way, we don't have a instantiate a new Docker client and could use the existing one.

I thought about using docker client provided from docker probe (to vpp subprobe), but there was no goroutine-safety (thread-safety) guarantee about the docker client and i was running docker "exec" concurrently in multiple goroutines due to performance. Ideally, the move to use govpp would probably solve performance issues that i tracked down to docker client (too many docker client calls for many docker containers). I would also allow us to get rid of docker client. I already discussed govpp usage in the end of https://github.com/skydive-project/skydive/pull/1857#issuecomment-501721194 . Using govpp could also unify vpp probe and docker vpp-subprobe code to use the same code for grabbing data from VPP. The only problem is that different version of VPP have different binary API (most thing doesn't change between version but some do)

fgschwan:
> > the Skydive executable do not rely on VPP libs (libvppinfra.so, libvppapiclient.so, .....), so we don't care about the version of VPP that is installed on the host or inside the container.
> 
> I don't think that we can ignore VPP version, because govpp is using binary API of VPP and that can change from version to version. You probably can compile it, but actual communication with VPP for those binary API that changes may fail.

I checked with govpp people what is the current state and future plans for handling differences in VPP versions. There are actually 2 binary API, one for handling configuration and on for statistics (2 different unix socket files). The important one (configuration binary api = get/set config values) currently needs to have generated structs from binary API json files from VPP. The statistic API needed in past generated structs as the configuration API, but now it works differently. It is more generic and without generated structs. To make binary api call you fill string name of method you want to call and values as parameters. Then govpp finds it in binary API definition files of VPP (json files) and tries to make the proper format for call to VPP. Of course if you try to call unsupported/dropped/modified API part of VPP, it will fail. But it will fail dynamically in runtime and not in compilation time. This gives more power to user of govpp, because with generation of struct any change in API invalidated usage of govpp for running VPP (version mismatch). With this approach, if API mismatch happens in part of API that is not used at all, everything works. As i mentioned earlier, this is approach is currently only in statistic binary API and some day it will be available for configuration binary API. So in the end, different VPP versions will still need to be handled, but it won't be that difficult as to have generated structs for all supported VPP versions and properly switch between them.

lebauce:
@fgschwan We recently merged the support for `seeds` in Skydive : it's a single Skydive probe executed as a standalone process in the Docker namespace.
This PR modifies the VPP existing probe to be able to run inside the namespace, using the socket interface. I integrated the memifs retrieval that you implemented and the links between them. I also cherry-picked the tests from your PR.
It would be very nice if you could give it a try.

PR #1885, opened 2019-06-26
flow: stop capturing if afpacket poll return error

safchain:


safchain:
run skydive-k8s-tests

PR #1891, opened 2019-06-30
dns for security advisor

andrekassis:
Hi guys,
following the recent PR from @hunchback , in this commit we extend the provided pipeline with custom stages to allow writing DNS packets to the object store. Since Skydive is a protocol analyzer, we utilize this powerful tool to provide the admin with the ability to monitor all the DNS traffic and log it for further use and analysis.
+ we enable logging all the DNS data via an array and not a single packet in each flow, since using a single packet results in overwriting data. e.g. for a typical DNS question at least two packets are generated, one with the AAAA code (for ipv6) and another with the A code (ipv4) and therefore keeping one packet will discard the others. To not overwhelm the system we keep this feature optional which can now be activated using a new extra layer flag.
also there was a minor bug in the DNS code that is now fixed in this commit.
Cheers


andrekassis:
run  skydive-cdd-overview-tests

andrekassis:
run  skydive-cdd-overview-tests

andrekassis:
run  skydive-cdd-overview-tests

andrekassis:

flow/flow.go
```diff
@@ -108,14 +108,17 @@ const (
...
 )
 
 var extraLayersMap = map[string]ExtraLayers{
-	"VRRP":   VRRPLayer,
-	"DNS":    DNSLayer,
-	"DHCPv4": DHCPv4Layer,
+	"VRRP":            VRRPLayer,
+	"DNS":             DNSLayer,
+	"DHCPv4":          DHCPv4Layer,
+	"DNS_ALL_PACKETS": DNSLayerAll,
```
safchain: I would keep only one option DNS, getting all the packets. The only thing we need to check that we can request using gremlin, we should add functional test for that

flow/storage/elasticsearch/elasticsearch.go
```diff
@@ -135,7 +135,7 @@ type embeddedFlow struct {
 	Transport    *flow.TransportLayer `json:"Transport,omitempty"`
 	ICMP         *flow.ICMPLayer      `json:"ICMP,omitempty"`
 	DHCPv4       *fl.DHCPv4           `json:"DHCPv4,omitempty"`
-	DNS          *fl.DNS              `json:"DNS,omitempty"`
+	DNS          []*fl.DNS            `json:"DNS,omitempty"`
```
safchain: as here you are changing the schema, you need to bump the version
https://github.com/skydive-project/skydive/blob/master/storage/elasticsearch/client.go#L42

flow/flow.go
```diff
@@ -108,14 +108,17 @@ const (
 	DNSLayer ExtraLayers = 2
 	// DHCPv4Layer extra layer
 	DHCPv4Layer ExtraLayers = 4
+	// DNSLayerAll extra layer, intercept every packet
+	DNSLayerAll ExtraLayers = 8
 	// ALLLayer all extra layers
-	ALLLayer ExtraLayers = 255
+	ALLLayer ExtraLayers = 247
```
safchain: any reason to change this ?

flow/storage/elasticsearch/elasticsearch.go
andrekassis: So all that needs to be done is just to change the number?

flow/flow.go
andrekassis: Just did this so that if one requests all layers they won't have to, by default, capture all the packets. But since we are only keeping one option I will modify this again

flow/flow.go
andrekassis: Right now you can request this via the --extra-layer flag. Isn't this enough?

safchain: I would just keep DNS, I do not see the point to get multiple level of analysis for DNS. My point was since we change the way the DNS fields are reported in flow we need to ensure we can still filter flow based on DNS fields. Adding functionals test will help to ensure it is working.

flow/storage/elasticsearch/elasticsearch.go
safchain: yes just increasing the version is enough

andrekassis: done

flow/flow.go

flow/flow.go
andrekassis: Added one simple test which ensures we can filter flows by certain DNS fields

tests/flow_test.go
```diff
@@ -1402,6 +1402,57 @@ func TestPcapInject(t *testing.T) {
 	RunTest(t, test)
 }
 
+func TestDNSRetrieval(t *testing.T) {
+	test := &Test{
+		setupFunction: func(c *TestContext) error {
+			file, err := os.Open("../flow/pcaptraces/dns.pcap")
```
safchain: There is already a test that use a pcap file with DNS packet. I would suggest to simply enhance it
https://github.com/skydive-project/skydive/blob/master/tests/flow_test.go#L1354

andrekassis:
run  skydive-functional-tests-backend-orientdb

andrekassis:
run  skydive-functional-tests-backend-orientdb

andrekassis:
run  skydive-functional-tests-backend-orientdb

andrekassis:
run  skydive-functional-tests-backend-elasticsearch

andrekassis:
run  skydive-functional-tests-backend-elasticsearch

andrekassis:
run tests

safchain:
In order to be able to filter DNS related field you need to add the corresponding getters
https://github.com/skydive-project/skydive/blob/master/flow/flow.go#L1121

contrib/pipelines/secadvisor/sadns/classify.go
```diff
@@ -0,0 +1,49 @@
...
+// classifyDNS classifies DNS flows
+type classifyDNS struct {
+}
+
+// GetFlowTag determines whether a flow holds DNS data
+func (fc *classifyDNS) GetFlowTag(fl *flow.Flow) core.Tag {
+	if fl == nil || fl.Transport == nil {
+		return tagNonDNS
+	}
+	if fl.Transport.A != 53 && fl.Transport.B != 53 {
```
hunchback: s/53/DNSPort/

PR #1893, opened 2019-07-02
flow: support partial flow update

masco:


masco-enovance:
run skydive-compile-tests
run skydive-unit-tests

flow/client/flow_client.go
```diff
@@ -160,29 +160,49 @@ retry:
...
 		bulkSize = 10
 	}
-
 	var msg flow.Message
-	for i := 0; i < len(flows); i += bulkSize {
-		e := i + bulkSize
-
-		if e > len(flows) {
-			e = len(flows)
+	switch items.(type) {
```
hunchback: rather use this construct: https://tour.golang.org/methods/16

api/server/pcap.go
```diff
@@ -36,8 +38,36 @@ type PcapAPI struct {
 }
 
 // SendFlows implements the flow Sender interface
-func (p *PcapAPI) SendFlows(flows []*flow.Flow) {
-	if p.Storage != nil && len(flows) > 0 {
+func (p *PcapAPI) SendFlows(item interface{}) {
```
hunchback: an alternative option is to add along side `SendFlows(flow []*flow.flow)` also `SendFlowUpdates(update []*flow.FlowUpdate)`

safchain: +1, we can use interfaces this is too costly. Please add a function to the Flow Sender interface. I would suggest to add (if possible) an option the the flow table in order to specify if we want to get "Full Flow update" or "Partial Flow update", so that NetFlow target can remain as it is and this one too.

flow/server/flow_server.go
```diff
@@ -208,15 +203,65 @@ func NewFlowServerUDPConn(addr string, port int) (*FlowServerUDPConn, error) {
...
+
+	var flows []*flow.Flow
+	for _, msg := range msgs {
+		if len(msg.Flows) > 0 {
+			flows = append(flows, msg.Flows...)
+		}
+		if len(msg.Updates) > 0 {
+			for _, update := range msg.Updates {
+				fsq := filters.SearchQuery{}
+				tsf := filters.NewTermStringFilter("UUID", update.FlowUUID)
```
safchain: We can request the datastore for each update, this will decrease the performance. I would add a UpdateFlows function to the storage interface.

flow/table.go
```diff
@@ -158,7 +158,8 @@ func NewTable(updateEvery, expireAfter time.Duration, sender Sender, uuids UUIDs
 	return t
 }
 
-func (ft *Table) getFlows(query *filters.SearchQuery) *FlowSet {
+//GetFlows returns flowset
+func (ft *Table) GetFlows(query *filters.SearchQuery) *FlowSet {
```
safchain: This has to remain private has we can call it from outside because of mutex/lock. This has to call be the flow table through a chan. If we had the option in order to specify whether we want partial updates it should remove the need of thi s modification.

flow/table.go
```diff
@@ -270,19 +271,48 @@ func (ft *Table) updateMetric(f *Flow, start, last int64) {
...
-	var updatedFlows []*Flow
+	var newFlows []*Flow
+	var updatedFlows []*FlowUpdate
 	for _, k := range ft.table.Keys() {
 		fl, _ := ft.table.Peek(k)
 		f := fl.(*Flow)
 		if f.XXX_state.updateVersion > ft.updateVersion {
 			ft.updateMetric(f, updateFrom, updateTime)
-			updatedFlows = append(updatedFlows, f)
+			if f.XXX_state.newFlow {
```
safchain: check here the "partial" option

flow/table.go
```diff
@@ -302,18 +332,23 @@ func (ft *Table) update(updateFrom, updateTime int64) {
...
 		if ft.Opts.RawPacketLimit > 0 {
-			for _, f := range updatedFlows {
+			for _, f := range newFlows {
 				f.LastRawPackets = f.LastRawPackets[:0]
 			}
 		}
 	}
+
+	if len(updatedFlows) != 0 {
+		ft.sender.SendFlows(updatedFlows)
```
safchain: and then call here either SendFlows or SendFlowUpdates

flow/server/flow_server.go
```diff
@@ -208,15 +203,65 @@ func NewFlowServerUDPConn(addr string, port int) (*FlowServerUDPConn, error) {
...
+	}
+
+	if len(flows) > 0 {
+		if err := s.storage.StoreFlows(flows); err != nil {
+			logging.GetLogger().Error(err)
+		} else {
+			logging.GetLogger().Debugf("%d flows stored", len(flows))
+		}
 
 		s.subscriberEndpoint.SendFlows(flows)
```
masco-enovance: @safchain if we use updateFlows in storage, how to send the full flows to the subscriber without querying?

masco:

PR #1923, opened 2019-07-24
Pipeline refactor

hunchback:
The main goal of this change is to enable seamless extention of pipelines such that a specific pipeline can add additional options (such as a custom transformer) while still having access to all the optional implementations that exist in the core. For example is secadvisor has a specific transform function I want to add that but still be able to set transform.type=none and disable that step.

The secondary goal was to reduce the amount of code for anyone implementing a pipeline to the bare minimum ... and after the change all you need is:

```
func main() {
        core.Main("/etc/skydive/xxx.yml")
}

func init() {
        // register any unique pipeline phases here
}
```

As the current test code broke many aspects of the code structure (for example imposing additional code to read object from store and decompress objects) I have removed the test code; at a later stage I will add an alternative set of tests (testing individual capabilities).


hunchback:

PR #1928, opened 2019-07-26
ebpf : use github.com/newtools/ebpf package

nplanel:
o need an upstream patch to avoid memcpy
o need for fastest functional test

### DIFF
--- a/flow/probes/ebpf.go
+++ b/flow/probes/ebpf.go
@@ -23,13 +23,14 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"math"
 	"runtime"
 	"sync"
 	"syscall"
 	"time"
 	"unsafe"
 
-	"github.com/iovisor/gobpf/elf"
+	"github.com/newtools/ebpf"
 
 	"github.com/skydive-project/skydive/api/types"
 	"github.com/skydive-project/skydive/common"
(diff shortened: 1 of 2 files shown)


flow/probes/ebpf.go
```diff
@@ -261,45 +275,59 @@ func (p *EBPFProbesHandler) Stop() {
...
 
 	if err == nil {
 		p.Ctx.Logger.Infof("Loaded eBPF module ", path)
 	}
 	return module, err
 }
 
-func (p *EBPFProbesHandler) loadModule() (*elf.Module, error) {
+func (p *EBPFProbesHandler) loadModule() (*ebpf.Collection, error) {
+	err := syscall.Setrlimit(8, &syscall.Rlimit{
```
dubek: I assume the `8` here is for `RLIMIT_MEMLOCK`; I suggest adding a comment or using a C constant from `#include <sys/resource.h>`.

nplanel: Yes I forgot to add it, thanks

probe/ebpf/bpf.h
```diff
@@ -28,6 +28,28 @@
...
+# define memcmp(src1, src2, n)   __builtin_memcmp((src1), (src2), (n))
+#endif
+#else
+#define memcmp discret_memcmp
+static __inline int discret_memcmp(char *s1, char *s2, int n) {
+	int i;
+#pragma unroll
+	for(i=0;i<n;s1++,s2++,i++) {
+		if (*s1 != *s2) {
+			return 1;
```
dubek: memcmp(2) return value is:

> RETURN VALUE
> The  memcmp() function returns an integer less than, equal to, or greater than zero if the first n bytes of s1 is found, respectively, to be less than, to match, or be greater than the first n bytes of s2.
>
> For a nonzero return value, the sign is determined by the sign of the difference between the first pair of bytes (interpreted as unsigned char) that differ in s1 and s2.
>
> If n is zero, the return value is zero.

If we ever rely on the negative/zero/positive return value (for example for using in qsort()), then this implementation of memcmp will break (because it just returns 0 or 1).

nplanel:

nplanel:

nplanel:

PR #1937, opened 2019-07-30
Fix flow SetUUIDs no link

safchain:


flow/ebpf.go
```diff
@@ -163,14 +164,18 @@ func (ft *Table) newFlowFromEBPF(ebpfFlow *EBPFFlow, key uint64) ([]uint64, []*F
 			}
 		}
 
+		if parent.Link == nil && parent.Network == nil {
+			return nil, nil, errors.New("packet unknown, no link, no network layer")
```
nplanel: the only case where this should happen is when we receive packet with less than 14 bytes, but how ?

PR #1944, opened 2019-08-01
Makefile : build easyjson targets in parallel

nplanel:


nplanel:
run skydive-go-fmt

lebauce:
I'm not sure we can do this now when using go generate.
Besides, as the generated files are part from the repository, speed is not a big issue now so I suggest we close this

nplanel:
Yes we can close this PR as you say it's convered by go mod / go generate #1984

PR #1950, opened 2019-08-07
gremlin: Properly escape query arguments

dubek:
When constructing the Gremlin query string, double quotes were not
escaped, and therefore using a double quote char inside a query argument
would break the query.  This fix uses json.Marshal to quote a string, so
special chars like quote, newline, and tab will be backslash-quoted
inside the resulting Gremlin query string.

Moreover, quoting was done by passing the input as the format string to
fmt.Sprintf, which causes "format string injection" issues (these cannot
cause segfaults like in C, but still not very nice). It seems that this
functionality is not used at all, but also broke query arguments that
have the `%` char in them.

Note that this changes the public function signatures of `gremlin.Quote` and `gremlin.Regex` - it seems they are not used outside this package but please take extra look.

This fix now allows the following queries:

    G.V().Has("foo", "%p") // --> `G.V().Has("foo", "%p")`
    G.V().Has("foo", `"`) // --> `G.V().Has("foo", "\\"")`

The commit adds unit tests for these cases:

    $ go test -v
    === RUN   TestQueryBuilder
    --- PASS: TestQueryBuilder (0.00s)
    === RUN   TestQueryWithPercentSign
    --- PASS: TestQueryWithPercentSign (0.00s)
    === RUN   TestQueryWithDoubleQuotes
    --- PASS: TestQueryWithDoubleQuotes (0.00s)
    === RUN   TestRegexWithPercentSignAndDoubleQuotes
    --- PASS: TestRegexWithPercentSignAndDoubleQuotes (0.00s)
    PASS
    ok      github.com/skydive-project/skydive/gremlin      0.007s

### DIFF
--- a/gremlin/value.go
+++ b/gremlin/value.go
@@ -18,6 +18,7 @@
 package gremlin
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 )
@@ -52,14 +53,18 @@ func (v ValueString) String() string {
 const DESC = ValueString("DESC")
 
 // Quote used to quote string values as needed by query
-func Quote(format string, a ...interface{}) ValueString {
-	s := fmt.Sprintf(format, a...)
-	return ValueString(fmt.Sprintf(`"%s"`, s))
+func Quote(s string) ValueString {
+	escaped, err := json.Marshal(s)
+	if err != nil {
+		// According to the encoding/json source code, json.Marshal will never return
+		// error if the input type is a string; so this case is currently unlikely.
+		return ValueString(fmt.Sprintf("Unexpected error when quoting string: %v", err))
+	}
+	return ValueString(string(escaped))
 }
 
 // Regex used for constructing a regexp expression string
-func Regex(format string, a ...interface{}) ValueString {
-	s := fmt.Sprintf(format, a...)
+func Regex(s string) ValueString {
 	return ValueString(fmt.Sprintf("Regex(%s)", Quote(s)))
 }
 
--- a/gremlin/query_test.go
+++ b/gremlin/query_test.go
@@ -26,7 +26,31 @@ func TestQueryBuilder(t *testing.T) {
 	const key, val = "Type", "host"
 	expected := fmt.Sprintf(`G.V().Has("%s", Regex("%s"), "%s", Regex("%s.*"), "%s", Regex(".*%s")).Flows().Sort()`,
 		key, val, key, val, key, val)
-	actual := G.V().Has(Quote(key), Regex(val), Quote(key), Regex("%s.*", val), Quote(key), Regex(".*%s", val)).Flows().Sort()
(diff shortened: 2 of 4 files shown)


gremlin/value.go
```diff
@@ -52,14 +53,16 @@ func (v ValueString) String() string {
 const DESC = ValueString("DESC")
 
 // Quote used to quote string values as needed by query
-func Quote(format string, a ...interface{}) ValueString {
-	s := fmt.Sprintf(format, a...)
-	return ValueString(fmt.Sprintf(`"%s"`, s))
+func Quote(s string) ValueString {
+	escaped, err := json.Marshal(s)
+	if err != nil {
+		panic(fmt.Sprintf("Quote failed escaping: %v", err))
```
safchain: I wouldn't panic here. Even if it is used only for test purpose currently later we could have this used by a user input. I would rather return a string/token that will generate a wrong gremlin expression so that the parser will catch the error. What do you think ?

dubek: Can `json.Marshal` panic if given a `string` argument? From examining the Go source code I couldn't find such a case, even if the the input string contains illegal UTF8 byte sequences. Since I can't think of such a case, I thought I could completely ignore the `err`, but maybe just a bit clearer with `panic` here. Not sure...

nplanel: @dubek I agree as json.Marshal(<string>) will not generate error ... just ignore the error and add a comment that we ignoring error here because the type is string  and/or return a static ValueString with the error. But don't panic here this could kill other (3rd party) clients

dubek: Good point, thanks for explaining. I changed this in d3a558d3ff3358f83e2efeaf752fb509b26df874.

What are those 3rd party clients? Might they break with the function signature change I introduced?

```diff
-func Quote(format string, a ...interface{}) ValueString {
+func Quote(s string) ValueString {
```

safchain:
run skydive-compile-tests
run skydive-k8s-tests

gremlin/value.go
```diff
@@ -52,14 +53,18 @@ func (v ValueString) String() string {
...
+		// error if the input type is a string; so this case is currently unlikely.
+		return ValueString(fmt.Sprintf("Unexpected error when quoting string: %v", err))
+	}
+	return ValueString(string(escaped))
 }
 
 // Regex used for constructing a regexp expression string
-func Regex(format string, a ...interface{}) ValueString {
-	s := fmt.Sprintf(format, a...)
+func Regex(s string) ValueString {
```
safchain: As you changed the signature of this function you need to callers, please check the CI test results. I have seen that (at least) istio test are failing because of this.

dubek: Thanks, fixed in 32a4ce91d9aeb8810b1a561074b2178561cc67b3 .

safchain:
run skydive-scale-tests

safchain:
run skydive-scale-tests

PR #1955, opened 2019-08-14
flowlogs: adding misc fields via configuration

KalmanMeth:
Given the Flow Exporter pipeline developed by @hunchback, we add to the secadvisor pipeline the capability to specify gremlin expressions with substitution in the yml file to generate additional fields in the output flow information. The basic syntax in the yml file looks like:
  transform:
    type: sa
    extend:
      - AA_Name=G.V().Has('RoutingTables.Src','{{.Network.A}}').Values('Host')
      - BB_Name=G.V().Has('RoutingTables.Src','{{.Network.B}}').Values('Host')
This uses the template golang feature, where template expressions are enclosed in {{ }}.
The fields surrounded by `{{ }}` are taken as the names of fields in the flow information and are replaced with their actual values, before evaluating the gremlin expression and placing the result in a new field which is added to the flow information.
@hunchback @eranra Your initial feedback is most valued.
There is also an extension to perform lookup on VM names instead of Container names in file resolve_vm.go.

### DIFF
--- a/contrib/exporters/secadvisor/mod/extend_gremlin.go
+++ b/contrib/exporters/secadvisor/mod/extend_gremlin.go
@@ -1,0 +1,132 @@
+/*
+ * Copyright (C) 2019 IBM, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy ofthe License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specificlanguage governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package mod
+
+import (
+	"bytes"
+	"encoding/json"
+	"strings"
+	"text/template"
+	"time"
+
+	cache "github.com/pmylund/go-cache"
+	"github.com/skydive-project/skydive/api/client"
+	"github.com/skydive-project/skydive/contrib/exporters/core"
+	"github.com/skydive-project/skydive/logging"
+	"github.com/spf13/viper"
+)
+
+// extendGremlin is used for transform extensions using gremlin expressions
+// Syntax supported is:
+//	VAR_NAME=<gremlin expression with a substitution strings>
+// Substitution strings are defined according to golang template usage using {{ and }}
+// For example:
+//	AA_Name=G.V().Has('RoutingTables.Src','{{.Network.A}}').Values('Host')
+// The substitution string refers to fields that exist in the data provided by the particular transform
+// If needed, be sure to put quotes around the substitution results
+// Use only single quotes in the gremlin expression
+// Need to define `Extend` field as a map[string]interface{} subfield of `transform`
+// Structure in yml file is as follows:
+//   transform:
+//     type: sa
+//     sa:
+//       extend:
+//         - VAR_NAME1=<gremlin expression with substitution strings>
+//         - VAR_NAME2=<gremlin expression with substitution strings>
+type extendGremlin struct {
+	gremlinClient    *client.GremlinQueryHelper
+	gremlinExprCache *cache.Cache
+	gremlinTemplates []*template.Template
+	newVars          []string
+	nTemplates       int
+}
+
+func NewExtendGremlin(cfg *viper.Viper) *extendGremlin {
+	gremlinClient := client.NewGremlinQueryHelper(core.CfgAuthOpts(cfg))
+	extendStrings := cfg.GetStringSlice(core.CfgRoot + "transform.extend")
+	var gremlinTemplates []*template.Template
+	var newVars []string
+	var nTemplates int
+	nTemplates = 0
+	gremlinTemplates = make([]*template.Template, len(extendStrings))
+	newVars = make([]string, len(extendStrings))
+	// parse the extension expressions and save them for future use
+	for _, ee := range extendStrings {
+		substrings := strings.SplitN(ee, "=", 2)
+		vv, err := template.New("extend_template").Parse(substrings[1])
+		if err != nil {
+			logging.GetLogger().Errorf("NewExtendGremlin: template error: %s", err)
+			continue
+		}
+		gremlinTemplates[nTemplates] = vv
+		newVars[nTemplates] = substrings[0]
+		nTemplates++
+	}
+	p := &extendGremlin{
+		gremlinClient:    gremlinClient,
+		gremlinExprCache: cache.New(10*time.Minute, 10*time.Minute),
+		gremlinTemplates: gremlinTemplates,
+		newVars:          newVars,
+		nTemplates:       nTemplates,
+	}
+	return p
+}
+
+// Extend: main function called to perform the extension substitutions
+// we expect the data of a single flow per call
+func (e *extendGremlin) Extend(in *SecurityAdvisorFlow) {
+	for i := 0; i < e.nTemplates; i++ {
+		var tpl bytes.Buffer
+		var err error
+		// apply the gremlin expression
+		err = e.gremlinTemplates[i].Execute(&tpl, in)
+		if err != nil {
+			logging.GetLogger().Errorf("Extend: gremlin template = %s", e.gremlinTemplates[i])
+			logging.GetLogger().Errorf("Extend: template parse error; err = %s", err)
+			continue
+		}
+		result := tpl.String()
+		// check if result is already in cache; if not, perform gremlin request
+		result2, ok := e.gremlinExprCache.Get(result)
+		if !ok {
+			// need to perform the gremlin query
+			result3, err := e.gremlinClient.Query(result)
+			if err != nil {
+				logging.GetLogger().Errorf("Extend: gremlin query = %s", result)
+				logging.GetLogger().Errorf("Extend: gremlin query error; err = %s", err)
+				continue
+			}
+			if len(result3) == 0 {
+				logging.GetLogger().Errorf("Extend: gremlin query error; result = %s", result)
+				continue
+			}
+			var result4 []string
+			if err = json.Unmarshal(result3, &result4); err != nil {
+				logging.GetLogger().Errorf("Extend: unmarshal error; err = %s", err)
+				continue
+			}
+			// add new field in map with the desired key name
+			if 0 == len(result4) {
+				continue
+			}
+			e.gremlinExprCache.Set(result, result4[0], cache.DefaultExpiration)
+			result2 = result4[0]
+		}
+		in.Extend[e.newVars[i]] = result2
+	}
+}
--- a/contrib/exporters/secadvisor/mod/transform.go
+++ b/contrib/exporters/secadvisor/mod/transform.go
@@ -39,10 +39,13 @@ func NewTransform(cfg *viper.Viper) (interface{}, error) {
 	resolver = NewResolveFallback(resolver)
 	resolver = NewResolveCache(resolver)
 
+	newExtend := NewExtendGremlin(cfg)
+
 	return &securityAdvisorFlowTransformer{
 		resolver:             resolver,
 		flowUpdateCountCache: cache.New(10*time.Minute, 10*time.Minute),
 		excludeStartedFlows:  excludeStartedFlows,
+		extendGremlin:        newExtend,
 	}, nil
 }
 
@@ -75,6 +78,7 @@ type SecurityAdvisorFlow struct {
 	UpdateCount      int64                     `json:"UpdateCount"`
 	NodeType         string                    `json:"NodeType,omitempty"`
 	LogStatus        string                    `json:"LogStatus,omitempty"`
+	Extend           map[string]interface{}    `json:"Extend,omitempty"`
 }
 
 // SecurityAdvisorFlowTransformer is a custom transformer for flows
@@ -82,6 +86,7 @@ type securityAdvisorFlowTransformer struct {
 	resolver             Resolver
 	flowUpdateCountCache *cache.Cache
 	excludeStartedFlows  bool
+	extendGremlin        *extendGremlin
 }
 
 func (ft *securityAdvisorFlowTransformer) setUpdateCount(f *flow.Flow) int64 {
@@ -177,7 +182,9 @@ func (ft *securityAdvisorFlowTransformer) Transform(f *flow.Flow) interface{} {
 
 	nodeType, _ := ft.resolver.TIDToType(f.NodeTID)
 
-	return &SecurityAdvisorFlow{
+	extendMap := make(map[string]interface{})
+
+	t := &SecurityAdvisorFlow{
 		UUID:             f.UUID,
 		LinkID:           ft.getLinkID(f),
 		L3TrackingID:     f.L3TrackingID,
@@ -193,5 +200,8 @@ func (ft *securityAdvisorFlowTransformer) Transform(f *flow.Flow) interface{} {
 		Last:             f.Last,
 		UpdateCount:      updateCount,
 		NodeType:         nodeType,
+		Extend:           extendMap,
 	}
+	ft.extendGremlin.Extend(t)
+	return t
 }
--- a/contrib/exporters/secadvisor/mod/transform_test.go
+++ b/contrib/exporters/secadvisor/mod/transform_test.go
@@ -19,6 +19,7 @@ package mod
 
 import (
 	"fmt"
+	"io/ioutil"
 	"runtime"
 	"testing"
 	"time"
@@ -26,9 +27,37 @@ import (
 	cache "github.com/pmylund/go-cache"
 
 	"github.com/skydive-project/skydive/common"
-	"github.com/skydive-project/skydive/contrib/exporters/core"
+	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/flow"
 )
+
+const testConfig = `---
+pipeline:
+  transform:
+    type: sa
+    extend:
+      - CA_Name=G.V().Has('RoutingTables.Src','{{.Network.A}}').Values('Host')
+      - CB_Name=G.V().Has('RoutingTables.Src','{{.Network.B}}').Values('Host')
+`
+
+func initConfig(conf string) error {
+	f, _ := ioutil.TempFile("", "extend_gremlin_test")
+	b := []byte(conf)
+	f.Write(b)
+	f.Close()
+	err := config.InitConfig("file", []string{f.Name()})
+	return err
+}
+
+func (e *extendGremlin) populateExtendGremlin(f *flow.Flow, transformer *securityAdvisorFlowTransformer) {
+	substExprA := "G.V().Has('RoutingTables.Src','" + f.Network.A + "').Values('Host')"
+	substExprB := "G.V().Has('RoutingTables.Src','" + f.Network.B + "').Values('Host')"
+	AName, _ := transformer.resolver.IPToName(f.Network.A, "")
+	BName, _ := transformer.resolver.IPToName(f.Network.B, "")
+	// place substitue expressions into the cache
+	e.gremlinExprCache.Set(substExprA, AName, cache.DefaultExpiration)
+	e.gremlinExprCache.Set(substExprB, BName, cache.DefaultExpiration)
+}
 
 type fakeResolver struct {
 }
@@ -48,11 +77,15 @@ func (c *fakeResolver) TIDToType(nodeTID string) (string, error) {
 	return "fake_node_type", nil
 }
 
-func getTestTransformer() core.Transformer {
+func getTestTransformer() *securityAdvisorFlowTransformer {
+	initConfig(testConfig)
+	cfg := config.GetConfig().Viper
+	newExtend := NewExtendGremlin(cfg)
 	return &securityAdvisorFlowTransformer{
 		flowUpdateCountCache: cache.New(10*time.Minute, 10*time.Minute),
 		resolver:             &fakeResolver{},
 		excludeStartedFlows:  false,
+		extendGremlin:        newExtend,
 	}
 }
 
@@ -64,6 +97,18 @@ func assertEqual(t *testing.T, expected, actual interface{}) {
 			msg += fmt.Sprintf(" on %s:%d", file, no)
 		}
 		t.Fatalf("%s: (expected: %v, actual: %v)", msg, expected, actual)
+	}
+}
+
+func assertEqualExtend(t *testing.T, expected, actual interface{}, suffix string) {
+	actual2 := actual.(map[string]interface{})
+	if expected != actual2[suffix] {
+		msg := "Equal assertion failed"
+		_, file, no, ok := runtime.Caller(1)
+		if ok {
+			msg += fmt.Sprintf(" on %s:%d", file, no)
+		}
+		t.Fatalf("%s: (expected: %v, actual: %v)", msg, expected, actual2[suffix])
 	}
 }
 
@@ -112,9 +157,10 @@ func getFlow() *flow.Flow {
 	}
 }
 
-func Test_Transform_basic_flow(t *testing.T) {
+func TestTransformBasicFlow(t *testing.T) {
 	transformer := getTestTransformer()
 	f := getFlow()
+	transformer.extendGremlin.populateExtendGremlin(f, transformer)
 	secAdvFlow := transformer.Transform(f).(*SecurityAdvisorFlow)
 	assertEqual(t, version, secAdvFlow.Version)
 	assertEqualInt64(t, 0, secAdvFlow.UpdateCount)
@@ -130,18 +176,20 @@ func Test_Transform_basic_flow(t *testing.T) {
 	assertEqual(t, "80", secAdvFlow.Transport.B)
 }
 
-func Test_Transform_UpdateCount_increses(t *testing.T) {
+func TestTransformUpdateCountIncreases(t *testing.T) {
 	transformer := getTestTransformer()
 	f := getFlow()
+	transformer.extendGremlin.populateExtendGremlin(f, transformer)
 	for i := int64(0); i < 10; i++ {
 		secAdvFlow := transformer.Transform(f).(*SecurityAdvisorFlow)
 		assertEqualInt64(t, i, secAdvFlow.UpdateCount)
 	}
 }
 
-func Test_Transform_Status_updates(t *testing.T) {
+func TestTransformStatusUpdates(t *testing.T) {
 	transformer := getTestTransformer()
 	f := getFlow()
+	transformer.extendGremlin.populateExtendGremlin(f, transformer)
 	secAdvFlow := transformer.Transform(f).(*SecurityAdvisorFlow)
 	assertEqual(t, "STARTED", secAdvFlow.Status)
 	secAdvFlow = transformer.Transform(f).(*SecurityAdvisorFlow)
@@ -149,4 +197,15 @@ func Test_Transform_Status_updates(t *testing.T) {
 	f.FinishType = flow.FlowFinishType_TCP_FIN
 	secAdvFlow = transformer.Transform(f).(*SecurityAdvisorFlow)
 	assertEqual(t, "ENDED", secAdvFlow.Status)
+}
+
+func TestExtendGremlin(t *testing.T) {
+	transformer := getTestTransformer()
+	f := getFlow()
+	transformer.extendGremlin.populateExtendGremlin(f, transformer)
+	secAdvFlow := transformer.Transform(f).(*SecurityAdvisorFlow)
+
+	// verify that extension fields were properly created
+	assertEqualExtend(t, "fake_container_one", secAdvFlow.Extend, "CA_Name")
+	assertEqualExtend(t, "fake_container_two", secAdvFlow.Extend, "CB_Name")
 }


contrib/exporters/secadvisor/mod/transform.go
```diff
@@ -38,7 +38,12 @@ func NewTransform(cfg *viper.Viper) (interface{}, error) {
 		return nil, err
 	}
 
-	resolver, err = NewResolveMulti(resolver.(Resolver))
+	resolver2, err := NewResolveVm(cfg)
```
hunchback: I guess you are missing a file - as I don't see the implementation of `NewResolveVM` anywhere ... also if resolution is a specific kind of Vm such as `Xen` then rather call the handler `NewResolveXen`

contrib/exporters/secadvisor/mod/transform.go
```diff
@@ -38,7 +38,12 @@ func NewTransform(cfg *viper.Viper) (interface{}, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	resolver, err = NewResolveMulti(resolver.(Resolver), resolver2.(Resolver))
```
hunchback: now that you have several resolvers then maybe use names such as:
- resolverRunc
- resolverVM (or resolverXen)

etc.

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/skydive-project/skydive/logging"
+	"github.com/skydive-project/skydive/api/client"
+)
+
+// for no transform extensions
+type extendNone struct {
+}
+
+func NewExtendNone(cfg *viper.Viper) (interface{}, error) {
+	logging.GetLogger().Infof("entering NewExtendNone")
```
hunchback: please remove this log message

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+//     extend:
+//       - VAR_NAME1=<gremlin expression with a single substitution string>
+//       - VAR_NAME2=<gremlin expression with a single substitution string>
+type extendGremlin struct {
+	extendStrings []string
+	gremlinClient *client.GremlinQueryHelper
+}
+
+func NewExtendGremlin(cfg *viper.Viper) (interface{}, error) {
+	logging.GetLogger().Infof("entering NewExtendGremlin")
```
hunchback: remove

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+	gremlinClient := client.NewGremlinQueryHelper(CfgAuthOpts(cfg))
+	extend := cfg.GetStringSlice(CfgRoot + "transform.extend")
+	logging.GetLogger().Infof("NewExtendGremlin, extend = %s", extend)
```
hunchback: remove

contrib/exporters/core/pipeline.go
```diff
@@ -110,6 +115,11 @@ func init() {
 	FiltererHandlers = make(HandlersMap)
 	FiltererHandlers.Register("subnet", NewFilterSubnet, true)
 
+	logging.GetLogger().Infof("init: before ExtenderHandlers")
```
hunchback: delete

contrib/exporters/core/pipeline.go
```diff
@@ -110,6 +115,11 @@ func init() {
...
+	ExtenderHandlers = make(HandlersMap)
+	ExtenderHandlers.Register("gremlin", NewExtendGremlin, true)
+	logging.GetLogger().Infof("init: after ExtenderHandlers")
```
hunchback: delete

contrib/exporters/core/pipeline.go
```diff
@@ -152,6 +163,13 @@ func NewPipeline(cfg *viper.Viper) (*Pipeline, error) {
 		return nil, err
 	}
 
+	logging.GetLogger().Infof("NewPipeline, before ExtenderHandlers.Init")
```
hunchback: delete

contrib/exporters/core/pipeline.go
```diff
@@ -152,6 +163,13 @@ func NewPipeline(cfg *viper.Viper) (*Pipeline, error) {
...
+	extender, err := ExtenderHandlers.Init(cfg, "extend")
+	if err != nil {
+		return nil, err
+	}
+	logging.GetLogger().Infof("NewPipeline, after ExtenderHandlers.Init")
```
hunchback: delete

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+                str := substrings[i]
+                jj = ii[str].(map[string]interface{})
+                ii = jj
+        }
+	// TODO: verify that each substring is a valid field so we don't get an exception
+        name := ii[substrings[len(substrings)-1]].(string)
+        return name
+}
+
+// use substitution value and perform gremlin query to obtain the desired data enrichment
```
hunchback: comment must start with `// PerformConversion `

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+        var ii map[string]interface{}
+        var jj map[string]interface{}
+        ii = S
+        for i := 0; i < len(substrings)-1; i++ {
+                str := substrings[i]
+                jj = ii[str].(map[string]interface{})
+                ii = jj
+        }
+	// TODO: verify that each substring is a valid field so we don't get an exception
+        name := ii[substrings[len(substrings)-1]].(string)
```
hunchback: put this inside a block:

```
if len(substrings) > 0 {
         
}
```

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+	result2, _ := e.gremlinClient.Query(result)
+	var result3 []string
+	if err := json.Unmarshal(result2, &result3); err != nil {
+		logging.GetLogger().Errorf("PerformConversion: err = %s", err)
+		return ""
+	}
+        return result3[0]
+}
+
+// main function called by pipeline to perform the extension substitutions
```
hunchback: comment must start with `// Extend ...`

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+// we expect the data of a single flow per call
+func (e *extendGremlin)  Extend(in interface{}) interface{} {
+	// convert the flow information to a map
+	S1, _ := json.Marshal(&in)
```
hunchback: if we are using other encoding forms (other than JSON) such as CSV then we will get incorrect values (such as wrong field names).

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+	var S2 map[string]interface{}
+	S2 = make(map[string]interface{})
+	json.Unmarshal(S1, &S2)
```
hunchback: doing a json.Marshel() and json.Unmarshel() is very expensive. Typically the Unmarshel() takes longer than Marchel() some benchmarks I noticed a 5x markup :-(

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+	// parse the extension string and perform the substitution
+        for _, ee := range e.extendStrings {
+		// TODO do this parsing once at the beginning and save in some data structure
+                substrings := strings.Split(ee, "=")
```
hunchback: you must handling escaping here as the Gremlin expression could contain `=` - for example `g.v().Has("X", "=")`

hunchback: verify that you got exactly 2 strings (key, value)

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+                newSubString := e.PerformConversion(substrings[1], S2)
+		// add new field in map with the desired key name
+                S2[substrings[0]] = newSubString
+        }
+	return S2
+}
+
+
```
hunchback: redundant newline

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+type extendNone struct {
+}
+
+func NewExtendNone(cfg *viper.Viper) (interface{}, error) {
+	logging.GetLogger().Infof("entering NewExtendNone")
+        return &extendNone{}, nil
+}
+
+func (e *extendNone)  Extend(in interface{}) interface{} {
+	return in
```
hunchback: did you run the code through `gofmt -s -w` ? asking as a notice two-spaces before `Extend`

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+// for no transform extensions
+type extendNone struct {
+}
+
+func NewExtendNone(cfg *viper.Viper) (interface{}, error) {
+	logging.GetLogger().Infof("entering NewExtendNone")
+        return &extendNone{}, nil
+}
+
+func (e *extendNone)  Extend(in interface{}) interface{} {
```
hunchback: did you run the code through `gofmt -s -w` ? asking as a notice two-spaces before `Extend`

contrib/exporters/core/extend.go
hunchback: I don't like the fact that the extendNone returns a `struct` while other extends return a `map` - you really want to align the two cases such that for example conversion to a map is done by the pipeline and Extend() signature is:

```
func Extend(map[string]interface{}) map[string]interface{}
```

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+
+// use the substitution string to extract the value from the provided map
+func (e *extendGremlin) DeriveValue(S map[string]interface{}, varString string) string {
+        substrings := strings.Split(varString, ".")
+        var ii map[string]interface{}
+        var jj map[string]interface{}
+        ii = S
+        for i := 0; i < len(substrings)-1; i++ {
+                str := substrings[i]
+                jj = ii[str].(map[string]interface{})
```
hunchback: you must add type assertion:

```
val, ok := ii[str]
// handle !ok
jj, ok := val.(map[string]interface{})
// handle !ok
```

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+//	AA_Name=G.V().Has('RoutingTables.Src',%Network.A%).Values('Host')
+// Substitution string must be surrounded with `%`
+// The substitution string refers to fields that exist in the data provided by the particular transform
+// No spaces allowed
+// Need to define `extend` field as a subfield of `transform`
+// Structure in yml file is as follows:
+//   transform:
+//     type: <xxx>
+//     extend:
+//       - VAR_NAME1=<gremlin expression with a single substitution string>
```
hunchback: calculating a value via a single Gremlin expression will not work in the general case and would even not suffice the SecAdvisor and other cases:

```
// IPToName resolve ip to name
func (r *resolveRunc) IPToName(ipString, nodeTID string) (string, error) {
        node, err := r.gremlinClient.GetNode(g.G.V().Has("Runc.Hosts.IP", ipString).ShortestPathTo(g.Metadata("TID", nodeTID)))
        if err != nil {
                return "", err
        }

        name := node.Metadata["Runc"].(map[string]interface{})["Hosts"].(map[string]interface{})["Hostname"].(string)
        return "0_0_" + name + "_0", nil
}
```



this will not work in the general case - as you may need to run multiple Gremlin for example:

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,127 @@
...
+        }
+	// TODO: verify that each substring is a valid field so we don't get an exception
+        name := ii[substrings[len(substrings)-1]].(string)
+        return name
+}
+
+// use substitution value and perform gremlin query to obtain the desired data enrichment
+func (e *extendGremlin) PerformConversion(origString string, S map[string]interface{}) string {
+	// `%` marks the substitutin string; split the string on `%`
+        substrings := strings.Split(origString, "%")
```
hunchback: this will not work for the general case of multiple substition values. You must use some type of template library such as: https://github.com/Masterminds/sprig

hunchback:
@KalmanMeth - there are some fundamental issues with your approach:

The functionality you offer does not even cover the existing SecAdvisor use case, as for that you work correctly you must also add tidNode value into your query (as IP addresses may not be unique throughout the cluster).

The `encoder` assumes a struct with annotations which then enable it to Marshel to JSON, CSV, <whatever> correctly - by covering to a map (using JSON serialize - you kill CSV and other non-JSON encoders).

You performance is poor - I assume an easy 10x slowdown due to using json.Marshel/json.Unmarshel for convert to a map - an alternative option would  to add a field within SecAdvisorFlow which contains a `map[string]interface{}` and then add fields to there.

contrib/exporters/core/extend.go
KalmanMeth: Multiple Gremlin expressions can be defined, one after another, to define whatever is needed. Do you mean we need to combine multiple Gremlin expressions in a single query? This is a first implementation with initial functionality. We should discuss what the additional requirements are and see how to address them with improvements over time.

contrib/exporters/core/extend.go
KalmanMeth: Is this a requirement to have multiple substitutions in a single query? We can expand the functionality over time to cover this case, if needed.

KalmanMeth: Is this a requirement to have multiple substitutions in a single query? We can expand the functionality over time to cover this case, if needed.

contrib/exporters/secadvisor/mod/transform.go
KalmanMeth: You are right. I forgot to add the file resolve_vm.go.

contrib/exporters/secadvisor/mod/transform.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/pipeline.go
KalmanMeth: OK

contrib/exporters/core/pipeline.go
KalmanMeth: OK

contrib/exporters/core/pipeline.go
KalmanMeth: OK

contrib/exporters/core/pipeline.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: added check earlier that len > 0.

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: As noted elsewhere, there is overhead in converting everything to a map in order to perform the substitution, so we want to avoid it if the conversion is unnecessary. Let's discuss how we want to handle this.

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: To calculate this particular name (not including the prefix 0_0_, etc), you can do:
G.V().Has("Runc.Hosts.IP", ipString).ShortestPathTo(g.Metadata("TID", nodeTID)).Values('Runc.Hosts.Hostname)

contrib/exporters/core/extend.go
hunchback: its' mandatory to be able to express the current resolvers (e.g. runC) using your queries. And as the runC requires both tidNode and IP you must support these. More generally using split according to `%` to replace even a single argument is a very bad idea as it does not account `%` which may be legitimately embedded into the query.

contrib/exporters/core/extend.go
hunchback: ok. For now lets' see if we can cover needed functionality w/o requiring multiple Gremlin queries.

contrib/exporters/core/extend.go
hunchback: the alternative suggestion (which helps also solve other issues) is to add an extra field into the SecAdvisorFlow struct which contains a map (this field could be serialized in json both muted for other encoders).

contrib/exporters/core/extend.go
hunchback: ok

contrib/exporters/core/extend.go
KalmanMeth: Implementation revised to return the original SecurityAdvisorFlow struct.

contrib/exporters/core/extend.go
KalmanMeth: We now avoid the marshel/Unmarshel by using the template abstraction.

contrib/exporters/core/extend.go
KalmanMeth: We now split on the first '=', so that any additional '=' is included in the gremlin expression.

contrib/exporters/core/extend.go
KalmanMeth: We now return the original SecurityAdvisorFlow struct as before.

contrib/exporters/core/extend.go
KalmanMeth: We now use templates, so this should be OK.

KalmanMeth:
We now use the template abstraction, as suggested. This avoids the Marshel/Unmarshel. Caching was added. We should add the ability for the user to guide us whether we may use caching.

hunchback:
@KalmanMeth - please squash all commits together its' also fine if you split your work  to several commits as long as each commit represents a logical step and does not introduce degredations (thus is stable)

hunchback:
@KalmanMeth - regarding testing you can generate a configuration object within your code as demonstrated here https://github.com/skydive-project/skydive/blob/fe4bd6815e1533e0027f0f9b32829f98624f3e91/tests/tests.go#L59

KalmanMeth:
I modified transform_test.go to include support for the extensions with misc fields and added a unit test to see that the substitution is done properly.

eranra:
@dubek  --> I want to make sure that the new feature that @KalmanMeth is adding in this PR supports the new format that David told me you are implementing for SA. Can you (a) review this PR and (b) work with Kalman on configuration for this PR that fits the new extended fields in format for SA. THX

dubek:
I'm on vacation, I'll take a look next week.

contrib/exporters/secadvisor/mod/transform.go
hunchback: you have not added NewResolveVm() to this commit ... so this will not build

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
+/*
```
hunchback: the commit comment is a mess ... please fix

contrib/exporters/core/store_s3.go
```diff
@@ -202,6 +203,7 @@ func (s *StoreS3) flushFlowsToObject(t Tag, endTime time.Time) error {
 		return err
 	}
 
+	// XXX use .gz suffix only if the compression was gzip
```
hunchback: rather use FIXME (not XXX)

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
...
+ *
+ */
+
+package core
+
+import (
+	"fmt"
+	"time"
+	"encoding/json"
+	"github.com/skydive-project/skydive/api/client"
```
hunchback: you want to place an empty line between std imports and others

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
...
+ * limitations under the License.
+ *
+ */
+
+package core
+
+import (
+	"fmt"
+	"time"
+	"encoding/json"
```
hunchback: order is wrong ..
 please run through gofmt

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/spf13/viper"
+	"strings"
+	cache "github.com/pmylund/go-cache"
+)
+
+// extendNone is used for no transform extensions
+type extendNone struct {
+}
+
+func NewExtendNone(cfg *viper.Viper) (interface{}, error) {
```
hunchback: add comment - // NewExtendNone ....

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
...
+	return &extendNone{}, nil
+}
+
+func (e *extendNone) Extend(in interface{}) interface{} {
+	return in
+}
+
+// extendGremlin is used for transform extensions using gremlin expressions
+// Syntax supported is:
+//	VAR_NAME=<gremlin expression with a single substitution string>
```
hunchback: wrong

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
...
+// For example:
+//	AA_Name=G.V().Has('RoutingTables.Src',%Network.A%).Values('Host')
```
hunchback: wrong

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
...
+	p := &extendGremlin{
+		extendStrings: extend,
+		gremlinClient: gremlinClient,
+		gremlinExprCache: cache.New(10*time.Minute, 10*time.Minute),
+	}
+	return p, nil
+}
+
+// DeriveValue: use the substitution string to extract the value from the provided map
+func (e *extendGremlin) DeriveValue(S map[string]interface{}, varString string) (string, error) {
```
hunchback: should be a private method deriveValue()

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	var ii map[string]interface{}
+	var jj map[string]interface{}
+	var ok bool
+	ii = S
+	for i := 0; i < len(substrings)-1; i++ {
+		str := substrings[i]
+		jj, ok = ii[str].(map[string]interface{})
+		if !ok {
+			logging.GetLogger().Errorf("substring %s extended gremlin not found", str)
```
hunchback: remove log if u are returning error to caller

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
...
+			return "", fmt.Errorf("substring %s extended gremlin not found", str)
+		}
+		ii = jj
+	}
+	name, ok := ii[substrings[len(substrings)-1]].(string)
+	if !ok {
+		logging.GetLogger().Errorf("substring %s extended gremlin not found", substrings[len(substrings)-1])
```
hunchback: remove log if u are returning error to caller

contrib/exporters/secadvisor/mod/transform_test.go
```diff
@@ -130,9 +203,10 @@ func Test_Transform_basic_flow(t *testing.T) {
 	assertEqual(t, "80", secAdvFlow.Transport.B)
 }
 
-func Test_Transform_UpdateCount_increses(t *testing.T) {
+func Test_Transform_UpdateCount_increases(t *testing.T) {
```
hunchback: in go the standard notation is CamelCase so name should rather be TestTransformUpdateCountIncreases()

contrib/exporters/secadvisor/mod/transform_test.go
```diff
@@ -150,3 +225,15 @@ func Test_Transform_Status_updates(t *testing.T) {
 	secAdvFlow = transformer.Transform(f).(*SecurityAdvisorFlow)
 	assertEqual(t, "ENDED", secAdvFlow.Status)
 }
+
+func Test_Extend_Gremlin(t *testing.T) {
```
hunchback: same

contrib/exporters/secadvisor/mod/transform_test.go
```diff
@@ -22,14 +22,70 @@ import (
...
+func initConfig(conf string) error {
+	f, _ := ioutil.TempFile("", "extend_gremlin_test")
+	b := []byte(conf)
+	f.Write(b)
+	f.Close()
+	err := config.InitConfig("file", []string{f.Name()})
+	return err
+}
+
+func (e *extendGremlin) populateExtendGremlin(f *flow.Flow, transformer *securityAdvisorFlowTransformer) {
```
hunchback: make a helper func testResolveToHost() to be called for A name and B name

contrib/exporters/secadvisor/mod/transform_test.go
```diff
@@ -22,14 +22,70 @@ import (
...
 	cache "github.com/pmylund/go-cache"
-
+	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/common"
-	"github.com/skydive-project/skydive/contrib/exporters/core"
+	//"github.com/skydive-project/skydive/contrib/exporters/core"
 	"github.com/skydive-project/skydive/flow"
 )
 
+const testConfig = `---
```
hunchback: you can remove all but the pipeline.transform section

contrib/exporters/secadvisor/mod/transform_test.go
```diff
@@ -22,14 +22,70 @@ import (
...
 	"testing"
 	"time"
-
+	"io/ioutil"
 	cache "github.com/pmylund/go-cache"
-
+	"github.com/skydive-project/skydive/config"
 	"github.com/skydive-project/skydive/common"
-	"github.com/skydive-project/skydive/contrib/exporters/core"
+	//"github.com/skydive-project/skydive/contrib/exporters/core"
```
hunchback: delete

contrib/exporters/core/extend.go
```diff
@@ -0,0 +1,160 @@
...
+		return "", fmt.Errorf("substring %s extended gremlin not found", substrings[len(substrings)-1])
+	}
+	return name, nil
+}
+
+// PerformConversion: use substitution value and perform gremlin query to obtain the desired data enrichment
+func (e *extendGremlin) PerformConversion(origString string, S map[string]interface{}) (string, error) {
```
hunchback: remove func

contrib/exporters/core/extend.go
KalmanMeth: This is an old version of the code. Not sure why the new code doesn't show up here.

contrib/exporters/core/extend.go
KalmanMeth: This file (core/extend.go) should have been removed. Please ignore code in this file.

contrib/exporters/core/store_s3.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: This file (core/extend.go) should have been removed. Please ignore code in this file.

contrib/exporters/core/extend.go
KalmanMeth: This file (core/extend.go) should have been removed. Please ignore code in this file.

contrib/exporters/core/extend.go
KalmanMeth: This file (core/extend.go) should have been removed. Please ignore code in this file.

contrib/exporters/core/extend.go
KalmanMeth: This file (core/extend.go) should have been removed. Please ignore code in this file.

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/secadvisor/mod/transform_test.go
KalmanMeth: OK

contrib/exporters/secadvisor/mod/transform_test.go
KalmanMeth: OK

contrib/exporters/secadvisor/mod/resolve_vm.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if node == nil {
+		return "", common.ErrNotFound
+	}
+
+	return node.Host, nil
+}
+
+// IPToName resolve tid to type
```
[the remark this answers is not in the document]
KalmanMeth: OK

contrib/exporters/core/extend.go
KalmanMeth: This file (core/extend.go) should have been removed. Please ignore code in this file.

contrib/exporters/core/extend.go
KalmanMeth: OK

contrib/exporters/secadvisor/mod/transform_test.go
KalmanMeth: Not sure what you mean here.

hunchback: notice that you are performing the same 3 operations per Network.A and Network.B - rather use a helper function

hunchback:
@KalmanMeth - please organize you commits - I expect to still have only 3 commits - also please fix the commit comments

contrib/exporters/secadvisor/mod/resolve_runc.go
```diff
@@ -44,6 +45,9 @@ func (r *resolveRunc) IPToName(ipString, nodeTID string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	if (node == nil) {
```
hunchback: redundant as gremlinClinet.GetNode() already addresses this case

hunchback:
please rebase

contrib/exporters/secadvisor/mod/extend_gremlin.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+
+	cache "github.com/pmylund/go-cache"
+	"github.com/skydive-project/skydive/api/client"
+	"github.com/skydive-project/skydive/contrib/exporters/core"
+	"github.com/skydive-project/skydive/logging"
+	"github.com/spf13/viper"
+)
+
+// extendGremlin is used for transform extensions using gremlin expressions
```
hunchback: I don't think you need such a big comment - what is obvious from the code does not require explanations

hunchback:
@KalmanMeth - please split into two PRs:
- adding VM resolver
- adding misc extension of fields
Each will be addressed separately

dubek:
One more idea:

For the key names (variable names) you currently use strings which will be keys under `"Extend"`. Instead, maybe we can give a JSONPath like:  `.Extend.my_var_name`. This will also allow to add stuff in other places in the JSON, for example in: `.Context.A.Image` or wherever.

contrib/exporters/secadvisor/mod/transform.go
KalmanMeth: done

contrib/exporters/secadvisor/mod/resolve_runc.go
KalmanMeth: OK

hunchback:
please remove the VM resolver from here

KalmanMeth:
The VM resolver appears because this branch is checked-out from the resolve_vm branch, which has this "adding VM resolver" commit.

KalmanMeth:
If I remove it, I have to rebase again after the resolve_vm PR is merged.

hunchback:
> If I remove it, I have to rebase again after the resolve_vm PR is merged.

no. these efforts are disjoint, remove the resolve-vm commit and rebase on origin/master

KalmanMeth:
I removed the relove_vm from this PR.

KalmanMeth:
PR moved to skydive-flow-exporters repository, PR 18. 
https://github.com/skydive-project/skydive-flow-exporter/pull/18

Issue #1930, opened 2019-07-28
pipeline: building skydive dependencies


hunchback:
currently to build a pipeline one need to run:

```
cd $GOPATH/src/github.com/skydive-project/skydive
make
make -C contrib/pipelines/secadvisor
```

Note that there we can't skip the `make` as the required skydive modules require running of:
- easyjson
- common.Getter code generation
- etc.

So currently if we would want to main a pipeline outside of skydive we would have to build skydive and include some of the generated files ... is there an alternative to this?

For example we could maintain a non-optimized standard build option:
- standard json rather than easyjson
- common.Getter code using reflection
- etc

hunchback:
@eranra / @liranschour - please elaborate ...

safchain:
`make contrib.objectstore` handle dependencies

safchain:
easyjson files should be easy to generate with go generate. For protobuf files it seems less obvious we need to investigate a bit on this...

PR #1956, opened 2019-08-14
Gogenerate easyjson

hunchback:


nplanel:
@hunchback this will not work as we using a vendorized version of easyjson

hunchback:
> @hunchback this will not work as we using a vendorized version of easyjson

can we mandate easyjson to exist on host? and can we install it as a tool using `go get ...`?

lebauce:
@hunchback I would also prefer to use `go generate` to run easyjson. The problem is mainly due to packaging : on RPM building hosts, there is no Internet connectivity so you can't do a `go get`, that's why we do vendorize `govendor` and `easyjson`. IIRC, the idea behind the `VENDOR_RUN` trick is that we did not want to override a `govendor` or an `easyjson` that the user would have installed using a regular `go get` but I think that's not a big problem.

To sum up, I think we could still vendorize govendor and easyjson as we do today but install them in the regular location ($GOPATH/bin instead of $GOPATH/src/github.com/skydive-project/skydive) and use go:generate to run easyjson. Does it makes sense ?

hunchback:
@lebauce - the go-elasticsearch go library also uses easyjson - they handle dependencies using `go.mod`: https://github.com/elastic/go-elasticsearch/search?q=easyjson&unscoped_q=easyjson what do you think about this?

nplanel:
@lebauce @hunchback After the weekly meeting today, we should consider generating a versionized skydive toolchain binaries. As it could be used by external libraries that relay for example on flow.pb.go ... this may speedup a little bit the build process as well.

safchain:
Just to be sure we are aligned :) The main goal of this is to allow users to use Skydive libs like websocket like any other packages. #1930

```
go get github.com/skydive-project/skydive/websocket/...
```

So having any generated files, protobuf or easyjson generated by go itself.

hunchback:
@nplanel the following should work:

```
cd $GOPATH
cd src/github.com
mkdir dummy
cd dummy
go mod init
cat <<EOF > main.go

package main

import (
        "github.com/skydive-project/skydive/contrib/exporters/core"
)

func main() {
        core.Main("/etc/skydive/dummy.yml")
}
EOF
go build
```

hunchback:
replaced by https://github.com/skydive-project/skydive/pull/1984

PR #1967, opened 2019-08-30
Fix Flow-Validation workflow : Add Capture Validator

dvandra:
Add Capture Validator : If any Node/Interface is DOWN - Capture will not be created on that node and no flows will be captured on that node.

### DIFF
--- a/statics/workflows/flow-validation.yaml
+++ b/statics/workflows/flow-validation.yaml
@@ -4,22 +4,23 @@ Name: "FlowValidation"
 Title: "Check Flow-Forwarding form Checkpoint-1 to Checkpoint-2"
 Abstract: "This workflow aims to validate the Flow-Forwarding on particular checkpoints. It returns the status of flow validation, true or false."
 Description: >
-    # How to use
-      1. To validate on just one checkpoint select it as Checckpoint-1
-      2. To validate on two checkpoints select it as Checkpoint-1 and Checkpoint-2
-      3. To validate on all nodes between Checkpoint-1 and Checkpoint-2, Select the checkbox given below Checkpoint-2
-      4. To validate some specific type of traffic/flows, Select the type of protocol - icmp/tcp/udp
-      5. You can also provide more details like Source IP Address, Destination IP Address, ICMP-ID/Destination Port No. to validate more specific type of traffic/flows
-      6. At the end provide the time-interval(in seconds), Till how long you want to validate the flows, Starts when you hit the 'Execute' button
-      7. As a result you will get Flow-validation status as 'Connectivity = true' OR 'Connectivity = false and NotReachedNodes = Nodes (With no flow captured/inappropriate flow-metric)'
-      8. Hit 'Execute' button to run the workflow
-    # How It Works
-      1. Starts Capture on selected Checkpoint/Checkpoints using the BPF-Filter
-      2. For BPF-Filter it will use all other provided information like - protocol, Source IP Address, Destination IP Address, ICMP-ID/Destination Port No (Just to capture specific type of flows)
-      3. Now it will check that all nodes having capture has flows with the capture-id of the capture created by this workflow
-      4. If all nodes have appropriate flows, then it will check for flow-metric (Like-'AB-Pkts and BA-Pkts > 0')
-      5. If it is Ok then gives result = true, else result = false and return the nodes with no flows/inappropriate flows/flow-metric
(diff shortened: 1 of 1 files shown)


dvandra:
run skydive-scale-tests

dvandra:

dvandra:

statics/workflows/flow-validation.yaml
```diff
@@ -92,38 +97,66 @@ Source: |
...
+        return noFlows
       }
 
-      if (!flowExpected.length) {
-        result["Error"] = "No node with capture"
-      } else if (noflows.length == 0) {
-        result["Connectivity"] = true
-      } else {
+      var noCapture = []
+      var flowExpected = client.gremlin.G().V().Has('Captures.ID', capture.UUID, 'Type', NE('ovsport')).result();
```
safchain: does the following gremlin expression not enough to ignore `DOWN` interfaces ?

```
client.gremlin.G().V().Has('Captures.ID', capture.UUID, 'Type', NE('ovsport'), 'State', 'DOWN')
```

dvandra: Yes, but with this it will end up giving all other nodes followed by the node (with 'State : DOWN') without including the node (with 'State : DOWN').

safchain: Not sure to understand your answer, but I'm even not sure we need to do something :)
I'm saying that because if the state is `DOWN` the capture won't start and the node won't have the `Capture.ID` field, meaning that we will get only the node being able to capture, those we want to check.

statics/workflows/flow-validation.yaml
```diff
@@ -92,38 +97,66 @@ Source: |
...
+
+      if (flowExpected.length != captureExpected.length) {
+        noCapture = captureValidation(captureExpected)
+      }
+
+      if (noCapture.length != 0) {
         result["Connectivity"] = false
-        result["NotReachedNodes"] = noflows
+        result["NotCapturedNodes"] = noCapture
+      } else {
```
safchain: I would always return `NotReachedNodes` merging noCapture and noFlows and I would return NotCapturedNodes as an information. So it will be like

```
result['NotCapturedNodes'] = noCapture
result['NotReachNodes'] = noFlows + noCapture
```

In fact `DOWN` interface didn't see the flow thus "not reached"

statics/workflows/flow-validation.yaml
```diff
@@ -92,38 +97,66 @@ Source: |
...
+
+        if (protocol != "udp" && flowExpected.length && flowExpected.length == flowCaptured.length) {
+          flowCaptured = client.gremlin.G().Flows().Has('CaptureID', capture.UUID, 'Metric.ABPackets', GT(0), 'Metric.BAPackets', GT(0)).CaptureNode().Dedup().result();
+        }
+
+        if (flowExpected.length != flowCaptured.length) {
+          noFlows = flowValidation(flowExpected, flowCaptured)
+        }
+
+        if (!flowExpected.length) {
```
safchain: this could be moved just after the request line 131

statics/workflows/flow-validation.yaml
```diff
@@ -92,38 +97,66 @@ Source: |
...
+        result["NotCapturedNodes"] = noCapture
+      } else {
+        var noFlows = [];
+        var flowCaptured = client.gremlin.G().Flows().Has('CaptureID', capture.UUID).CaptureNode().Dedup().result();
+
+        if (protocol != "udp" && flowExpected.length && flowExpected.length == flowCaptured.length) {
+          flowCaptured = client.gremlin.G().Flows().Has('CaptureID', capture.UUID, 'Metric.ABPackets', GT(0), 'Metric.BAPackets', GT(0)).CaptureNode().Dedup().result();
+        }
+
+        if (flowExpected.length != flowCaptured.length) {
```
safchain: I think we need to remove this check as we avoid the check of the ID if the length are equals

statics/workflows/flow-validation.yaml
dvandra: Yes, we can merge noCapture with noFlows.
Sounds Good!

dvandra:
run skydive-export-sa-tests

dvandra:
skydive-export-sa-tests

dvandra:
run skydive-export-sa-tests

dvandra:
run skydive-export-sa-tests

PR #1970, opened 2019-09-02
contrib/exporters: Add localfiles storer for debugging

dubek:
Storing transformed flow objects in a local directory eases debugging
process for exporters.

To use this storer, use the following configuration for the exporter
main process:

    pipeline:
      store:
        type: localfiles
        localfiles:
          base_dir: /tmp/flows
          max_seconds_per_stream: 3600

The directory structure underneath the `base_dir` is identical to the
one used in S3 (Object Store).

Each file is accompanied by a `.metadata.json` file which holds the
fields that are stored as metadata headers in S3.

For example:

```
$ find /tmp/flows -type f
/tmp/flows/egress/20190902T080845Z/00000000.gz
/tmp/flows/egress/20190902T080845Z/00000000.gz.metadata.json
/tmp/flows/internal/20190902T080645Z/00000000.gz
/tmp/flows/internal/20190902T080645Z/00000000.gz.metadata.json
```

@hunchback please review.

### DIFF
--- a/contrib/exporters/core/pipeline.go
+++ b/contrib/exporters/core/pipeline.go
@@ -121,6 +121,7 @@ func init() {
 	StorerHandlers = make(HandlersMap)
 	StorerHandlers.Register("stdout", NewStoreStdout, true)
 	StorerHandlers.Register("s3", NewStoreS3, false)
+	StorerHandlers.Register("localfiles", NewStoreLocalFiles, false)
 }
 
 // Pipeline manager
--- a/contrib/exporters/core/store_localfiles.go
+++ b/contrib/exporters/core/store_localfiles.go
@@ -1,0 +1,172 @@
+/*
+ * Copyright (C) 2019 IBM, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy ofthe License at
+ *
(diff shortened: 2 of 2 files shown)


contrib/exporters/core/pipeline.go
```diff
@@ -121,6 +121,7 @@ func init() {
 	StorerHandlers = make(HandlersMap)
 	StorerHandlers.Register("stdout", NewStoreStdout, true)
 	StorerHandlers.Register("s3", NewStoreS3, false)
+	StorerHandlers.Register("localfiles", NewStoreLocalFiles, false)
```
hunchback: change `localfiles` to `file`

hunchback:
the problem with this implementation is that it suggests we need to split the current s3 store into:
- the part which decides how many flows go into an object/file --> then then make this into a destinct phase in the Pipeline --> Batcher
- the part which does the actual store to S3

Once this is implemented then the actual store to a file is very simplified and does not introduce code duplication.

hunchback:
@dubek - please let me know if https://github.com/skydive-project/skydive/pull/1974 can help you ... basically it would enable you to implement a `writeFile` and use the `storeBuffered` which contains all the buffering; naming etc.

dubek:
@hunchback yes that makes sense; I'll wait for the #1974 refactoring and then try this again. Closing.

PR #1971, opened 2019-09-02
contrib/exporters/secadvisor: Encode flows in a JSON object

dubek:
During the refactoring to pipelines the export format for flows was
changed to JSON array, whereas Security Advisor expects the objects in
COS/S3 to be a JSON object with a top-level "data" key whose value is an
array of flow objects.

This change adds a `secadvisor_json` encode type which adds this
top-level element (and uses the core JSON encoder to perform the actual
encoding, so `pretty: true` still works).

To use this encoder, the secadvisor exporter should have this config:

    pipeline:
      encode:
        type: secadvisor_json
        json:
          pretty: true

@hunchback please review.


contrib/exporters/secadvisor/main.go
```diff
@@ -27,5 +27,6 @@ func main() {
 }
 
 func init() {
+	core.EncoderHandlers.Register("secadvisor_json", mod.NewEncodeSecadvsiorJSON, false)
```
hunchback: just call this encoder "sa"

hunchback: register it as the *default* encoder (2nd argument should be true)

hunchback: you also want to register this to exporters/uber so that it is available in the single unified binary

contrib/exporters/secadvisor/mod/encode.go
```diff
@@ -0,0 +1,47 @@
...
+
+package mod
+
+import (
+	"github.com/spf13/viper"
+
+	"github.com/skydive-project/skydive/contrib/exporters/core"
+)
+
+type encodeSecadvisorJSON struct {
```
hunchback: you can just call it `encode` as:
(1) it already is part of secadvisor code base
(2) today it is JSON but tomorrow it may change something else (CSV, Parquet)

contrib/exporters/secadvisor/mod/encode.go
```diff
@@ -0,0 +1,47 @@
...
+	jsonEncoder core.Encoder
+}
+
+type secadvisorTopLevelObject struct {
```
hunchback: no need for secadvisor prefix ... any code here is secadvisor

contrib/exporters/secadvisor/mod/encode.go
```diff
@@ -0,0 +1,47 @@
...
+package mod
+
+import (
+	"github.com/spf13/viper"
+
+	"github.com/skydive-project/skydive/contrib/exporters/core"
+)
+
+type encodeSecadvisorJSON struct {
+	jsonEncoder core.Encoder
```
hunchback: a more elegant implementation is to replace:

```
jsonEncoder core.Encoder
```

with:

```
*core.EncodeJSON
```

contrib/exporters/secadvisor/mod/encode_test.go
```diff
@@ -0,0 +1,62 @@
...
+		&SecurityAdvisorFlow{
+			Status: "ENDED",
+			Last: 5678,
+		},
+	}
+	result, err := getEncoder(t).Encode(in)
+	if err != nil {
+		t.Fatalf("Encode returned unexpected error: %v", err)
+	}
+	assertEqual(t, `{"data":[{"Status":"STARTED","Start":1234,"Last":0,"UpdateCount":0},{"Status":"ENDED","Start":0,"Last":5678,"UpdateCount":0}]}` + "\n", string(result))
```
hunchback: you must check for JSON equality (not string equality) - for example if I change default for JSON encoder to pretty - I don't want to break this test.

For that end you can use something like:


```
func AreEqualJSON(s1, s2 string) (bool, error) {
	var o1 interface{}
	var o2 interface{}

	var err error
	err = json.Unmarshal([]byte(s1), &o1)
	if err != nil {
		return false, fmt.Errorf("Error mashalling string 1 :: %s", err.Error())
	}
	err = json.Unmarshal([]byte(s2), &o2)
	if err != nil {
		return false, fmt.Errorf("Error mashalling string 2 :: %s", err.Error())
	}

	return reflect.DeepEqual(o1, o2), nil
}
```

hunchback: format expected output string to:

```
expected := \
`{  
   "data":[  
      {  
         "Status":"STARTED",
         "Start":1234,
         "Last":0,
         "UpdateCount":0
      },
      {  
         "Status":"ENDED",
         "Start":0,
         "Last":5678,
         "UpdateCount":0
      }
   ]
}`
```

contrib/exporters/secadvisor/secadvisor.yml.default
```diff
@@ -27,7 +27,7 @@ pipeline:
     sa:
       exclude_started_flows: true
   encode:
-    type: json
+    type: secadvisor_json
```
hunchback: change to `sa`

contrib/exporters/secadvisor/main.go
dubek: Done, thanks.

I agree here to use `"sa"` for consistency, but generally I would not use this unclear abbreviation and would opt instead for `secadvisor` and similar. Consider for future.

hunchback: I am fine with using "secadvisor" consistently throughout

hunchback:
Please also fix:

```
[{"linter":"golint","severity":"warning","path":"contrib/exporters/core/encode.go","line":55,"col":6,"message":"exported type EncodeCSV should have comment or be unexported"},
  {"linter":"golint","severity":"warning","path":"contrib/exporters/core/encode.go","line":28,"col":6,"message":"exported type EncodeJSON should have comment or be unexported"}
]```

contrib/exporters/secadvisor/mod/encode.go
dubek: done.

contrib/exporters/secadvisor/main.go
dubek: done in 25308601b7fd97ec265b37730f23a32b4b43b828.

hunchback:
@dubek - please rebase

dubek:
@hunchback - rebased and force-pushed. See the 3 different commits here (eventually modified all "sa" to "secadvisor" throughout the code and config files).

dubek:
@hunchback - We want to get this merged for Security Advisor (relies on the toplevel `"data"` object key). Do you prefer that I remove the rename of `sa` to `secadvisor` from this PR?

dubek:

PR #1974, opened 2019-09-04
exporters: split storer and writer functions

hunchback:


contrib/exporters/core/store_buffered.go
```diff
@@ -40,16 +39,15 @@ type stream struct {
...
 }
 
-// StoreS3 allows writing flows to an object storage service
-type StoreS3 struct {
-	bucket            string
-	objectPrefix      string
+// storeBuffered allows writing flows to an object storage service
+type storeBuffered struct {
+	dirname           string
+	filenamePrefix    string
```
dubek: Feels to me that `dirname` and `filenamePrefix` don't belong here.

Instead they can be part of the s3 writer config (named `bucket` and `object_prefix` as before), and for stdout writer you don't need them at all.

hunchback: no. they are part of of the "logic" (store) not the actual write operation. In this why you can debug more of the ObjectStore behavior (including bucket/object names) by replacing the s3 writer to file writer

hunchback:
bad branch name - reopening  a different PR

PR #1996, opened 2019-09-12
Adds BESS topology probe

trozet:
Berkley Extensible Software Switch (BESS) is a modular network data
plane. This patch adds topology support for BESS instances running on
different hosts through an agent probe. Adding metrics and
monitoring/tcpdump is still WIP and will be added in a follow up patch.

Signed-off-by: Tim Rozet <trozet@redhat.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/skydive-project/skydive
 
 require (
+        github.com/nimbess/nimbess-agent v0.0.0-20190919205041-4e6f317ac4fd
 	git.fd.io/govpp.git v0.0.0-20190321220742-345201eedce4
 	github.com/GehirnInc/crypt v0.0.0-20170404120257-5a3fafaa7c86
 	github.com/IBM/ibm-cos-sdk-go v0.0.0-20190328184230-08c1143e8d36
--- a/topology/probes/bess/bess.go
+++ b/topology/probes/bess/bess.go
@@ -1,0 +1,251 @@
+// Copyright (c) 2019 Red Hat and/or its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bess
+
+import (
+	"context"
+	"fmt"
+	"github.com/skydive-project/skydive/filters"
+	"google.golang.org/grpc"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/nimbess/nimbess-agent/pkg/proto/bess_pb"
+
+	"github.com/skydive-project/skydive/common"
+	"github.com/skydive-project/skydive/graffiti/graph"
+	"github.com/skydive-project/skydive/probe"
+	"github.com/skydive-project/skydive/topology"
+	"github.com/skydive-project/skydive/topology/probes"
+)
+
+const BessManager = "BESS"
+
+type Probe struct {
+	graph.DefaultGraphListener
+	Ctx         probes.Context
+	bessClient  bess_pb.BESSControlClient
+	state       common.ServiceState
+	bessPort    int
+	bessHost    string
+	bessContext context.Context
+}
+
+func (p *Probe) walkAndRender(moduleName string, parentNode *graph.Node, edgeList *[]*graph.Edge) {
+	modRes, err := p.bessClient.GetModuleInfo(p.bessContext, &bess_pb.GetModuleInfoRequest{Name: moduleName})
+	if err != nil {
+		p.Ctx.Logger.Errorf("Error getting module information for module: %s", moduleName)
+		return
+	}
+
+	nodeID := graph.GenID(moduleName)
+	node := p.Ctx.Graph.GetNode(nodeID)
+	if node == nil {
+		node, err = p.Ctx.Graph.NewNode(nodeID, graph.Metadata{"Name": moduleName,
+			"Type": modRes.GetMclass(), "Manager": BessManager})
+		p.Ctx.Logger.Infof("Creating node: %s", moduleName)
+		if err != nil {
+			p.Ctx.Logger.Errorf("Error creating node: %s, %v", moduleName, err)
+			return
+		}
+	}
+	if parentNode != nil && !topology.HaveOwnershipLink(p.Ctx.Graph, parentNode, node) {
+		topology.AddOwnershipLink(p.Ctx.Graph, parentNode, node, nil)
+	}
+	for _, oGate := range modRes.GetOgates() {
+		p.walkAndRender(oGate.GetName(), parentNode, edgeList)
+		edgeID := graph.GenID(moduleName, string(oGate.Igate), string(oGate.Ogate), oGate.GetName())
+		if p.Ctx.Graph.GetEdge(edgeID) == nil {
+			p.Ctx.Graph.NewEdge(edgeID, node,
+				p.Ctx.Graph.GetNode(graph.GenID(oGate.GetName())),
+				graph.Metadata{"RelationType": "layer2", "Directed": true, "Manager": BessManager})
+		}
+		*edgeList = append(*edgeList, p.Ctx.Graph.GetEdge(edgeID))
+	}
+
+}
+
+func (p *Probe) Do(ctx context.Context, wg *sync.WaitGroup) error {
+	p.Ctx.Logger.Info("Connecting to BESS")
+	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", p.bessHost, p.bessPort), grpc.WithInsecure())
+	if err != nil {
+		return err
+	}
+	p.Ctx.Logger.Info("Connected to BESS")
+	p.bessClient = bess_pb.NewBESSControlClient(conn)
+	verResponse, err := p.bessClient.GetVersion(context.Background(), &bess_pb.EmptyRequest{})
+	if err != nil {
+		return err
+	} else if verResponse.Error != nil {
+		return fmt.Errorf("could not get version: %s, %s", verResponse.GetError(), err)
+	}
+	p.Ctx.Logger.Infof("BESS connected with version: %s", verResponse.Version)
+
+	metadata := graph.Metadata{
+		"Name":    "bess",
+		"Type":    "bess",
+		"Program": "bessd",
+		"Version": verResponse.GetVersion(),
+		"Manager": BessManager,
+	}
+	wg.Add(1)
+
+	go func() {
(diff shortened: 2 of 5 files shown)


topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+				graph.Metadata{"RelationType": "layer2", "Directed": true, "Backend": BessBackend})
+		}
+		*edgeList = append(*edgeList, p.Ctx.Graph.GetEdge(edgeID))
+	}
+
+}
+
+func (p *Probe) run() {
+	p.state.Store(common.RunningState)
+	fmt.Println("Starting BESS Probe...")
```
safchain: nit: remove this line

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+	logging.GetLogger().Info("Starting BESS Probe")
+	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", p.bessHost, p.bessPort), grpc.WithInsecure())
+	if err != nil {
+		log.Fatalf("Did not connect: %v", err)
```
safchain: Usually probe don't panic or exit, they just log.Error and keep trying to reconnect. I would suggest to use a `NodeProbeWrapper` which can handle reconnection loop.
https://github.com/skydive-project/skydive/blob/master/topology/probes/wrapper.go#L125

You can have a look how the `docker` probe is using it. Basically you run function will be replaced by a `Do` function

nit: use `%s` for error

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	defer conn.Close()
+	log.Info("Connected to BESS")
+	p.bessClient = bess_pb.NewBESSControlClient(conn)
+	verResponse, err := p.bessClient.GetVersion(context.Background(), &bess_pb.EmptyRequest{})
+	if err != nil || verResponse.Error != nil {
+		log.Fatalf("Could not get version: %v, %v", verResponse.GetError(), err)
```
safchain: same here

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+		"Program": "bessd",
+		"Version": verResponse.GetVersion(),
+		"Backend": BessBackend,
+	}
+
+	// get modules and data from bess
+	for p.state.Load() == common.RunningState {
+		modsRes, err := p.bessClient.ListModules(p.bessContext, &bess_pb.EmptyRequest{})
+		if err != nil {
+			log.Errorf("Unable to query module list for bess: %v", err)
```
safchain: a continue is not needed here ? can we proceed with a ListPorts even with an error here ?

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+
+		portRes, err := p.bessClient.ListPorts(p.bessContext, &bess_pb.EmptyRequest{})
+		if err != nil {
+			log.Errorf("Unable to query port list for bess: %v", err)
```
safchain: same here

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+				}
+			}
+			if !edgeFound {
+				log.Infof("Deleting edge: %+v", *edge)
+				p.Ctx.Graph.DelEdge(edge)
+			}
+		}
+
+		p.Ctx.Graph.Unlock()
+		time.Sleep(1 * time.Second)
```
safchain: any way to get an event instead of looping with a sleep ?

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bess
+
+import (
+	"context"
+	"fmt"
+	log "github.com/sirupsen/logrus"
```
nplanel: please don't use your on logger, but use the probe context to log your event, like p.Ctx.Logger.Errorf("")

topology/probes/bess/bess.go
trozet: will do

topology/probes/bess/bess.go
trozet: OK I'll look into this, thanks.

topology/probes/bess/bess.go
trozet: will do

topology/probes/bess/bess.go
trozet: I think you are right we probably do need a continue to here to avoid accidentally removing nodes at line 181

lebauce:
@trozet I spent some time to investigate the issue you're facing with kiali : `../../../../pkg/mod/github.com/kiali/kiali@v0.0.0-20181002102028-f86edc2d20df/kubernetes/client.go:159:26: undefined: serializer.DirectCodecFactory` It was a pretty long journey in fact :-)
The `istio` probe makes use of Kiali to access Istio resources. The library `github.com/nimbess/nimbess-agent` you are using requires a newer version of Kubernetes libraries (Kiali relies on pretty old ones). I tried to move away from Kiali to access Istio resources. I'll push a PR soon. Could you please try it to see if it fixes your problem ?

lebauce:
@trozet Could you please try to rebase on https://github.com/skydive-project/skydive/pull/2028 ? Thanks !

trozet:
Thanks @lebauce will rebase now and address the comments

topology/probes/bess/bess.go
trozet: I don't think there is a way to do this, but I will double check

trozet:
@lebauce rebasing worked for me locally. Thanks!

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	nodeID := graph.GenID(moduleName)
+	node := p.Ctx.Graph.GetNode(nodeID)
+	if node == nil {
+		node, err = p.Ctx.Graph.NewNode(nodeID, graph.Metadata{"Name": moduleName,
+			"Type": modRes.GetMclass(), "Backend": BessBackend})
+		p.Ctx.Logger.Infof("Creating node: %s", moduleName)
+		if err != nil {
+			p.Ctx.Logger.Errorf("Error creating node: %s, %v", moduleName, err)
```
safchain: If the node creation failed I think you should return here otherwise the `AddOwnershipLink` will fail

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+				bessPort = p.Ctx.Graph.GetNode(portID)
+				if bessPort == nil {
+					bessPort, err = p.Ctx.Graph.NewNode(portID, graph.Metadata{
+						"Name":    port.GetName(),
+						"MAC":     port.GetMacAddr(),
+						"Type":    port.GetDriver(),
+						"Backend": BessBackend,
+					})
+					if err != nil {
+						p.Ctx.Logger.Errorf("Failed to create bessPort: %v", bessPort)
```
safchain: should we `continue` ? in order to avoid error later in AddOwnershipLink ?

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+	node := p.Ctx.Graph.GetNode(nodeID)
+	if node == nil {
+		node, err = p.Ctx.Graph.NewNode(nodeID, graph.Metadata{"Name": moduleName,
+			"Type": modRes.GetMclass(), "Backend": BessBackend})
+		p.Ctx.Logger.Infof("Creating node: %s", moduleName)
+		if err != nil {
+			p.Ctx.Logger.Errorf("Error creating node: %s, %v", moduleName, err)
+		}
+	}
+	if parentNode != nil && !topology.HaveOwnershipLink(p.Ctx.Graph, parentNode, node) {
```
safchain: Maybe i miss-read the code, but I would prefer to not have node not attached to any root node. Could it be the case ?

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+	p.Ctx.Logger.Info("Connecting to BESS")
+	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", p.bessHost, p.bessPort), grpc.WithInsecure())
+	if err != nil {
+		return err
+	}
+	p.Ctx.Logger.Info("Connected to BESS")
+	p.bessClient = bess_pb.NewBESSControlClient(conn)
+	verResponse, err := p.bessClient.GetVersion(context.Background(), &bess_pb.EmptyRequest{})
+	if err != nil || verResponse.Error != nil {
+		return errors.New(fmt.Sprintf("Could not get version: %s, %s", verResponse.GetError(), err))
```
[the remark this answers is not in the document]
trozet: will fix

topology/probes/bess/bess.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		p.Ctx.Logger.Errorf("Error getting module information for module: %s", moduleName)
+		return
+	}
+
+	nodeID := graph.GenID(moduleName)
+	node := p.Ctx.Graph.GetNode(nodeID)
+	if node == nil {
+		node, err = p.Ctx.Graph.NewNode(nodeID, graph.Metadata{"Name": moduleName,
+			"Type": modRes.GetMclass(), "Backend": BessBackend})
```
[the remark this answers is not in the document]
trozet: yes, will change

go.mod
```diff
@@ -69,12 +64,13 @@ require (
...
 	github.com/mattn/go-runewidth v0.0.0-20160315040712-d6bea18f7897 // indirect
 	github.com/mattn/goveralls v0.0.2
-	github.com/mitchellh/go-homedir v0.0.0-20160621174243-756f7b183b7a
+	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/hashstructure v0.0.0-20170116052023-ab25296c0f51
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826
 	github.com/networkservicemesh/networkservicemesh v0.1.0
 	github.com/newtools/ebpf v0.0.0-20190820102627-8b7eaed02eb9
+	github.com/nimbess/nimbess-agent v0.0.0-20190919205041-4e6f317ac4fd
```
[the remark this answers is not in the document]
trozet: Yeah that is the one. I will try but go modules can be tricky with dependencies.

topology/probes/bess/bess.go
trozet: agree, good catch

topology/probes/bess/bess.go
trozet: yes

topology/probes/bess/bess.go
trozet: do you mean for this comment to apply to lines 135-137? There I associate ownership to the Root to ensure that the local bess instance is owned by the local node.

In lines 156-158, the ports are being set to be owned by the local bess instance.

In these lines you are referring to (line 65), the ownership is being set for the underlying modules that make up the pipeline for the port (where the port is the parent of the pipeline of modules)

safchain:
Thanks for addressing the comments and for the PR

PR #2005, opened 2019-09-17
VPP memif

lebauce:


lebauce:
run skydive-scale-tests

lebauce:

lebauce:

lebauce:
Please review / merge

lebauce:
run skydive-scale-tests

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,508 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
 	var err error
-	node := p.Ctx.Graph.LookupFirstNode(metadata)
+	node := p.getInterface(intf.SwIfIndex)
 	if node == nil {
 		node, err = p.Ctx.Graph.NewNode(graph.GenID(), metadata)
 		if err != nil {
 			p.Ctx.Logger.Error(err)
 			return nil
 		}
+		p.NotifyEvent(graph.NodeAdded, node)
```
safchain: Does NewNode not notified already ?

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,508 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
+
 		if _, err = p.Ctx.Graph.Link(p.vppRootNode, node, topology.OwnershipMetadata()); err != nil {
 			p.Ctx.Logger.Error(err)
 			return nil
 		}
+	} else {
+		p.NotifyEvent(graph.NodeUpdated, node)
```
safchain: What is the reason to notify an event, the node doesn't seem to be updated

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,508 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
+	nsPath := fmt.Sprintf("/proc/%d/ns", pid)
+	sa, _ := common.ServiceAddressFromString(config.GetString("agent.listen"))
+	command := exec.Command(seed, "seed",
+		"--root="+string(root.ID),
+		"--pid="+path.Join(nsPath, "pid"),
+		"--mount="+path.Join(nsPath, "mnt"),
+		"--agent="+fmt.Sprintf("%s:%d", sa.Addr, sa.Port),
+		fmt.Sprintf("--filter=G.V('%s').Descendants().Subgraph()", root.ID),
+		"vpp")
+	command.Env = append(command.Env, "SKYDIVE_LOGGING_LEVEL=debug")
```
safchain: we should take the log level from the config file ?

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,508 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
-	p.createOrUpdateInterface(ch, intf)
+func (p *Probe) eventAddInterface(ch api.Channel, intf *interfaces.SwInterfaceDetails, metadata graph.Metadata) *graph.Node {
+	return p.createOrUpdateInterface(ch, intf, metadata)
 }
 
 func (p *Probe) eventDelInterface(node *graph.Node) {
-	p.Ctx.Graph.Lock()
-	defer p.Ctx.Graph.Unlock()
 	if err := p.Ctx.Graph.DelNode(node); err != nil {
+		p.NotifyEvent(graph.NodeDeleted, node)
```
safchain: same here, why sending a event ? why not subscribing to the graph ?

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,508 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
-		"Program":   string(msg.Program),
-		"Version":   string(msg.Version),
-		"BuildDate": string(msg.BuildDate),
-	}
+	p.Ctx.Logger.Debugf("Registering for VPP events")
+	p.interfaceEventsEnableDisable(ch, true)
 
-	p.Ctx.Graph.Lock()
-	defer p.Ctx.Graph.Unlock()
+	pollingTicker := time.NewTicker(VPPPollingTime)
```
safchain: this could/should be moved after line 485 ?

topology/probes/vpp/vpp.go
lebauce: Indeed. We should update the metadata here. Fixing this

topology/probes/vpp/vpp.go
lebauce: Right

topology/probes/vpp/vpp.go
lebauce: Small optimization explained in response above

topology/probes/vpp/vpp.go
lebauce: Done

topology/probes/vpp/vpp.go
lebauce: The probe embeds a graph.EventHandler so that listeners can be notified of the events generated by the probe only and not all the events on the graph.

lebauce:

lebauce:
run skydive-go-fmt

lebauce:
run skydive-go-fmt

lebauce:
run skydive-unit-tests

lebauce:
run skydive-k8s-tests

lebauce:

lebauce:
run skydive-go-fmt

lebauce:
run skydive-go-fmt

lebauce:
run skydive-go-fmt

lebauce:

lebauce:

lebauce:
run skydive-scale-tests

lebauce:
run skydive-scale-tests

lebauce:
@fgschwan We recently merged the support for seeds in Skydive : it's a single Skydive probe executed as a standalone process in the Docker namespace.
This PR modifies the VPP existing probe to be able to run inside the namespace, using the socket interface. I integrated the memifs retrieval that you implemented and the links between them. I also cherry-picked the tests from your PR.
It would be very nice if you could give it a try.

topology/probes/vpp/vpp.go
```diff
@@ -2,6 +2,7 @@
 
 //go:generate go run git.fd.io/govpp.git/cmd/binapi-generator --input-file=/usr/share/vpp/api/interface.api.json --output-dir=./bin_api
 //go:generate go run git.fd.io/govpp.git/cmd/binapi-generator --input-file=/usr/share/vpp/api/vpe.api.json --output-dir=./bin_api
+//go:generate go run git.fd.io/govpp.git/cmd/binapi-generator --input-file=/usr/share/vpp/api/memif.api.json --output-dir=./bin_api
```
milanlenco: The major issue with the design of VPP binary APIs, is that practically any change breaks backward compatibility. The binary API files generated in this PR are I suppose for 19.04 VPP. I tried to run skydive against container with VPP 19.08 and got some expected errors, such as:
```
skydive-agent_1     | 2019-10-17T10:09:29.960Z	WARN	vpp/vpp.go:649 (*Probe).Fire	vpp-93026bb9-358b-4bff-414b-6c852edd130b: time="2019-10-17T10:09:29Z" level=warning msg="Unknown message received, ID: 829"
skydive-agent_1     | 2019-10-17T10:09:27.753Z	ERROR	probes/wrapper.go:52 (*serviceManager).Start.func1	vpp-a31cbfb5-a85f-4503-5ad3-9b5225f85436: no reply received within the timeout period 1s
```
This is because any change in binary API message changes its CRC which will cause it to be refused and dropped on either VPP or GoVPP side. Similary, a newly added or removed message can shift sequence IDs of existing messages, once again breaking compatibility between client and VPP.
Although you have added the `.vppbinapi` makefile target that allows to build skydive for a custom VPP version, a single skydive binary will always be limited to only one version of VPP and might not interpret network containing multiple different containerized VPPs properly. Unfortunately, the `interface.api` that you use to dump interfaces changes pretty much with every release. For example, `SwInterfaceDump` structure has been edited (interface name is `string`, not array of bytes) in 19.08 and `probes/vpp/vpp.go` needs to be edited to reflect the change for it to compile. Actually, between `19.04` and `19.08` even the API language changed a bit (`string` type was introduced) and govpp had to be adjusted. Therefore to generate APIs for 19.08 you would also need to update the govpp dependency first. To make this even worse, in 19.08 the path to installed `*.api.json` files has changed - core APIs and plugin APIs are further separated to `core` and `plugins` sub-directories.

What we do in [ligato/vpp-agent](https://github.com/ligato/vpp-agent) to mitigate this issue, is that we hide interaction with VPP behind an interface. Then we provide implementation of that interface for the latest VPP version (top of the master branch, frequently updated) and for the last 2 releases (for example: https://github.com/ligato/vpp-agent/tree/master/plugins/vpp/ifplugin/vppcalls). Each of these implementations has its own generated binary APIs corresponding to its version. For example [in this file](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/ifplugin/vppcalls/if_vppcalls.go) we define APIs for managing VPP interfaces. Then every implementation of this API registers itself into [this map](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/ifplugin/vppcalls/if_vppcalls.go#L238). When connection to a VPP instance is made, [CheckCompatibility method is used](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/ifplugin/vppcalls/if_vppcalls.go#L248) to find and select that implementation which is compatible with the given VPP. In your case you could define API with methods like `FetchMemifs()`, `FetchInterfaces()`, `SubscribeInterfaceEvents` and define your own structures for interface metadata with fields that you actually use.

If support of multiple VPP versions is not a priority right now, then maybe at least document which version is supported out-of-the-box right now and perhaps allow to switch the `.vppbinapi` makefile target off, so that skydive can be compiled with binary APIs already generated and included in the repo, without having to have VPP installed in the host.

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,511 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
-	if err != nil {
+	if err := ch.SendRequest(req).ReceiveReply(msg); err != nil {
 		p.Ctx.Logger.Error(err)
 	}
 }
 
-func (p *Probe) interfacesEvents() {
-	defer p.wg.Done()
+type memifSocket struct {
+	*memif.MemifDetails
```
milanlenco: This is an example where interface configuration is defined be directly re-using structures from the VPP binary API. Since that changes quite often, it might not compile if skydive is build against a different VPP version.

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,511 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
+	msg2 := &memif.MemifDetails{}
+	reqCtx = ch.SendMultiRequest(req2)
+	for {
+		if lastReply, err := reqCtx.ReceiveReply(msg2); lastReply {
+			break
+		} else if err != nil {
+			return nil, err
+		}
+
+		sharedMemifSockets[msg2.SocketID].MemifDetails = msg2
```
milanlenco: Actually there can be multiple (listener) memifs (on single VPP) that use the same socket. Master-Slave memif connection is created not only when the same socket file is used, but also when IDs (field `ID`) are the same.
In other words there can be multiple listeners with different IDs using the same socket and based on the ID of the connecting memif the match is made.

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,511 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
-		msg := notif.(*interfaces.SwInterfaceEvent)
+	}
+}
 
-		node := p.getInterface(msg.SwIfIndex)
-		p.Ctx.Graph.RLock()
+func (p *Probe) GetABLinks(node *graph.Node) (edges []*graph.Edge) {
+	socketFilename, _ := node.GetFieldString("VPP.SocketFilename")
+	socketFilename, _ = p.getHostSocketFilename(socketFilename, node)
+	nodes, _ := p.slaveMemifs.Get(socketFilename)
```
milanlenco: Use also IDs to determine the links between memifs.

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,511 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
+	sub.Unsubscribe()
 	ch.Close()
 }
 
-// Start VPP probe and get all interfaces
-func (p *Probe) Start() error {
-	conn, err := govpp.Connect(p.shm)
+// Do starts the VPP probe and get all interfaces
+func (p *Probe) Do(ctx context.Context, wg *sync.WaitGroup) error {
+	conn, err := govpp.Connect(p.addr)
```
milanlenco: I'm getting lot's of these error messages for the host and containers that do not run any VPP:
```
skydive-agent_1     | 2019-10-17T12:23:27.844Z	ERROR	probes/wrapper.go:52 (*serviceManager).Start.func1	dev-VirtualBox: VPP connection error: VPP API socket file /run/vpp/api.sock does not exist
skydive-agent_1     | 2019-10-17T12:23:28.016Z	ERROR	probes/wrapper.go:52 (*serviceManager).Start.func1	vpp-9d93ad24-9cc6-4d85-584e-c333eb1777ae: VPP connection error: VPP API socket file /run/vpp/api.sock does not exist
```

Perhaps `Do()` method could start by checking if the socket files exists or if a process named `vpp` exists and continue only if it is the case to avoid polluting the log.

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,511 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
+
+		sharedMemifSockets[msg.SocketID] = &memifSocket{filename: interfaceName(msg.SocketFilename)}
 	}
 
-	sub, err := ch.SubscribeNotification(p.notifChan, &interfaces.SwInterfaceEvent{})
+	req2 := &memif.MemifDump{}
+	msg2 := &memif.MemifDetails{}
+	reqCtx = ch.SendMultiRequest(req2)
+	for {
+		if lastReply, err := reqCtx.ReceiveReply(msg2); lastReply {
```
milanlenco: there is a bug: `msg2 := &memif.MemifDetails{}` should be inside the for loop just before `ReceiveReply()`. Otherwise it is the same struct instance that is being rewritten with each iteration and all the memifs will end up in `sharedMemifSockets` with the same configuration of the last memif. (i.e. this issue occurs if VPP has multiple memifs)

topology/probes/vpp/vpp.go
```diff
@@ -103,268 +124,511 @@ func (p *Probe) getInterfaceVrfID(ch api.Channel, index uint32) int64 {
...
+					SocketID:       int64(socket),
+					Master:         memifSock.Role == 0,
+					Mode:           memifModes[memifSock.Mode],
+					RingSize:       int64(memifSock.RingSize),
+					BufferSize:     int64(memifSock.BufferSize),
+					LinkUpDown:     memifSock.LinkUpDown == 1,
+				}
+			}
+		}
+
```
milanlenco: Btw, the original Filip's PR also contained a support for AF-PACKET interfaces. If it is of an interest for you, [this is how you can tell if the given interface is AF_PACKET](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go#L1010) and [this is how to obtain the name of the Linux interface to which it is attached](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go#L149).

milanlenco:
I was able to get it running and overall it looks good, only added few comments.

Also one general comment: If it is not already done/planned, I would also welcome if it was documented how to build and use vpp probe. For example mention that `WITH_VPP=true` has to be defined, that `vpp` probe is not loaded by default and needs to be listed in `agent.topology.probes` config and also document which version of VPP is supported out-of-the-box (for which version the binary API files have been generated and included in the repo).

lebauce:
@milanlenco Thanks a lot for taking the time to test and for the review. This is really useful, I will try to address the issues

PR #2011, opened 2019-09-18
exporters: Test gremlin resolvers using local topology files

dubek:
This is an attempt to increase unit test coverage of the exporters functionality.

This PR adds tests for the two name resolvers: `resolveDocker` and `resolveRunc`. The tests use implement a local gremlin instance (see `resolvers_test.go`), whose topology graph is populated from a local JSON file. The resolver's queries are executed against this local instance (instead of being sent as HTTP request in the real resolver case).

In order not to commit large hard-to-understand topology files, I removed unnecessary info from the JSON topology files so that they include only the relevant nodes and links.

Following @hunchback 's advice, after we gain some experience with this local gremlin instance for testing here, it might be useful to extract it to the a Skydive-wide package (perhaps under `gremlin`) and then be used for any test that runs gremlin queries.

### DIFF
--- a/contrib/exporters/secadvisor/mod/resolve_docker.go
+++ b/contrib/exporters/secadvisor/mod/resolve_docker.go
@@ -34,7 +34,7 @@ func NewResolveDocker(cfg *viper.Viper) Resolver {
 }
 
 type resolveDocker struct {
-	gremlinClient *client.GremlinQueryHelper
+	gremlinClient GremlinNodeGetter
 }
 
 // IPToName resolve IP address to Docker container name.
--- a/contrib/exporters/secadvisor/mod/resolve_docker_test.go
+++ b/contrib/exporters/secadvisor/mod/resolve_docker_test.go
@@ -1,0 +1,62 @@
+package mod
+
+import (
+	"testing"
+)
+
+func newTestResolveDocker(t *testing.T) Resolver {
+	return &resolveDocker{gremlinClient: newLocalGremlinQueryHelper(newDockerTopologyGraph(t))}
+}
+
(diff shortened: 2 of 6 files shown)


dubek:
I also added a test for the whole secadvisor Transform logic (the entire resolvers chain), using a local gremlin instance running queries over a graph loaded from a local JSON file.

dubek:
I force-pushed ce0a99ca80a6aaed38b16e94858db4cf90a31cad which includes `go fmt` fix and also better checking of validity of local graph files.

dubek:
... and pushed again 5f3859fdd3f6fc17a469ea34df115ceb2575d282 - rebased on top of current master.

dubek:
The `skydive-compile-tests` fail on some Windows-related error in `prometheus/client_golang`:

```
# Compile Skydive for Windows
GO111MODULE=on GOOS=windows GOARCH=amd64 go build github.com/skydive-project/skydive
# github.com/prometheus/client_golang/prometheus
../../../../pkg/mod/github.com/prometheus/client_golang@v1.1.0/prometheus/process_collector_windows.go:78:9: assignment mismatch: 2 variables but 1 values
Build step 'Execute shell' marked build as failure
```

Any idea?

hunchback:
i have only one concern - that the json representation of the graph is very frigle -  maybe better if you can construct the graph programmatically.

lebauce:
@dubek You can try to rebase on master

dubek:
> i have only one concern - that the json representation of the graph is very frigle - maybe better if you can construct the graph programmatically.

@hunchback  - Done - see new functions `newRuncTopologyGraph` and `newDockerTopologyGraph` (replacing the JSON files). I hope this is better.

----

> @dubek You can try to rebase on master

Done.

PR #2015, opened 2019-09-19
common: Add unit test for GetMapField

dubek:
In my attempts to understand how gremlin query like `G.V().Has('a.b.c.d', 'value')` accesses the value of `a.b.c.d`, and especially what happens if the node's metadata contains arrays (Kubernetes pod metadata, I'm looking at you), I ended up writing a unit test for `common.GetMapField` to validate my assumptions.

Ideally eventually we would add some documentation to http://skydive.network/documentation/api-gremlin , which has some examples of lookup of keys inside a nested map (like `G.V().Has('Docker.Labels.io.kubernetes.pod.namespace' ,'my-namespace').Descendants()`) but doesn't mention what happens if the metadata contains arrays.

### DIFF
--- a/common/types_test.go
+++ b/common/types_test.go
@@ -201,6 +201,63 @@ func TestDelField(t *testing.T) {
 	}
 }
 
+func TestGetMapField(t *testing.T) {
+	data := map[string]interface{}{
+		"a": map[string]interface{}{
+			"b": map[string]interface{}{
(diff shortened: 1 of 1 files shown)


safchain:
Can you rebase on top of master please ?

dubek:
Rebased.

dubek:

PR #2016, opened 2019-09-19
graph/traversal: Fix Sum() unit test

dubek:
Before this fix the tests printed the following log message (while passing):

    === RUN   TestBasicTraversal
    --- PASS: TestBasicTraversal (0.00s)
        traversal_test.go:204: Error in Sum() step: %!s(<nil>)

The fix correctly casts the result to int64 and fails if the `Sum()` function returns the wrong value.


safchain:
run skydive-scale-tests

PR #2027, opened 2019-09-26
analyzer: Allow setting startup capture type

dubek:
When using a startup capture, the config allowed setting the capture's
gremlin (which nodes should capture) and BPF filter string, but the
capture type was hard-coded to "pcap".

This change allows setting the type from the configuration: either
setting the `analyzer.startup.capture_type` key in the YAML config file or
by setting the `SKYDIVE_ANALYZER_STARTUP_CAPTURE_TYPE` environment
variable.

Possible values for `capture_type` are those that appear in `ProbeTypes`
(see `common/capture.go`); they are listed in a comment in the example
config file.

The default value for this config is `""`, which lets Skydive choose the capture type.

@eranra or @hunchback : please review.

### DIFF
--- a/etc/skydive.yml.default
+++ b/etc/skydive.yml.default
@@ -69,6 +69,11 @@ analyzer:
     # capture_gremlin: "G.V().has('Name', NE('lo'))"
     # capture_bpf: "port 80"
 
+    # By default (capture_type: "") the capture type is chosen automatically;
+    # or set here to one of pcap, afpacket, ebpf, sflow, pcapsocket, ovsmirror,
+    # dpdk, ovssflow, or ovsnetflow.
+    # capture_type: ""
+
   # Flow storage engine
   flow:
     # Storage backend name: myelasticsearch, myorientdb
--- a/analyzer/server.go
+++ b/analyzer/server.go
@@ -123,9 +123,10 @@ func (s *Server) createStartupCapture(ch *api.CaptureAPIHandler) error {
 	}
 
 	bpf := config.GetString("analyzer.startup.capture_bpf")
-	logging.GetLogger().Infof("Invoke capturing from the startup with gremlin: %s and BPF: %s", gremlin, bpf)
+	captureType := config.GetString("analyzer.startup.capture_type")
+	logging.GetLogger().Infof("Invoke capturing of type '%s' from startup with gremlin: %s and BPF: %s", captureType, gremlin, bpf)
 	capture := types.NewCapture(gremlin, bpf)
-	capture.Type = "pcap"
+	capture.Type = captureType
 	return ch.Create(capture, nil)
 }
 


etc/skydive.yml.default
```diff
@@ -68,6 +68,8 @@ analyzer:
     # from the beginning automatically
     # capture_gremlin: "G.V().has('Name', NE('lo'))"
     # capture_bpf: "port 80"
+    # Capture type: pcap [default], afpacket, ebpf, sflow, pcapsocket, ovsmirror, dpdk, ovssflow, or ovsnetflow
```
lebauce: nit: The colon makes the sentence look a bit like valid YAML. Can you please change it to :
`Specify the capture type. Can be either pcap (the default), afpacket, ebpf, sflow, pcapsocket, ovsmirror, dpdk, ovssflow, or ovsnetflow`

dubek: I'll fix the comment sentence.

Regrading the default: for *startup* captures, `analyzer/server.go` line 128 specifically hard-coded:

    capture.Type = "pcap"

Or maybe this is overridden somewhere else?

lebauce: It seems you did set the default value in config/config.go in the first version of your PR, or I missed it.

I would prefer to remove the line in config/config.go and let Skydive decide the type of the capture (so that it works on Openvswitch bridges for instance)

dubek: I did set the default in `config/config.go` in order to keep the behavior before the change, which hard-coded startup captures to be of type `pcap`.

I'll modify it so the default is an empty string, which will let Skydive decide the type of the capture per node.

Note that this will break existing behavior of startup captures which were always of type `pcap`; now such users will have to add this config to keep their existing behavior:

```
analyzer:
  startup:
    capture_type: pcap
```

(or the corresponding environment variable.)

dubek: Fixed in 83b1d8c05114581a98550c37ec8bccd28decb0f7 (force-pushed).

config/config.go
```diff
@@ -135,6 +135,7 @@ func init() {
 	cfg.SetDefault("analyzer.topology.k8s.config_file", "/etc/skydive/kubeconfig")
 	cfg.SetDefault("analyzer.topology.ovn.address", "unix:///var/run/openvswitch/ovnnb_db.sock")
 	cfg.SetDefault("analyzer.topology.istio.config_file", "/etc/skydive/kubeconfig")
+	cfg.SetDefault("analyzer.startup.capture_type", "")
```
lebauce: You can simply remove the line

dubek: Removed in 1222b9696ed92681b5a29536e4029cd42bd96694 .

lebauce:
run skydive-scale-tests

lebauce:
run skydive-scale-tests

lebauce:
run skydive-scale-tests

lebauce:
run skydive-scale-tests

etc/skydive.yml.default
```diff
@@ -69,6 +69,11 @@ analyzer:
     # capture_gremlin: "G.V().has('Name', NE('lo'))"
     # capture_bpf: "port 80"
 
+    # By default (capture_type: "") the capture type is chosen automatically;
+    # or set here to one of pcap, afpacket, ebpf, sflow, pcapsocket, ovsmirror,
+    # dpdk, ovssflow, or ovsnetflow.
+    # capture_type: ""
```
nplanel: if capture_type is set to ""  bpf_filter/probe capabilities will not be checked 
please set to "pcap" as default and in config/config.go as well

nplanel: please not that not all capture type support BPF filtering

dubek: Ok that's mote complicated than I thought.

Originally I had a "pcap" as default, but @safchain suggested leaving blank as default to allow skydive to choose the best capture type for each node.

My original problem was that "pcap" was hard coded for startup captures and I want to try ebpf. 

Would be glad to hear suggestions on how to allow this without confusing the user. I guess similar problems should appear in the skydive client when starting a capture.

lebauce: @nplanel It's true but this is the case for all the captures we create, not only the startup one. Setting the type as `pcap` also bring problem : if you select OVS bridges and leave `capture_type` empty, it will not work. So I think this PR has value and already fixes a problem

PR #2034, opened 2019-09-30
tests: fix scale test compilation

lebauce:


lebauce:
run skydive-scale-tests

PR #2048, opened 2019-10-04
Report drop n stats

safchain:


flow/server/flow_server.go
```diff
@@ -233,16 +233,16 @@ func (s *FlowServer) handleFlows(flows []*flow.Flow) {
 	}
 }
 
-func (s *FlowServer) handleStatus(status *flow.Status) {
-	s.subscriberEndpoint.SendStatus(status)
+func (s *FlowServer) handleStatus(stats *flow.Stats) {
+	s.subscriberEndpoint.SendStatus(stats)
```
dubek: should be `handleStats` and `SendStats`

flow/table.go
```diff
@@ -582,14 +584,18 @@ func (ft *Table) Run() {
...
+			ft.stats.KernelFlowDropped += s.KernelFlowDropped
+			ft.stats.PacketsReceived += s.PacketsReceived
+			ft.stats.PacketsDropped += s.PacketsDropped
+		case <-statsTicker.C:
+			ft.stats.CaptureID = ft.uuids.CaptureID
+			ft.stats.FlowCount = int64(ft.table.Len())
+
+			ft.sender.SendStats(ft.stats)
+
+			logging.GetLogger().Warningf("Flow table stats: %+v", ft.stats)
```
dubek: Why is this a warning? Should be info.

flow/table.go
```diff
@@ -582,14 +584,18 @@ func (ft *Table) Run() {
...
-			ft.status.FlowCount = int64(ft.table.Len())
-
-			ft.sender.SendStatus(ft.status)
-
-			logging.GetLogger().Warningf("Flow table status: %+v", ft.status)
-			ft.status = Status{}
+		case s := <-ft.statsChan:
+			ft.stats.KernelFlowDropped += s.KernelFlowDropped
+			ft.stats.PacketsReceived += s.PacketsReceived
+			ft.stats.PacketsDropped += s.PacketsDropped
```
dubek: This looks like a candidate for an extracted function:

    func (s *Stats) merge(other *Stats)

Can we add methods on `Stats` even when it's generated from a proto file?

flow/probes/ebpf.go
```diff
@@ -132,14 +132,14 @@ func (p *EBPFProbe) run() {
...
 			if statsMap := p.module.Maps["stats_map"]; statsMap != nil {
-				var statsKey uint32
-				var statsVal int64
+				var dropKey uint32
+				var dropValue int64
 
-				if found, err := statsMap.Get(statsKey, &statsVal); err == nil && found {
-					if statsVal > 0 {
-						p.Ctx.Logger.Warningf("flow table overflow, %d flows were dropped from kernel table", statsVal)
+				if found, err := statsMap.Get(dropKey, &dropValue); err == nil && found {
```
hunchback: does dropKey not need to be initialized?

flow/table.go
hunchback: u probably want to delete this log message entirly

flow/probes/ebpf.go
safchain: it is, golang initializes it to 0

flow/server/flow_server.go
safchain: indeed, I renamed it, forgot to rename this one

flow/table.go
safchain: just to keep to previous behavior, I'll add it as debug

flow/table.go
safchain: Only used here, no real point to extract for now and the function won't be generic as only few metrics are "merged"

safchain:
run tests

safchain:
run skydive-scale-tests

PR #2050, opened 2019-10-06
ci/create-docker-image.sh: Remove unused BASE build-arg

dubek:
Both Docker.compile and Docker.crosscompile don't use the `BASE`
build-arg at all.  The output log of the CI ([example](http://ci-logs.skydive.network/skydive-create-docker-image/builds/3/log)) shows this:

    [Warning] One or more build-args [BASE] were not consumed

This patch removes this unused build-arg from the
`skydive-compile-build` image building (it is still used in the skydive
or snapshot images to allow setting `BASE` per architecture).


hunchback:
run skydive-scale-tests

PR #2052, opened 2019-10-07
ebpf is limited, try to avoid passing parameters

nplanel:


lebauce:
I'm not sure to remember which issue was fixed by this one :-) Would it be possible to close it ?

nplanel:
yeah was a attempt to fix the tunneling

PR #2057, opened 2019-10-09
topology: add block device topology probe

trgill:
The probe reads JSON from the lsblk command to get a
description of the block devices on the system.  The
JSON is parsed into structures that describe the relationships
between the devices (parent/child,LVM,MD,multipath).

If the environment variable SKYDIVE_BLOCKDEV_TEST_FILE is set
the file is read rather than invoking "lsblk --json".  This is
for testing more complex block device setups that might not
be available on the local system.

Signed-off-by: Todd Gill <tgill@redhat.com>


trgill:
This is a WIP - I opened the PR to give people an early look to see if it is headed in the right direction.

To test the probe - set SKYDIVE_BLOCKDEV_TEST_FILE to a file containing the following JSON:

https://pastebin.com/bJfcFxin

The JSON has the contents of "lsblk" from a system with several types of block storage.

topology/probes/blockdev/metadata.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/skydive-project/skydive/common"
+	"github.com/skydive-project/skydive/graffiti/graph"
+)
+
+// Metadata describe the metadata of a docker container
```
hunchback: s/docker container/block device/

agent/topology_probes.go
```diff
@@ -47,6 +48,8 @@ import (
 // NewTopologyProbe creates a new topology probe
 func NewTopologyProbe(name string, ctx tp.Context, bundle *probe.Bundle) (probe.Handler, error) {
 	switch name {
+	case "blockdev":
+		return new(blockdev.ProbeHandler).Init(ctx, bundle)
```
lebauce: We recently merged a PR so that Skydive handles probes as external .so
Therefore, we now try to use the `module.NewProbe(ctx, bundle)` scheme. Could you please do this small refactoring ?

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,462 @@
...
+	return nil
+}
+
+// Start the probe
+func (p *ProbeHandler) Start() error {
+
+	p.addGroupByName(blockdevGroupName, "")
+
+	if _, err := topology.AddLink(p.Ctx.Graph, p.Ctx.RootNode, p.Groups[blockdevGroupName], "connected", nil); err != nil {
+		p.Ctx.Logger.Error(err)
```
lebauce: Should we return err ?

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,462 @@
...
+
+	for i := range result.Blockdevices {
+		p.registerBlockdev(result.Blockdevices[i], "")
+	}
+
+	return nil
+}
+
+// Start the probe
+func (p *ProbeHandler) Start() error {
```
lebauce: You could make NewProbe (renamed from Init) return a probe wrapper as in https://github.com/skydive-project/skydive/blob/master/topology/probes/docker/docker.go#L281
Therefore, you don't need to implement Start and Stop, but only the Do method. That would give something like :

```
 func (p *ProbeHandler) Do(ctx context.Context, wg *sync.WaitGroup) error {
    p.addGroupByName(blockdevGroupName, "")
    ...

    wg.Add(1)
    go func() {
        defer wg.Done()

        after := time.After(30*time.Second)
        for {
            if err := p.connect(); err != nil {
                p.Ctx.Logger.Error(err)
                return
            }

            select {
                case <- ctx.Done():
                    return
                case <-after:
            }
        }
    }
```

topology/probes/blockdev/metadata.go
```diff
@@ -0,0 +1,98 @@
...
+	Type         string
+	UUID         string
+	Vendor       string
+	Wsame        string
+	WWN          string
+	Labels       graph.Metadata `field:"Metadata"`
+}
+
+// MetadataDecoder implements a json message raw decoder
+func MetadataDecoder(raw json.RawMessage) (common.Getter, error) {
```
lebauce: You need to register your MetadataDecoder to the analyzer in the `init` function of analyzer/server.go

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	metadata = graph.Metadata{
+		"Path":       p.getPath(blockdev),
+		"Type":       nodeType,
+		"Name":       p.getName(blockdev),
+		"MajorMinor": blockdev.MajMin,
+		"WWN":        WWN,
+		"Manager":    managerType,
+		"BlockID":    p.getID(blockdev),
+		"MetaData":   blockdevMetadata,
```
lebauce: We are already in metadata, so a MetaData field is a bit confusing.
What about `BlockDevice` or `BlockDev` ? By the way, you should use the same as in the `init` function of analyzer/server.go

I would also suggest to move "BlockID", "WWN" and "MajorMinor" to the Metadata struct you added

topology/probes/blockdev/metadata.go
```diff
@@ -0,0 +1,98 @@
+//go:generate go run github.com/skydive-project/skydive/scripts/gendecoder -package github.com/skydive-project/skydive/topology/probes/blockdev
```
lebauce: Please add the metadata_easyjson.go and metadata_gendecoder.go files to the repository

topology/probes/blockdev/metadata.go
```diff
@@ -0,0 +1,98 @@
...
+//go:generate go run github.com/safchain/easyjson/easyjson $GOFILE
```
lebauce: go run github.com/mailru/easyjson/easyjson

topology/probes/blockdev/metadata.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/skydive-project/skydive/graffiti/graph"
+)
+
+// Metadata describe the metadata of a docker container
+// easyjson:json
+// gendecoder
+type Metadata struct {
+	ID           string
+	Name         string
+	Alignment    int
```
lebauce: We usually try to use int64 instead of int in Metadata

topology/probes/blockdev/metadata.go
```diff
@@ -0,0 +1,98 @@
...
+	Wsame        string
+	WWN          string
+	Labels       graph.Metadata `field:"Metadata"`
+}
+
+// MetadataDecoder implements a json message raw decoder
+func MetadataDecoder(raw json.RawMessage) (common.Getter, error) {
+	var m Metadata
+	if err := json.Unmarshal(raw, &m); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal docker metadata %s: %s", string(raw), err)
```
lebauce: s/docker/blockdev/

topology/probes/blockdev/no_blockdev.go
```diff
@@ -0,0 +1,43 @@
...
+
+package blockdev
+
+import (
+	"github.com/skydive-project/skydive/common"
+	"github.com/skydive-project/skydive/probe"
+	tp "github.com/skydive-project/skydive/topology/probes"
+)
+
+// ProbeHandler describes a Docker topology graph that enhance the graph
```
lebauce: s/Docker/Block device/

lebauce:
run tests

lebauce:
run tests

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,462 @@
...
+	Partflags    string `json:"partflags"`
+	Partlabel    string `json:"partlabel"`
+	Parttype     string `json:"parttype"`
+	Partuuid     string `json:"partuuid"`
+	Path         string `json:"path"`
+	PhySec       int    `json:"hpy-sec"`
+	Pkname       string `json:"pkname"`
+	Pttype       string `json:"pttype"`
+	Ptuuid       string `json:"ptuuid"`
+	Ra           int    `json:"ra"`
```
safchain: It seems that depending of the lsblk version the output types are not the same :( So you would have to probably maintain mutliple struct. With my version I get a string type for the Ra field
lsblk from util-linux 2.32.1

lebauce: Or just comment `Ra` as it's not used

lebauce: An other option would be to json.Unmarshal to get an interface{}
Then use mapstructure to map it to a BlockDevice struct (we already use mapstructure in a couple of places)

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,462 @@
...
+		}
+	} else {
+		if cmdOut, err = exec.Command("lsblk", "-pO", "--json").Output(); err != nil {
+			p.Ctx.Logger.Error(err)
+			os.Exit(1)
+		}
+	}
+
+	jsonBlockdevs := string(cmdOut[:])
+	if err = json.Unmarshal([]byte(jsonBlockdevs), &result); err != nil {
```
safchain: json.Unmarshal(cmdOut, ....
is not working ? instead of passing via a string

safchain:
Hi Todd, thanks for submitting this PR. Just to let you know that we are actively working on a new WebUI. Here how your PR looks like for now. There is a way to configure some rendering things but I did not do anything for now.

screenshot-0 0 0 0_8080-2019 10 10-14_17_00

trgill:
@safchain the new UI looks great!

The JSON was generated on a system that has a mirrored LVM logical volume that uses 2 multipath devices (4 paths each).  It might be a good test case for the new UI.  This is what it looks like currently:
skydive-lvm-over-multipath

trgill:
@hunchback @lebauce @safchain thanks for the feedback.  I'll make the suggested changes.

The icons are intended as just place holders for something better.  I think it might be useful to visually distinguish physical block devices, DM, LVM, MD, and multipath etc.

I started looking at using "iostat" to map in performance data to the physical devices.   This is a JSON example of the multipath devices under and IO load:

https://pastebin.com/NT1YpBE5

Do you have any example probes that I should follow to map in the performance data?

topology/probes/blockdev/metadata.go
trgill: thanks.

topology/probes/blockdev/blockdev.go
trgill: yes.  will fix

topology/probes/blockdev/blockdev.go
trgill: Ok - thanks.

topology/probes/blockdev/metadata.go
trgill: ok

topology/probes/blockdev/blockdev.go
trgill: ok

topology/probes/blockdev/metadata.go
trgill: ok

topology/probes/blockdev/blockdev.go
trgill: It is the "Read Ahead" for the device.  I was testing with 2.33.2 and it is an integer in that version.  I'll look at this more.

lebauce:
run skydive-scale-tests

topology/probes/blockdev/blockdev.go
lebauce: if err = json.Unmarshal([]byte(cmdOut[:]), &result); err != nil {

statics/js/components/topology.js
```diff
@@ -1031,8 +1031,8 @@ Node.prototype = {
   },
 
   isCaptureAllowed: function() {
-    var allowedTypes = ["device", "veth", "ovsbridge", "geneve", "vlan", "bond", "ovsport",
-                        "internal", "tun", "bridge", "vxlan", "gre", "gretap", "dpdkport"];
+    var allowedTypes = ["device", "veth", "blockdev", "blockdevleaf", "blockdevlvm", "ovsbridge", "geneve",
```
lebauce: I'm not sure you need this

topology/probes/blockdev/blockdev.go
trgill: Ok.

topology/probes/blockdev/blockdev.go
trgill: It looks like lsblk version 2.33 "has been improved to differentiate between numbers,
booleans and strings in JSON output".  See:

https://mirrors.edge.kernel.org/pub/linux/utils/util-linux/v2.33/v2.33-ReleaseNotes

I added some code to deal with integers and boolean types possibly encoded as strings.  But, I think it introduces a problem with how skydive looks at the meta data - I'm not sure what the issue is, but currently the blockdev meta data doesn't expand in the UI.  (?)

I'll see if I can make it work.

statics/js/components/topology.js
trgill: looks like things work fine without - so I removed from PR

tests/blockdev_test.go
```diff
@@ -0,0 +1,324 @@
...
+func setupBlockDev(c *TestContext) error {
+	if err := ioutil.WriteFile(testFileName, jsondata, 0644); err != nil {
+		return err
+	}
+
+	if err := Setenv("SKYDIVE_BLOCKDEV_TEST_FILE", testFileName); err != nil {
+		return err
+	}
+
+	// TODO restart the blockdev probe?
```
trgill: Is there a way to restart the probe here?  Or should I take a different approach?

tests/tests.go
```diff
@@ -80,7 +80,8 @@ agent:
   listen: {{.AgentAddr}}:{{.AgentPort}}
   topology:
     probes:
-    - ovsdb
+	- ovsdb
```
lebauce: Please fix indentation (spaces instead of tabs)

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+
+func (p *ProbeHandler) registerBlockdev(blockdev BlockDevice, parentWWN string) *graph.Node {
+	var groupNode *graph.Node
+	childCount := len(blockdev.Children)
+	if childCount > 0 {
+		for i := range blockdev.Children {
+			groupNode = p.registerBlockdev(blockdev.Children[i], blockdev.WWN)
+		}
+		if blockdev.Type == multipathType {
+			groupNode = p.addGroup(blockdev, parentWWN)
```
lebauce: addGroup seems to require the lock on ProbeHandler (line 178). For simplicity, you could do
```
p.Lock()
defer p.Unlock()
```
at the start of the function, like in `unregisterBlockdev`

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+}
+
+func (p *ProbeHandler) getInt(field json.RawMessage) int64 {
+	// If the JSON has an integer value of "field": null field parameter
+	// to this function will be set to nil
+	if field == nil {
+		return 0
+	}
+	if utf8.Valid(field) {
+		var i int64
```
lebauce: nit:
```
if i, err := strconv.ParseInt(strings.Replace(string(field), "\"", "", -1), 10, 64); err == nil {
    return i
}
```

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+		if err == nil {
+			return i
+		}
+	}
+	p.Ctx.Logger.Error(errors.New("blockdev probe failed to parse integer JSON field from lsblk"))
+	return -1
+}
+
+func (p *ProbeHandler) getBool(field json.RawMessage) bool {
+
```
lebauce: nit: remove empty line

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+// NewProbe initializes a new topology blockdev probe
+func NewProbe(ctx tp.Context, bundle *probe.Bundle) (probe.Handler, error) {
+
+	p := &ProbeHandler{
+		blockdevMap: make(map[string]blockdevInfo),
+		Groups:      make(map[string]*graph.Node),
+		Ctx:         ctx,
+	}
+
+	p.Ctx = ctx
```
lebauce: this line can be removed

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+	p.blockdevMap = make(map[string]blockdevInfo)
```
lebauce: what about putting this in the `p` initialization above :
```
    blockdevMap: make(map[string]blockdevInfo),
    Groups: make(map[string]*graph.Node),
```

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+
+// BlockDevice used for JSON parsing
+// Prior to version 2.33 lsblk generated JSON that encodes all values
+// as strings.  Version 2.33 changed how integers and booleans are
+// encoded.  Using json.RawMessage allows the code deal with either
+// encoding.
+type BlockDevice struct {
+	Children []BlockDevice `json:"children"`
+
+	Alignment    json.RawMessage `json:"alignment"` //int64
```
lebauce: An alternative would be :
- use json.Unmarshal to a map[string]interface{}
- then use https://godoc.org/github.com/mitchellh/mapstructure#WeakDecode to map to a `BlockDevice` struct. It converts strings to ints (and more) so it could handle a change of type of a member.

It's really up to you :-) We can merge it without it

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+// ProbeHandler describes a block device graph that enhances the graph
+type ProbeHandler struct {
+	common.RWMutex
+	Ctx         tp.Context
+	blockdevMap map[string]blockdevInfo
+	wg          sync.WaitGroup
+	Groups      map[string]*graph.Node
+}
+
+func (p *ProbeHandler) getPath(blockdev BlockDevice) string {
```
lebauce: nit: what about making `getPath` a method on BlockDevice instead of ProbeHandler

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+	if blockdev.Path != "" {
+		return blockdev.Path
+	}
+	if blockdev.Name != "" {
+		return blockdev.Name
+	}
+	return blockdev.Kname
+}
+
+func (p *ProbeHandler) getID(blockdev BlockDevice) string {
```
lebauce: same than above

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+	if blockdev.WWN != "" {
+		return blockdev.WWN
+	}
+	if blockdev.Serial != "" {
+		return blockdev.Serial
+	}
+	return blockdev.Name
+}
+
+func (p *ProbeHandler) getName(blockdev BlockDevice) string {
```
lebauce: same than above

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+		cmdOut []byte
+		err    error
+		result Devices
+	)
+
+	testFile := os.Getenv("SKYDIVE_BLOCKDEV_TEST_FILE")
+
+	if testFile != "" {
+		if cmdOut, err = exec.Command("cat", testFile).Output(); err != nil {
+			p.Ctx.Logger.Error(err)
```
lebauce: No need to log (it will be done by the called), just `return err`

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+			os.Exit(1)
+		}
+	} else {
+		if cmdOut, err = exec.Command("lsblk", "-pO", "--json").Output(); err != nil {
+			p.Ctx.Logger.Error(err)
```
lebauce: Same then above

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,515 @@
...
+			os.Exit(1)
+		}
+	}
+
+	if err = json.Unmarshal([]byte(cmdOut[:]), &result); err != nil {
+		p.Ctx.Logger.Error(err)
```
lebauce: Same than above

tests/tests.go
trgill: oops.  will fix.

lebauce:
To test, I see 2 solutions :
- execute real commands (using losetup ?) but, as I'm not a storage expert, I don't know if this is feasible
- add a configuration knob to be able to specify the location of the `lsblk` executable, so that we can create a fake `lsblk` that returns the JSON output

I created a branch based on yours with 2 additional commits:
- one (https://github.com/lebauce/skydive/commit/a9e2d52d6bcc314d50aa3859942e7fec03f6a848) that implements the second solution and improves testing a bit
- an other (https://github.com/lebauce/skydive/commit/8500b5f7e93e5864bb46e8abfed0870cf2f8a277) that makes use of mapstructure (this has to be checked though)

Feel free to cherry pick or ignore any of them :-)

topology/probes/blockdev/blockdev.go
trgill: ok

topology/probes/blockdev/blockdev.go
trgill: thanks - that's better

topology/probes/blockdev/blockdev.go
trgill: I think I've implemented it the way you've suggested.  It looks better to me - please have a look and let me know if I got it right. :-)  Will push soon.

topology/probes/blockdev/blockdev.go
trgill: I agree - but the functions are no longer needed when I the JSON parsing technique

topology/probes/blockdev/blockdev.go
trgill: Ok - I see your 2nd commit looks better than what I came up with... I'll cherry-pick and test.

trgill: @lebauce I squashed your JSON parser changes (hope that is ok) and cherry-picked the "lsblk" path changes.

where are you thinking we should put the fake lsblk?

trgill:
@lebauce loopback devices would be an option - but I've had trouble with them when test frameworks fail.  The tear down can be tricky when they are used as storage for higher levels. 

I'm thinking if we could add a few .json files from complex storage environments - we could avoid having to deal with actually using DM/MD/LVM.

config/config.go
```diff
@@ -47,6 +47,7 @@ var (
 		"agent.flow.sflow.bind_address":        {"sflow.bind_address"},
 		"agent.flow.sflow.port_min":            {"sflow.port_min"},
 		"agent.flow.sflow.port_max":            {"sflow.port_max"},
+		"agent.topology.blockdev.lsblk_path":   {"lsblk.path"},
```
lebauce: relocationMap is for legacy configuraion (renamed or moved entries)

tests/tests.go
```diff
@@ -81,12 +81,15 @@ agent:
...
     - ovsdb
+    - blockdev
     - docker
     - lxd
     - lldp
     - runc
     - socketinfo
     - libvirt{{block "agentProbes" .}}{{"\n"}}{{range .AgentProbes}}{{println "    -" .}}{{end}}{{end}}
+    blockdev:
+      lsblk_path:
```
lebauce: This seems unnecessary (as the SKYDIVE_AGENT_TOPOLOGY_BLOCKDEV_LSBLK_PATH overrides it)

tests/blockdev_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+		checks: []CheckFunction{func(c *CheckContext) error {
+			gremlin := c.gremlin.V().Has("Type", "blockdev", "Manager", "blockdev")
+
+			nodes, err := c.gh.GetNodes(gremlin)
+			if err != nil {
+				return err
+			}
+
+			if len(nodes) > 0 {
```
lebauce: len(nodes) == 0 ?

config/config.go
trgill: ok - will remove

tests/blockdev_test.go
trgill: I was missing:

setupFunction: setupBlockDev

I updated the test.

trgill:
When I click on a block device in the UI - the right side displays the top level data, but the BlockDev doesn't expand as it does for other probes.  Is it obvious to anyone what I'm doing wrong?

lebauce:
@trgill That was a bug in the UI. We just merged a fix for it. Can you please try to rebase on master ?

trgill:
@lebauce thanks!  Rebase over master fixed the problem.

When I expand the metadata, I see minor formatting issues for Label & Mode.  Is this due to the probe adding metadata that includes HTML characters?

image

trgill:
@safchain I think this is close to displaying basic metrics via iostat.  The table displays properly - but I'm missing some part that populates the data for the table.  Do you see what I'm doing wrong?

trgill:
@safchain I think I've at least partially figured out the problem.  I had used the BlockdevMetric JSON annotations to parse the output from iostat.  But, that looks like it causes problems with the UI.  I'll fix and re-push.

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
 
 	return m1, m2
 }
+
+// BlockdevMetric the blockdevice IO stats
```
safchain: Do you think that this struct could be reused by another probe ? if so we can let it here, if not I would move it in the blockdev folder

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
...
+	AverageWriteSize   float64 `json:"wareq-sz"`
+	ServiceTime        float64 `json:"svctm"`
+	Utilization        float64 `json:"util"`
+	Start              int64   `json:"Start"`
+	Last               int64   `json:"Last"`
+}
+
+// GetBlockdevFields returns the JSON annotation for each field in
+// BlockdevMetric
+func GetBlockdevFields() []string {
```
safchain: this should be generated by the gendecoder

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
...
+		"wareq-sz",
+		"svctm",
+		"util",
+		"Start",
+		"Last",
+	}
+}
+
+// IsZero returns true if all the values are equal to zero
+func (im *BlockdevMetric) IsZero() bool {
```
safchain: If don't plan to use gremlin to do operation on metrics this function is useless

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
...
+		im.WriteServiceTime +
+		im.AverageQueueSize +
+		im.AverageReadSize +
+		im.AverageWriteSize +
+		im.ServiceTime +
+		im.Utilization) == 0
+}
+
+// Split splits a metric into two parts
+func (im *BlockdevMetric) Split(cut int64) (common.Metric, common.Metric) {
```
safchain: same here

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
...
+		AverageQueueSize:   im.AverageQueueSize * ratio,
+		AverageReadSize:    im.AverageReadSize * ratio,
+		AverageWriteSize:   im.AverageWriteSize * ratio,
+		ServiceTime:        im.ServiceTime * ratio,
+		Utilization:        im.Utilization * ratio,
+	}
+}
+
+// GetStart returns start time
+func (im *BlockdevMetric) GetStart() int64 {
```
safchain: same

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
...
+	return im.Start
+}
+
+// SetStart set start time
+func (im *BlockdevMetric) SetStart(start int64) {
```
safchain: same

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
...
+	im.Start = start
+}
+
+// GetLast returns last time
+func (im *BlockdevMetric) GetLast() int64 {
```
safchain: same

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
...
+	return im.Last
+}
+
+// SetLast set last tome
+func (im *BlockdevMetric) SetLast(last int64) {
```
safchain: same

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
...
+	var metric BlockdevMetric
+	if err := json.Unmarshal(raw, &metric); err != nil {
+		return nil, err
+	}
+
+	return &metric, nil
+}
+
+// Add sum two metrics and return a new Metrics object
+func (im *BlockdevMetric) Add(m common.Metric) common.Metric {
```
safchain: same

topology/metrics.go
```diff
@@ -232,3 +232,190 @@ func (im *InterfaceMetric) Split(cut int64) (common.Metric, common.Metric) {
...
+		AverageWriteSize:   im.AverageWriteSize + om.AverageWriteSize,
+		ServiceTime:        im.ServiceTime + om.ServiceTime,
+		Utilization:        im.Utilization + om.Utilization,
+		Start:              im.Start,
+		Last:               im.Last,
+	}
+}
+
+// Sub subtracts two metrics and return a new metrics object
+func (im *BlockdevMetric) Sub(m common.Metric) common.Metric {
```
safchain: same

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,620 @@
...
+
+func (p *ProbeHandler) connect() error {
+	var (
+		cmdOut []byte
+		err    error
+		intf   interface{}
+		result Devices
+	)
+
+	lsblkPath := p.Ctx.Config.GetString("agent.topology.blockdev.lsblk_path")
```
safchain: I think you should check that this is not empty, I think it is possible to set an empty value in the config

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,620 @@
...
+	return blockdev.Name
+}
+
+func newMetricsFromBlockdev(blockdevPath string) *topology.BlockdevMetric {
+	var (
+		cmdOut []byte
+		err    error
+		stats  sysstat
+	)
+	if cmdOut, err = exec.Command("/usr/bin/iostat", "-dx", "-o", "JSON", blockdevPath).Output(); err != nil {
```
safchain: does it make sense to make it part of the config ?

scripts/gendecoder/main.go
```diff
@@ -95,6 +95,8 @@ func handleBasic(getter *getter, name, kind string) {
 	switch kind {
 	case "int", "int8", "int16", "int32", "int64", "uint", "uint8", "uint16", "uint32", "uint64":
 		getter.Fields = append(getter.Fields, field{Name: name, Type: "int64"})
+	case "float32", "float64":
```
safchain: unfortunately this is not enough to have float handled by the gendecoder. The function used to retrieve the value are only generated for int64 and I think that the gremlin traversal thing will have to be updated too. Any chance to have int64 only values ? conversion into ms ?

trgill: Ok.  I changed to int64 - it looks like it is just truncating the floating point. I think for a first version that will be ok.

I removed the "LastUpdate" code.  It doesn't really make sense for rates (like KB per second).  I think we can come back to that at some point.

I'd like to be able to display some IO rates on the edges - but it looks like a fair amount of work for the legacy UI.  So, if you support it in the new UI - please keep in mind that a block device link would want to display the rates as IOPs or KB/s etc.

Does the way I implemented the "metrics" make sense?  I didn't see a better way.  The two structures are necessary for parsing the output of iostat and the other for passing to the UI. (?)

topology/metrics.go
trgill: OK.  I moved it to blockdev.go.  I ran into some problems with the build if I put it in a separate file.

topology/metrics.go
trgill: I was thinking the columns displayed in the UI should match the output of "iostat".  So, I call this function in server.go to use these field names rather than what is returned by .GetFieldKeys().

Some of the field names are not allowed in go structures.

topology/metrics.go
trgill: Thanks - I removed it.

topology/metrics.go
trgill: removed

topology/metrics.go
trgill: removed

topology/metrics.go
trgill: removed

topology/probes/blockdev/blockdev.go
trgill: Ok.

topology/probes/blockdev/blockdev.go
trgill: yes - done.

topology/metrics.go
safchain: I not sure that it is good the keep the orignal field name as this is not how the field are reported in skydive in general and can/will raise error with the datastore as not all characters are supported. If we want to keep the original information present in the UI then I would prefer to have a translation there ?

scripts/gendecoder/main.go
safchain: Thanks for changing to int64, we discussed about this limitation during our last public weekly call, we will remove this limitation in the future. I plan to work soon on the link labels on the "new" UI, I'll take care of this. Having two structures one for parsing and another one for the output seems to be a good approach (IMHO).

safchain:
Just a comment regarding the output format, otherwise looks good. Test needs to be fixed

topology/metrics.go
trgill: Ok - I changed to follow the conventions used for the other probes. :)

trgill:
@safchain I think I've addressed your concerns.  Does this look like a decent "first version"?   I'm thinking there are still a lot of improvements to be made - but getting feedback from users might be a good idea.

safchain:
@trgill yes pretty decent version. If you can fix functional and unit test + fmt, I'll merge it with pleasure

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,623 @@
...
+	return stats.Sysstat.Hosts[0].Statistics[0].Metrics[0].MakeCopy()
+}
+
+func (p *ProbeHandler) updateBlockDevMetric(now, last time.Time) {
+	for _, blockdev := range p.blockdevMap {
+		node := blockdev.Node
+		if node == nil {
+			continue
+		}
+		path, err := node.GetField("Path")
```
lebauce: This should be done under graph lock.

nit: you can probably use `GetFieldString` (and therefore "%s" in the Sprintf below).

topology/probes/blockdev/blockdev.go
```diff
@@ -0,0 +1,623 @@
...
+		}
+		if p.findBlockDev(path, id, newDevInfo[i].Children) == true {
+			return true
+		}
+	}
+	return false
+}
+
+func (p *ProbeHandler) deleteIfRemoved(currentDevInfo blockdevInfo, newDevInfo []BlockDevice) {
+	blockID, err := currentDevInfo.Node.GetFieldString("BlockID")
```
lebauce: It seemsBlockID is never set. Should it be "Index" instead ?

lebauce:
@trgill I took the liberty to fix the tests that failed by rebasing your PR. I added one commit (https://github.com/skydive-project/skydive/pull/2057/commits/d36956fac0febf928b221101c7506d199edf3b74) to fix the "BlockID" issue. Does it look correct to you ?

lebauce:

trgill:
@lebauce yes - that is correct.  Thanks for the improvement.

safchain:
Thanks a lot @trgill

PR #2065, opened 2019-10-13
exporters: add accounting

hunchback:


contrib/exporters/secadvisor/mod/account_file.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+type accountRecord struct {
+	StartupTime  int64 `json:"startup-time"`
+	StartupBytes int64 `json:"startup-bytes"`
+	UpdateTime   int64 `json:"update-time"`
+	UpdateBytes  int64 `json:"update-bytes"`
+}
+
+func (a *accountFile) read(filename string) *accountRecord {
```
eranra: make sense to return error code if we fail

contrib/exporters/secadvisor/mod/account_file.go
```diff
@@ -0,0 +1,92 @@
...
+	record := &accountRecord{
+		StartupTime: makeTimestamp(),
+	}
+
+	file, err := os.Open(filename)
```
eranra: Add err to return value

contrib/exporters/secadvisor/mod/account_file.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return record
+	}
+	defer file.Close()
+
+	content, _ := ioutil.ReadAll(file)
```
eranra: Need to handle failures in read (file might be open by some other process)

contrib/exporters/secadvisor/mod/account_file.go
```diff
@@ -0,0 +1,92 @@
...
+	json.Unmarshal(content, &record)
```
eranra: What happens if the record is not 'exactly' what we wrote (due to mistake / disk etc ..._ need to get the error code and handle. So `Unmarshal` can fail and we need to work on this

contrib/exporters/secadvisor/mod/account_file.go
eranra: Need to add retry mechanism to read (and also to write) since we know that other processes are going to `sit` on this file in parallel to us 
(I also advice on this point to add dedicated test as this is a classic 'long term' failure of sure feature)

contrib/exporters/secadvisor/mod/account_file.go
```diff
@@ -0,0 +1,92 @@
...
+	return time.Now().UnixNano() / int64(time.Millisecond)
+}
+
+type accountFile struct {
+	filename string
+}
+
+type accountRecord struct {
+	StartupTime  int64 `json:"startup-time"`
+	StartupBytes int64 `json:"startup-bytes"`
```
eranra: I do not understand this field ... why this is not always zero ... when we started the file (created the file) ... 'startup-time' was some value and the `startup-bytes` where zero ...

contrib/exporters/secadvisor/mod/account_file.go
```diff
@@ -0,0 +1,92 @@
...
+	UpdateTime   int64 `json:"update-time"`
+	UpdateBytes  int64 `json:"update-bytes"`
```
eranra: Change key to be called `accommodated-bytes` .... can add additional field with 'current-bytes' that has just the current update number of bytes

contrib/exporters/secadvisor/mod/account_file.go
```diff
@@ -0,0 +1,92 @@
...
+
+func makeTimestamp() int64 {
+	return time.Now().UnixNano() / int64(time.Millisecond)
+}
+
+type accountFile struct {
+	filename string
+}
+
+type accountRecord struct {
```
eranra: add field called 'version' having some known version number and check the number in read/write ... this will be useful as we can easily assume that this file will be changed due to future req. from billing / accounting

contrib/exporters/secadvisor/mod/account_file.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Reset counters
+func (a *accountFile) Reset() {
+	os.Remove(a.filename)
+	os.Remove(a.bakfile())
+}
+
+// Add to counters
+func (a *accountFile) Add(bytes int64) {
```
eranra: Add to all the steps bellow that write, read , rename ... etc ... checks for errors and at least in `debug` print them out

contrib/exporters/secadvisor/mod/account_file.go
```diff
@@ -0,0 +1,92 @@
...
+type accountRecord struct {
+	StartupTime  int64 `json:"startup-time"`
+	StartupBytes int64 `json:"startup-bytes"`
+	UpdateTime   int64 `json:"update-time"`
+	UpdateBytes  int64 `json:"update-bytes"`
+}
+
+func (a *accountFile) read(filename string) *accountRecord {
+	record := &accountRecord{
+		StartupTime: makeTimestamp(),
```
eranra: The field `StartupTime` need to reflect the exporter StartupTime and if we find an 'old' file we will use the record from that file ... need to change that behavior.

contrib/exporters/secadvisor/mod/account_file.go
hunchback: bytes accumulated since startup

hunchback:
@eranra - the reliance on a file for maintaining accounting/billing information is inherently poor; by using two files I have minimized chances of receiving incorrect data by a reader (and of course there is by-design only one writer). so I suggest we leave as-is and hopefully future versions will migrate away from this. anyway I am closing PR as all work has moved to the new repo: `skydive-flow-exporter`

contrib/exporters/secadvisor/mod/account_file.go
hunchback: I suggest the following:

- aggregate-start
- aggregate-end
- aggregate-bytes
- current-start
- current-end
- current-bytes

contrib/exporters/secadvisor/mod/account_file.go
hunchback: I can add a version number and for now just abort if the version of the file if not identical to version of software (and later we can consider migrating old format to new)

contrib/exporters/secadvisor/mod/account_file.go
hunchback: will add log messages

contrib/exporters/secadvisor/mod/account_file.go
hunchback: the behavior is as you described - the filed is overridden when reading for file and thus gets the previous timestamp recorded during the previous run

contrib/exporters/secadvisor/mod/account_file.go
hunchback: I don't see a common case in which read may fail (as we write to a temporary location and then rename). Also I don't see a case in which readers can block the single writer. Are you able (via bash commands) to construct a case which requires addressing?

contrib/exporters/secadvisor/mod/account_file.go
hunchback: will write a message to log

contrib/exporters/secadvisor/mod/account_file.go
hunchback: will log an error

contrib/exporters/secadvisor/mod/account_file.go
hunchback: will log the error

contrib/exporters/secadvisor/mod/account_file.go
hunchback: will log the error

PR #2068, opened 2019-10-15
ci/create-docker-image: Fix GOPATH inside the build container

dubek:
When running the skydive build (compile) inside the build container:
Instead of using `$GOPATH` from the host (which is something like
`/home/user/go`) as-is inside the builder container, use the proper
`$GOPATH` directory which is `/root/go`.


lebauce:
run skydive-create-docker-image

lebauce:
run skydive-create-docker-image

PR #2092, opened 2019-10-31
Add display all variables if true variable 

patsevanton:
display_all_variables by default
Add examples to inventory

### DIFF
--- a/contrib/ansible/roles/skydive_common/tasks/main.yml
+++ b/contrib/ansible/roles/skydive_common/tasks/main.yml
@@ -1,6 +1,7 @@
 - name: Display all variables/facts known for a host
   debug:
     var: hostvars[inventory_hostname]
+  when: display_all_variables|default(false)|bool == true
 
 - name: Set analyzers list
   set_fact:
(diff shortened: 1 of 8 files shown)


contrib/ansible/roles/skydive_common/tasks/main.yml
```diff
@@ -1,6 +1,7 @@
 - name: Display all variables/facts known for a host
   debug:
     var: hostvars[inventory_hostname]
+  when: display_all_variables|default(false)|bool == true
```
safchain: I would prefer to use `verbosity`. it seems to handle what you try to introduce nicely. What do you think ?
https://docs.ansible.com/ansible/latest/modules/debug_module.html

patsevanton:
```- name: Display all variables/facts known for a host
  debug:
    var: hostvars[inventory_hostname]
    verbosity: 4```

Is it ok?

safchain: yes

patsevanton:
For Clean history
https://github.com/skydive-project/skydive/pull/2106

PR #2098, opened 2019-11-04
Fixed an issue with lldp

qinze118:
hi,this is my pr!

### DIFF
--- a/topology/probes/lldp/lldp.go
+++ b/topology/probes/lldp/lldp.go
@@ -98,23 +98,9 @@ func addMulticastAddr(intf string, addr string) error {
 	copy(name[:], []byte(intf))
 
 	mac, _ := net.ParseMAC(addr)
-
-	var sockaddr syscall.RawSockaddr
-	// RawSockaddr.Data is []int8 on amd64 and []uint8 on ppc64le
-	switch data := interface{}(sockaddr.Data).(type) {
-	case [14]int8:
-		for i, n := range mac {
-			data[i] = int8(n)
-		}
-	case [14]uint8:
-		for i, n := range mac {
-			data[i] = uint8(n)
-		}
-	}
-
 	ifr := &ifreq{
 		ifrName:   name,
-		ifrHwaddr: sockaddr,
+		ifrHwaddr: rawSocketaddrFromMAC(mac),
 	}
 
 	_, _, ep := unix.Syscall(unix.SYS_IOCTL, uintptr(fd),
--- a/topology/probes/lldp/rawsockaddr_signed_int.go
+++ b/topology/probes/lldp/rawsockaddr_signed_int.go
@@ -1,0 +1,32 @@
+// +build linux,!ppc64le
+
+/*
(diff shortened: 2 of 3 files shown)


topology/probes/lldp/lldp.go
```diff
@@ -101,15 +101,25 @@ func addMulticastAddr(intf string, addr string) error {
...
-	switch data := interface{}(sockaddr.Data).(type) {
-	case [14]int8:
-		for i, n := range mac {
-			data[i] = int8(n)
-		}
-	case [14]uint8:
-		for i, n := range mac {
-			data[i] = uint8(n)
-		}
+//	switch data := interface{}(sockaddr.Data).(type) {
```
safchain: can you remove commented lines, please

topology/probes/lldp/lldp.go
```diff
@@ -101,15 +101,25 @@ func addMulticastAddr(intf string, addr string) error {
...
+//			data[i] = uint8(n)
+//		}
+//	}
+
+	sockaddr := syscall.RawSockaddr{}
+	for i,n := range mac {
+		sockaddr.Data[i] = int8(n)
+	}
+
+	ifr := &ifreq{
```
safchain: this block is duplicated, see line 125

qinze118:
oh,sorry,I have changed

lebauce:
run create-docker-image

lebauce:
@qinze118 Thanks for your PR. The switch/case on the type is required for compilation on ppc64le - the job `skydive-create-docker-image` is failing because of this.

Are you sure this PR really fixes your problem ? (and not any other side effect)

qinze118:
Running on linux(centos,ubuntu) is no problem, without considering other platforms, I added your condition again and i submmited, there is no problem with local operation,please verify.

qinze118:
Yes,Solved my problem, there is no problem in my system(centos,ubuntu), I think cross-platform should be separated, selectively compiled according to the system

lebauce:
@qinze118 Ok, I understand. In the original code was modifying a local variable... so it did nothing :-)

In fact, using separated files is required to properly fix the problem. I pushed on your branch a fix - keeping you as the author. Could you please check if it fixes the problem. If this is ok for you, I'll merge it. Thanks !

lebauce:
run create-docker-image

qinze118:
Thanks,II tried to make the next modification to be compatible with multi-platform compilation problems, by modifying the pointer to bypass the type constraint, but I did not have a platform for ppc64le to test, please test it to see if it compiles.

lebauce:
@qinze118 Using seperate files as you suggested is cleaner so I'm not sure the second PR is required. If the first one works for you, I suggest we merge it.

We do have a job to test compilation on ppc64le (using cross compilation) but we don't have any platform to do the real testing so we can't be sure the second PR works :-/

safchain:
@qinze118 @lebauce I'm going to merge this one as I prefer to have dedicated file per arch. Thanks

PR #2111, opened 2019-11-12
Removed go.sum

lebauce:


dubek:
(Not urgent)

I'm curious why `go.sum` is not committed (as opposed to the [official recommendation]( https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file)). I thought it is supposed to support reproducible builds, which sounds like a good thing.

safchain:
Hi, It was an attempts to fix an issue with package and release artifacts in general as go.sum/go.mod change every time we trigger a build. We have read the same thing but from my point of view everything around go.mod in real life differ from what the documentation suggest  :( We need probably to investigate a bit more 

We removed it thinking of re-adding it if needed.

PR #2117, opened 2019-11-22
build: Fix build error due to implicit cast to C.int

siddhesh:
When building on arm64 the build fails due to invalid cast from golang
int to C.int.  Explicitly casting to C.int like so fixes the
build.


lebauce:
run compile-tests
run functional-hw-tests

nplanel:
@siddhesh thanks for the fix

PR #2129, opened 2019-12-18
Change apiVersions and add selectors to fits new k8s release >=1.16

redno2:
On the new kubernetes release some changes is needed like the name of the apiVersion and selector which is mandatory.


lebauce:
Thank you for this @redno2 !

PR #2149, opened 2020-02-06
Change the blockdev node name to just the base name insead of full path

trgill:
When the full path to a block device is used for the node name it can be difficult to read on a graph with many devices.

Also, update block metadata from "WWN" to "ID" since not all block devices have WWNs.


safchain:
It seems that functional tests need to be adapted

PR #2153, opened 2020-02-16
ebpf: always set last timestamp whether it a new flow or not

safchain:


safchain:
run skydive-scale-tests

PR #2158, opened 2020-02-18
common: removed Retry in favor of github.com/avast/retry-go

lebauce:


lebauce:
run scale-tests

lebauce:
run scale-tests

lebauce:
run scale-tests

lebauce:
run go-fmt
run unit-tests
run compile-tests
run functional-hw-tests
run cdd-overview-tests
run k8s-tests

tests/api_test.go
```diff
@@ -19,20 +19,20 @@ package tests
...
 	shttp "github.com/skydive-project/skydive/http"
 )
 
 func getCrudClient() (c *shttp.CrudClient, err error) {
-	common.Retry(func() error {
+	retry.Do(func() error {
 		c, err = client.NewCrudClientFromConfig(&shttp.AuthenticationOpts{})
 		return err
-	}, 10, time.Second)
+	})
```
nplanel: ... , Delay(10*time.Second))

default delay between retries in retry.Do() is 100ms
So there are many place to fix ...


Why changing everything for no impact, except one more dependency ?

lebauce: 10 retries with a delay of 1 second was chosen without any real reason and this is the case for almost every common.Retry (even more for the tests where we basically copied/pasted the same 10 * 1 second line)

We are trying to remove as much as possible of the ugly code we have in common/ and the retry-go library does exactly the same (and more) in a more elegant way so I don't see any reason to keep.

I took the opportunity of this PR to look at every common.Retry and see if the number of retry, the delay and the delay mechanism (linear or exponential).

lebauce:
run unit-tests

lebauce:
run scale-tests

tests/api_test.go
lebauce: For the Retry used in tests, if they pass, that's probably enough.

For the others, we should be careful. I tried to be but it's definitely I may have missed 1 or 2

PR #2168, opened 2020-02-22
add cluster separation

roytman:
When skydive probes several K8s clusters, we have to separate their topology
skydive-clusters


roytman:
run skydive-scale-tests

PR #2187, opened 2020-03-22
probe: move hardware related root node data to a probe

safchain:


safchain:

lebauce:

topology/probes/hardware/hardware.go
```diff
@@ -0,0 +1,128 @@
...
+// Start the probe
+func (p *Probe) Start() error {
+	p.Ctx.Graph.Lock()
+	defer p.Ctx.Graph.Unlock()
+
+	tr := p.Ctx.Graph.StartMetadataTransaction(p.Ctx.RootNode)
+	defer tr.Commit()
+
+	if instanceID, err := getInstanceID(); err != nil {
+		p.Ctx.Logger.Error(err)
```
lebauce: I'm not sure we should log this as an error as the presence of the /var/lib/cloud/data/instance-id file is not the most common one. So you should probably use `Info` or even nothing if the file is missing

safchain: I will use debug, so that we can still have the information if needed

safchain: I finally added something to ignore "no such file" error

PR #2208, opened 2020-04-28
Update nsenter.c

Kuldip-Nanda:
POWER and other architectures define char to be unsigned (as the C standard leaves this undefined).
The function getopt_long returns a integer, as opposed to char.  In "c" the char type is implementation depenedent. In X86/AMD  architecture char is treated as signed character and in Power PPC64le it is treated as unsigned char. 

It is better to change the variable type from char to int at cmd/skydive/nsenter.c:117. And  then to cast this back to char at cmd/skydive/nsenter.c:120 switch statement. 

Otherwise the while loop on P9/Power machine never terminates.

### DIFF
--- a/cmd/skydive/nsenter.c
+++ b/cmd/skydive/nsenter.c
@@ -114,10 +114,10 @@ void nsexec(void)
 	int verbose = 0;
 	int need_to_fork = 0;
 	int option_index;
-	char c;
+	int c;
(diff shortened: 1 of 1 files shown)


lebauce:
run tests

lebauce:
run unit-tests
run go-fmt
run cdd-overview-tests

lebauce:

lebauce:

PR #2211, opened 2020-05-07
Fix python client to get the correct content type

adrianlzt:
API server renders the Content-Type header based on the Accept header
(graffiti/api/server/topology.go#L120).
If the client does not specify that header, the response is not
unmarshaled and raise an exception tryinh to get values from a byte
string (instead of a python dict), in function from_object

### DIFF
--- a/contrib/python/api/skydive/rest/client.py
+++ b/contrib/python/api/skydive/rest/client.py
@@ -76,7 +76,10 @@ class RESTClient:
         for k, v in self.cookies.items():
             opener.append = (k, v)
 
-        headers = {'Content-Type': 'application/json'}
+        headers = {
+                'Content-Type': 'application/json',
+                'Accept': 'application/json'
+        }
         req = request.Request(url,
                               data=encoded_data,
                               headers=headers)


adrianlzt:
The exception:
```
➜ python -c "from skydive.rest.client import RESTClient; restclient = RESTClient(\"10.20.20.151:8082\"); nodes = restclient.lookup_nodes(\"G.V().Has('Name', 'TOR3')\");"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/adrian/.local/share/virtualenvs/python-Gtx6w0x9/lib/python3.8/site-packages/skydive/rest/client.py", line 118, in lookup_nodes
    return self.lookup(gremlin, Node)
  File "/home/adrian/.local/share/virtualenvs/python-Gtx6w0x9/lib/python3.8/site-packages/skydive/rest/client.py", line 114, in lookup
    return [klass.from_object(o) for o in objs]
  File "/home/adrian/.local/share/virtualenvs/python-Gtx6w0x9/lib/python3.8/site-packages/skydive/rest/client.py", line 114, in <listcomp>
    return [klass.from_object(o) for o in objs]
  File "/home/adrian/.local/share/virtualenvs/python-Gtx6w0x9/lib/python3.8/site-packages/skydive/graph.py", line 50, in from_object
    return self(obj["ID"], obj["Host"],
TypeError: 'int' object is not subscriptable
```

safchain:
Thanks for the fix.

adrianlzt:
Is it possible to release a new python package with this fix?

I'm hitting it again and again in each new install

safchain:
Hi, I just uploaded a new version.

adrianlzt:
@safchain the new version in pypi (0.8.0) does not contains this fix

safchain:
I re-upload a version 0.8.1. It seems to be correct now

PR #2219, opened 2020-05-31
makefile: check and alert if GOPATH is undefined

safchain:


.mk/check.mk
```diff
@@ -1,4 +1,11 @@
 
+.PHONY: gopath
+gopath:
+
+ifndef GOPATH
+	$(error GOPAHT is undefined, Skydive requires to have GOPATH defined to build)
```
lebauce, line 6: not: s/GOPAHT/GOPATH/

PR #2232, opened 2020-06-24
Return to the user the reason of the invalid schema

adrianlzt:
Currently an error in the node/edge schema gives no clue to the user to
where is the error.
Append to that message the list of errors returned by the validator.

Eg.:
```
Invalid schema: [CreatedAt: Invalid type. Expected: integer, given: null UpdatedAt: Invalid type. Expected: integer, given: null (root): Must validate all the schemas (allOf)]
```

Could be improved to separate different messages

### DIFF
--- a/graffiti/schema/validator.go
+++ b/graffiti/schema/validator.go
@@ -30,13 +30,11 @@ type ErrInvalidSchema struct {
 }
 
 func (e ErrInvalidSchema) Error() string {
-	stringErrors := []string{}
-
-	for _, e := range e.Errors {
-		stringErrors = append(stringErrors, fmt.Sprint(e))
+	msg := make([]string, len(e.Errors))
+	for i, err := range e.Errors {
+		msg[len(e.Errors)-i-1] = err.String()
 	}
-
-	return fmt.Sprintf("invalid schema: %v", strings.Join(stringErrors, "; "))
+	return fmt.Sprintf("invalid schema: %v", strings.Join(msg, "\n"))
 }
 
 // Validator is the interface to implement to validate REST resources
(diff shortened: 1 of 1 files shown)


graffiti/schema/validator.go
```diff
@@ -48,7 +49,7 @@ func (v *JSONValidator) Validate(kind string, obj interface{}) error {
 	if err != nil {
 		return err
 	} else if !result.Valid() {
-		return ErrInvalidSchema
+		return fmt.Errorf("%v: %v", ErrInvalidSchema, result.Errors())
```
safchain, line 52: I would rather make change ErrInvalidSchema to become a custom error accepting a `Errors` field so that we keep the error type

```
type ErrInvalidSchema struct {
  Errors []goschema.ResultError
}

func (e ErrInvalidSchema) String() {
    return fmt.Errorf("invalid schema: %v", e.Errors)
}
```

using it like :

```
return &ErrInvalidSchema{Errors: result.Errors()}
```

adrianlzt:
How the errors are displayed still not looking good:
```
➜ curl 127.0.0.1:8082/api/node -H "Content-Type: application/json" -d '{}'
invalid schema: ID: String length must be greater than or equal to 1; CreatedAt: Invalid type. Expected: integer, given: null; UpdatedAt: Invalid type. Expected: integer, given: null; (root): Must validate all the schemas (allOf)
```

Suggestions?

safchain:
What do you think of 

```
func (e ErrInvalidSchema) Error() string {
	msg := make([]string, len(e.Errors))
	for i, err := range e.Errors {
		msg[len(e.Errors)-i-1] = fmt.Sprintf("%s", err)
	}
	return strings.Join(msg, "\n")
}
````

PR #2234, opened 2020-06-26
Add PATCH method to modify resources -> moved to #2263

adrianlzt:
Allow a new call to the REST API, using method PATCH, to modify
resources.
The body of the request should use JSON Patch
(https://tools.ietf.org/html/rfc6902) to specify the modifications.

Only metatada could be modified.
For node resources, some fields inside metadata could not be changed.

TODO:
- [x] test noderule/edgerule modifications
- [x] test with ES backend
- [X] tests
- [ ] add to swagger

### DIFF
--- a/graffiti/api/server/server.go
+++ b/graffiti/api/server/server.go
@@ -47,6 +47,7 @@ import (
 
 	auth "github.com/abbot/go-http-auth"
 	etcd "github.com/coreos/etcd/client"
+	jsonpatch "github.com/evanphx/json-patch"
 	yaml "gopkg.in/yaml.v2"
 
 	"github.com/skydive-project/skydive/graffiti/api/rest"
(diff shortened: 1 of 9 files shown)


lebauce:
run tests

adrianlzt:
Example of use:
```
➜ curl 127.0.0.1:8082/api/node/test -XPATCH -d '
[
                {
                  "op": "add",
                  "path": "/Metadata/NewProp",
                  "value": "foobar"
                },
                {
                  "op": "modify",
                  "path": "/Metadata/Prop",
                  "value": "changed prop"
                },
                {
                  "op": "remove",
                  "path": "/Metadata/OldProp"
                },
]'
```

graffiti/graph/graph.go
```diff
@@ -724,6 +725,8 @@ func (g *Graph) SetMetadata(i interface{}, m Metadata) error {
 	case *Edge:
 		e = &i.graphElement
 		kind = EdgeUpdated
+	default:
```
lebauce, line 728: nit: probably not required

graffiti/graph/memory.go
```diff
@@ -41,14 +41,19 @@ type MemoryBackend struct {
 func (m *MemoryBackend) MetadataUpdated(i interface{}) error {
 	switch i := i.(type) {
 	case *Node:
-		if _, ok := m.nodes[i.ID]; !ok {
-
+		node, ok := m.nodes[i.ID]
+		if !ok {
 			return ErrNodeNotFound
 		}
+		node.graphElement.Metadata = i.graphElement.Metadata
```
lebauce, line 48: should be removed

graffiti/api/server/server.go
```diff
@@ -206,6 +207,94 @@ func (a *Server) RegisterAPIHandler(handler rest.Handler, authBackend shttp.Auth
...
+
+				// Patch are applied to the json body in []byte format
+				resourcePatchedJSON, err := patch.Apply(resourceJSON)
+				if err != nil {
+					http.Error(w, err.Error(), http.StatusBadRequest)
+					return
+				}
+
+				resourcePatched := handler.New()
+				err = json.Unmarshal(resourcePatchedJSON, resourcePatched)
```
lebauce, line 261: nit: if err := json.Unmarshal(resourcePatchedJSON, resourcePatched); err != nil {

adrianlzt:
Moved to PR #2263 to allow maintainers modifications

PR #2238, opened 2020-06-30
Use name and type parameters for node/edge command line

lebauce:


lebauce:

PR #2242, opened 2020-07-02
Fix tests for Go 1.14

lebauce:


adrianlzt:
Running ``make functional TEST_PATTERN=Alert`` gives fatal errors:
```
fatal error: checkptr: unsafe pointer conversion
```

Looks like it is related with a new feature on go1.14 and bbolt: https://github.com/etcd-io/bbolt/issues/187

Removing the ``-race`` flags, it get stuck trying to connect to elastic. Running one by hand tests finished correctly.

Issue #2246, opened 2020-07-06
Elasticsearch backend: missing disable_sniffing in default config


adrianlzt:
In this [commit](https://github.com/skydive-project/skydive/commit/36982a150371cd553357fd6c36ac682c76b52f36) the option ``disable_sniffing`` was added, but not included in the default [config file](https://github.com/skydive-project/skydive/blob/9fdfbd87045ae9f4fd85330e3574c7e4e61c0279/etc/skydive.yml.default#L384)

lebauce:
Fixed

PR #2248, opened 2020-07-07
Elasticsearch: define several hosts and allow invalid certs

adrianlzt:
Allow skydive to connect to Elasticsearch servers with TLS but invalid
certs.

For Elasticsearch clusters, be able to only define a single endpoint
creates a SPoF at start time. Allow to pass an array of endpoints.
This breaks the current config, which expect an string in the key
"host".
Now is an slice in the key "hosts".

Expose configuration for basic auth (previously supported with
http://user:password@...)

Add option 'disable_sniffing' to etc/skydive.yml.default

Fix a typo in the error string when Elasticsearch version could not be
obtained.

Some options that could be set previously in the Elasticsearch URL are
now not available: Errorlog, Tracelog, Infolog and Healthcheck
These options could be added easily if needed.
The elastic lib function handling the creation of a new elasticsearch
client from the URL is:
https://github.com/olivere/elastic/blob/v7.0.17/client.go#L431:6

TODO:
- [x] Create a [map](https://github.com/skydive-project/skydive/blob/master/config/config.go#L51) from ``host`` to ``hosts``
- [x] Check that a simple value in ``host`` is mapped as a single element array
- [ ] Tests to check backward compatibility (.host vs .hosts, URI vs host:port)

### DIFF
--- a/graffiti/storage/elasticsearch/client.go
+++ b/graffiti/storage/elasticsearch/client.go
@@ -47,7 +47,7 @@ const (
 
 // Config describes configuration for elasticsearch
 type Config struct {
-	ElasticHost        []string
+	ElasticHosts       []string
 	InsecureSkipVerify bool
 	Auth               map[string]string
 	BulkMaxDelay       int
@@ -533,7 +533,7 @@ func NewClient(indices []Index, cfg Config, electionService etcd.MasterElectionS
 	}
 
 	client := &Client{
-		hosts:          cfg.ElasticHost,
+		hosts:          cfg.ElasticHosts,
 		cfg:            cfg,
 		indices:        indicesMap,
 		masterElection: electionService.NewElection("/elections/es-index-creator-" + u5.String()),
--- a/graffiti/storage/elasticsearch/client.go
+++ b/graffiti/storage/elasticsearch/client.go
@@ -19,9 +19,10 @@ package elasticsearch
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
-	"net/url"
+	"net/http"
 	"sort"
 	"strings"
 	"sync"
@@ -31,7 +32,6 @@ import (
 	version "github.com/hashicorp/go-version"
 	uuid "github.com/nu7hatch/gouuid"
 	elastic "github.com/olivere/elastic/v7"
-	esconfig "github.com/olivere/elastic/v7/config"
 
 	etcd "github.com/skydive-project/skydive/graffiti/etcd/client"
 	"github.com/skydive-project/skydive/graffiti/filters"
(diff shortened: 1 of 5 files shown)


adrianlzt:
The breaking change is changing ``myelasticsearch.host: string`` to ``myelasticsearch.hosts: []strings``

adrianlzt:
```
config.GetStringSlice(path + ".hosts")
```
return the default value if defined as a string (instead of slice)

adrianlzt:
Fixes #2246

adrianlzt:
Still one breaking change: .host must be defined as an URI, instead of host:port

adrianlzt:
Remove relocation stuff.

Change ``.host`` to change to an URL

adrianlzt:
@lebauce could you trigger tests again?

graffiti/storage/elasticsearch/client.go
```diff
@@ -217,14 +256,22 @@ func (c *Client) start() error {
...
 	}
 
 	c.bulkProcessor = bulkProcessor
 
-	vt, err := esClient.ElasticsearchVersion(c.url.String())
+	// Get the version from the first working node
+	// Return error if all nodes are failing
+	var vt string
+	for _, host := range c.hosts {
+		vt, err = esClient.ElasticsearchVersion(host)
```
safchain, line 268: I would rather check the minimal version for all the instances

graffiti/storage/elasticsearch/client.go
```diff
@@ -90,6 +94,30 @@ type Client struct {
...
+func (l TraceLogger) Printf(format string, v ...interface{}) {
+	logging.GetLogger().Debugf(format, v)
+}
+
+// InfoLogger implements the oliviere/elastic Logger interface to be used with info messages
+type InfoLogger struct{}
+
+// Printf sends elastic info messages to skydive logger Info
+func (l InfoLogger) Printf(format string, v ...interface{}) {
+	logging.GetLogger().Infof(format, v)
```
safchain, line 110: I can see in the log some incorrect line coming for this line, ex:

```
2020-11-10T19:36:08.583+0200    ESC[34mINFOESC[0m       elasticsearch/client.go:110 InfoLogger.Printf   skydiveci: [POST http://172.17.0.2:9200/skydive_topology_live%2Cskydive_to
pology_archive_v12%2A/_search %!s(int=200) %!s(float64=0.112)] %!s(MISSING) [status:%!d(MISSING), request:%!f(MISSING)s]
```

graffiti/storage/elasticsearch/client.go
```diff
@@ -90,6 +94,30 @@ type Client struct {
 	masterElection etcd.MasterElection
 }
 
+// TraceLogger implements the oliviere/elastic Logger interface to be used with trace messages
+type TraceLogger struct{}
+
+// Printf sends elastic trace messages to skydive logger Debug
+func (l TraceLogger) Printf(format string, v ...interface{}) {
+	logging.GetLogger().Debugf(format, v)
```
safchain, line 102: This generates huge lot entry, especially it dumps all the thing we insert. It seems to slow down the CI job to have it killed after a timeout. I would rather use Tracef here

safchain (changes requested):

graffiti/storage/elasticsearch/client.go
adrianlzt: But ``Tracef`` does not exist, right?
https://github.com/skydive-project/skydive/blob/f4de9c5201a2cc4b5b98a448aedb07e370adaf91/graffiti/logging/logging.go#L31

safchain: You're right, so I would make it optional, what about adding a config parameter to enable the tracing ?

adrianlzt:
Moved to PR #2302

PR #2257, opened 2020-07-13
Do not convert hosts to IPs

adrianlzt:
While using TLS, certs use to be valid for domain names, not for IPs.
Without this patch, communication between skydive itself is done using
the IP, failing if the cert does not contain the IP.


safchain:
The functional tests seem to be OK and It look good to me. Just need to adapt/fix the unit tests

PR #2269, opened 2020-09-07
Descendants step parameter to include multiple edge types

adrianlzt:
A new parameter has been added to the Descendants step so nodes linked
with edge types "ownership_shared" are also "descendants" nodes.

This is added so we can have several parents for a single node.

### DIFF
--- a/gremlin/traversal/descendants.go
+++ b/gremlin/traversal/descendants.go
@@ -18,8 +18,9 @@
 package traversal
 
 import (
-	"fmt"
+	"github.com/pkg/errors"
 
+	"github.com/skydive-project/skydive/graffiti/filters"
 	"github.com/skydive-project/skydive/graffiti/graph"
 	"github.com/skydive-project/skydive/graffiti/graph/traversal"
 	"github.com/skydive-project/skydive/topology"
@@ -32,8 +33,9 @@ type DescendantsTraversalExtension struct {
 
 // DescendantsGremlinTraversalStep rawpackets step
 type DescendantsGremlinTraversalStep struct {
-	context  traversal.GremlinTraversalContext
-	maxDepth int64
+	context    traversal.GremlinTraversalContext
+	maxDepth   int64
+	edgeFilter graph.ElementMatcher
 }
 
 // NewDescendantsTraversalExtension returns a new graph traversal extension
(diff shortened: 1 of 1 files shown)


adrianlzt:
I have _created_ a different RelationType, _ownership_shared_, an added a new parameter for de _Descendants_ step so it uses edges _ownership_ and _ownership_shared_ to find _descendants_ nodes.

This is a workaround to a _bigger_ problem, a node could have only one parent.

Probably this PR should not be merged, but it is to open a discussion about if the limitation of one node - one parent could be removed.

lebauce:
What about passing the relation type as parameter ?

G.V("123").Descendants() // Defaults to "ownership"
G.V("123").Descendants("ownership_shared")

safchain:
Can you elaborate about the limitation ? The only limitation I see is the tree that we build with "ownership" relation type, but you should be able to build a graph on top of it using another type of link.

adrianlzt:
@lebauce the idea is to have both

@safchain sure, our current _problem_ is represented in this graph:
image

``We have a _Service_ owned by two different _Category_, but I cannot add two ``ownership`` edges to the same child node. So my workaround is to detect that a current ``ownership`` edge exists and create the subsequent edges as ``ownership_shared`` (just a _random_ name).

Nodes and Edges are agnostic, but there are some _magic_ values (like the ``RelationType=ownership``) that have particular behaviours, like only one owership edge per child node and ``Descendants`` step only taking into account that kind of edge type.

The Gremlin query (with this PR merged) to get the nodes below ``IT / Tooling`` and ``Operaciones`` categories:
```
G.V().Has('Type', 'Category').HasEither('Description', 'IT / Tooling','Description', 'Operaciones',).Descendants(30, true).Dedup().SubGraph()
```

If I try to rewrite that query to use both kinds of edge types (``ownership`` and ``ownership_shared``) it becomes quite complicated, because in each level I have to store those nodes and then join all using ``Select``, and this is only equivalent to ``Descendants(2)``:
```
G.V().Has('Type', 'Category').HasEither('Description', 'IT / Tooling','Description', 'Operaciones',).OutE().HasEither('RelationType', 'ownership', 'RelationType', 'ownership_shared').OutV().As('0').
G.V().Has('Type', 'Category').HasEither('Description', 'IT / Tooling','Description', 'Operaciones',).OutE().HasEither('RelationType', 'ownership', 'RelationType', 'ownership_shared').OutV().OutE().HasEither('RelationType', 'ownership', 'RelationType', 'ownership_shared').OutV().As('1').
Select('0', '1').Dedup().SubGraph()
```

safchain:
What about specifying a list of `RelationType` that the `Descendants` step will follow ? would it fit your need ?

adrianlzt:
That will work.

Something like:
```
Descendants(2, 'ownership', 'ownership_shared')
```

Or like:
```
Descendants(2, Within('ownership', 'ownership_shared'))
```
?

safchain:
Sorry I didn't see your last message. I would use `Within`.

lebauce:
@adrianlzt I don't know if you still use this feature but I added a commit on top of yours to accept the same parameters than for the `Has` step in addition to the maximal depth. Could you please have a look ? Thanks !

adrianlzt:
I have tested it. Works perfect. For my use case now I can use
```
Descendants('RelationType', Within('ownership', 'ownership_shared'), 30)
```

lebauce:

lebauce:
@adrianlzt Thanks for testing !

lebauce:

lebauce:

lebauce:

lebauce:

lebauce:

PR #2327, opened 2021-01-19
add support for flow export

eranra:
- Add pod with (1) flow_exporter and (2) minio (3) mc (mino command line) to create a bucket and tail new log files
- Add Skydive analyzer startup capture on host device interfaces


eranra:

contrib/openshift/skydive-template.yaml
```diff
@@ -202,6 +204,74 @@ objects:
         - hostPath:
             path: /var/lib/origin
           name: lib-origin
+# Skydive FlowExporter (Deployment)
```
safchain, line 207: could it be a dedicated template file adding the support of flow exporter ?

eranra: @safchain done.

PR #2330, opened 2021-01-20
Elasticsearch: store prevRevision when fetching nodes/edges

### LINKED ISSUE #2328: Unable to patch nodes recovered from Elasticsearch backend
If skydive has Elasticsearch configured as backend, with some data stored from a previous run, trying to patch some node returns an error.

# How to reproduce
1. Start skydive with Elasticsearch as topology backend
2. Create a node
```
curl 127.0.0.1:8082/api/node -H "Content-Type: application/json" -d "{\"ID\": \"server1\", \"CreatedAt\": $(date +%s%3N), \"UpdatedAt\": $(date +%s%3N), \"Metadata\": {\"Name\": \"foobar\", \"Type\": \"Server\"}}"
```
3. Restart skydive
4. Try to patch the node
```
➜ curl -H "Content-Type: application/json" -XPATCH localhost:8082/api/node/server1 -d '[{"op": "add", "path": "/Metadata/XXX", "value": "foo"}]'
updating resource with patched version: Unable to update an unkwown node: server1
```

adrianlzt:
When skydive fetchs nodes and edges at boot time from elasticsearch,
those nodes/edges were not being saved in the b.prevRevision map.

If later we tried to change the metadata of that nodes/edges (using
MetadataUpdated) it failed because it does not recognise that node/edge.

This commits modifies the GetNodes and GetEdges functions to store into
b.prevRevision nodes/edges fetched from Elasticsearch.

Fixes #2328

### DIFF
--- a/graffiti/graph/elasticsearch.go
+++ b/graffiti/graph/elasticsearch.go
@@ -435,6 +435,16 @@ func (b *ElasticSearchBackend) GetEdges(t Context, m ElementMatcher, e ElementMa
 		edges = dedupEdges(edges)
 	}
 
+	for _, e := range edges {
+		raw, err := edgeToRaw(e)
+		if err != nil {
+			b.logger.Errorf("Ignoring edge, failing to marshal: %v", err)
+			continue
+		}
+
+		b.prevRevision[e.ID] = raw
+	}
+
 	return edges
 }
 
@@ -472,6 +482,16 @@ func (b *ElasticSearchBackend) GetNodes(t Context, m ElementMatcher, e ElementMa
 
 	if len(nodes) > 1 && t.TimePoint {
 		nodes = dedupNodes(nodes)
+	}
+
+	for _, n := range nodes {
+		raw, err := nodeToRaw(n)
+		if err != nil {
+			b.logger.Errorf("Ignoring node, failing to marshal: %v", err)
+			continue
+		}
+
+		b.prevRevision[n.ID] = raw
 	}
 
 	return nodes


adrianlzt:
I am not sure the full implications could take this change.

Maybe there are cases when we want to ``GetNodes`` but not to _index_ them?

I guess another approach is in ``func (c *CachedBackend) OnStarted() {``, for each node/edge fetched, add it again, to populate ``b.prevRevision``. But that will create unneeded traffic to elasticsearch.

lebauce:
Thanks for spotting and fixing the problem. I don't see a use case for not indexing neither.

Doing so in GetNodes/GetEdges work but it needs to be done only once and GetNodes/GetEdges can be called many times.
I like your proposal to do it in the `OnStarted` function. One possibility would be to add a need a  `Sync(Backend, *ElementFilter)` on the `PersistentBackend` interface that would do https://github.com/skydive-project/skydive/blob/master/graffiti/graph/cachedbackend.go#L224-L231, and in the case of Elasticsearch would provision the `prevRevision` field (it would also avoid having to return a potentially big list of nodes and edges just to call `NodeAdded` and `EdgeAdded`.

Then in CachedBackedn.OnStarted, we would simply call `c.persistent.Sync(c.memory, elementFilter)`. What do you think ?

lebauce:
I just pushed a branch with the proposal : https://github.com/lebauce/skydive/commit/c4d00c95cb0300bacaba700b7b16472b2dbf7b63

adrianlzt:
Sure, much better your option

Issue #2347, opened 2021-02-18
elasticsearch: dynamic_templates not applied when rolling


adrianlzt:
Skydive [creates the mapping](https://github.com/skydive-project/skydive/blob/8f0c4349ed0dc41b716342a434974c0eecd66696/graffiti/graph/elasticsearch.go#L598) for indexes _skydive_topology_live_ and _skydive_topology_archive_ correctly in the first boot.
But when _skydive_topology_archive_ rolls, the new index (_skydive_topology_archive_v13_-000002) does not inherit that mapping.

This new index does not _flatten_ values set in ``exclude_from_mapping``, producing a mapping explosion and the error:
```
Failed to insert entry 316: &{illegal_argument_exception Limit of total fields [1000] in index [skydive_topology_archive_v13-000002] has been exceeded     false map[] [] [] []   <nil>}
```

adrianlzt:
[roll](https://github.com/skydive-project/skydive/blob/846fda84d07c50dc46d20688b99ff6e1b5f69449/graffiti/storage/elasticsearch/rollindex.go#L81) -> [RolloverIndex](https://github.com/olivere/elastic/blob/v7.0.21/client.go#L1603) -> [NewIndicesRolloverService](https://github.com/olivere/elastic/blob/v7.0.21/indices_rollover.go#L46).[Alias](https://github.com/olivere/elastic/blob/v7.0.21/indices_rollover.go#L96) -> [Do()](https://github.com/olivere/elastic/blob/v7.0.21/indices_rollover.go#L273)

Looks like a mappnig could be passes to the rollover API call: https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-rollover-index.html

Implemented [here](https://github.com/olivere/elastic/blob/v7.0.21/indices_rollover.go#L170)

safchain:
@adrianlzt thanks for reporting this. Do you plan to submit a fix or do you want us to provide it ?

adrianlzt:
Working on it :)

safchain:
Thanks for the fix :)

PR #2348, opened 2021-02-19
elasticsearch: rolles indices should keep the same mapping

adrianlzt:
When the archive index is rolled, it was not keeping the mappings from
the original archive index.
This produces an incorrect index of values and could reach the max
fields limit if we are using some of the fields that should be
"flattened"


adrianlzt:
Fixes #2347

PR #2358, opened 2021-03-08
Add SSL support to OVN

rcarrillocruz:
Up till now, OVN would only work if UNIX sockets were used or TCP without authentication.
This change adds parameters and plumbing for client cert, key
and CA cert file.
That is then passed on to the goovn NewClient.

### DIFF
--- a/analyzer/probes.go
+++ b/analyzer/probes.go
@@ -114,7 +114,10 @@ func NewTopologyProbeBundleFromConfig(g *graph.Graph) (*probe.Bundle, error) {
 		switch t {
 		case "ovn":
 			addr := config.GetString("analyzer.topology.ovn.address")
-			handler, err = ovn.NewProbe(g, addr)
+			certFile := config.GetString("analyzer.topology.ovn.cert")
+			keyFile := config.GetString("analyzer.topology.ovn.key")
+			cacertFile := config.GetString("analyzer.topology.ovn.cacert")
+			handler, err = ovn.NewProbe(g, addr, certFile, keyFile, cacertFile)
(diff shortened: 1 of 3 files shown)


analyzer/probes.go
```diff
@@ -114,7 +114,10 @@ func NewTopologyProbeBundleFromConfig(g *graph.Graph) (*probe.Bundle, error) {
 		switch t {
 		case "ovn":
 			addr := config.GetString("analyzer.topology.ovn.address")
-			handler, err = ovn.NewProbe(g, addr)
+			certFile := config.GetString("analyzer.topology.ovn.cert")
+			keyFile := config.GetString("analyzer.topology.ovn.key")
+			cacertFile := config.GetString("analyzer.topology.ovn.cacert")
+			handler, err = ovn.NewProbe(g, addr, certFile, keyFile, cacertFile)
```
eranra, line 120: Can you add infomration in https://github.com/skydive-project/skydive/blob/master/etc/skydive.yml.default about those new options

rcarrillocruz: done

analyzer/server_easyjson.go
```diff
@@ -1,520 +1,16 @@
-// Code generated by easyjson for marshaling/unmarshaling. DO NOT EDIT.
+// TEMPORARY AUTOGENERATED FILE: easyjson stub code to make the package
```
lebauce, line 1: This file was not properly generated by easyjson. You probably only need to restore it.

lebauce (approved):

PR #2359, opened 2021-03-08
Add Dockerfile and make target

rcarrillocruz:
There is no clear Dockerfile or easy way to build a local docker image.
This patch allows to run 'make docker' and it will use a multi-stage
Dockerfile inspired by the CI Dockerfile for the build.
It allows to use DOCKER_IMAGE and DOCKER_TAG, which are defined on
.mk/dist for docker-image target, which is not great for local
development since the build runs on the host, not in container.


rcarrillocruz:
Yeah, I wanted to leave that for discussion, to avoid breakages.

PR #2371, opened 2021-04-08
how  switches can be discovered in the network topology?how to configure ？My question just likes yours.

15367060916:
My question is similar to yours. There are two hosts and one switch. The host numbers are set to 192.168.1.100 and 192.168.1.50 respectively. I am not sure whether my switch has activated the LLDP probe, so I manually grab the port information in the form of fiber configuration. Why are there no two host nodes and one switch in my topology diagram.@lebauce @RunningXJ

### DIFF
--- a/agent/probes.go
+++ b/agent/probes.go
@@ -40,7 +40,9 @@ import (
 )
 
 // NewTopologyProbeBundleFromConfig creates a new topology probe.Bundle based on the configuration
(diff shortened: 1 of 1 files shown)


nplanel:
I think you would need to configure your agent, and enable the lldp probe, like in https://github.com/skydive-project/skydive/blob/88812a79f6d89f599c877e88cb16b09d074c7adb/etc/skydive.yml.default#L217
Of course your switch need to configured to send lldp frames to the interface where the agent is running.

Issue #2375, opened 2021-06-08
Patched nodes lose typed metadata


adrianlzt:
The PATCH method of the API convert the Metadata to JSON and back to ``Node``.

In the [Unmarshal process](https://github.com/skydive-project/skydive/blob/bd5b4ca87a78a1be6130c442ba7ac4cd0e87fea1/graffiti/api/server/server.go#L472), if I understand correctly, all Metadata is processed as map[string]interface ([here](https://github.com/skydive-project/skydive/blob/bd5b4ca87a78a1be6130c442ba7ac4cd0e87fea1/graffiti/api/types/types_easyjson.go#L460)).

So, if some Metadata value is, for example, a go net.IP, that info is lost.

I think that to unmarshal types.Nodes we should use the unmarshaler of graph.Node, that uses MetadataDecoders to handle this kind of typed values correctly.

As a POC, removing the ``easyjson`` generator from types.Nodes and using this instead:
```
func (n *Node) UnmarshalJSON(data []byte) error {
	gNode := graph.Node(*n)
	err := gNode.UnmarshalJSON(data)
	*n = Node(gNode)
	return err
}
```
Seems to work correctly, but maybe there is a better way.

Maybe doing composition?
```
struct types.Node { graph.Node }
```
instead of
```
type types.Node graph.Node
```
?

adrianlzt:
Same should happen with edges

lebauce:
I pushed this PR, I think it fixes the problem but I would really appreciate if you could give it a try. Thanks !

adrianlzt:
Sorry @lebauce. I didn't see this.

The change that you propose don't use MetadataDecoders. Metadata fields using that won't be processed correctly.

Example, running skydive analyzer and skydive agent.

Analyzer runs with the debuger, breakpoint in the [Node.Get](https://github.com/lebauce/skydive/blob/80f2fa55bf9e1bc27fb4d91ccbf94c037346410f/graffiti/api/server/node.go#L76) function.

Triggering the breakpoint with the original data (interface node created by the agent, probe netlink):
```
➜ curl -sH "Content-Type: application/json"  localhost:8082/api/node/bda42bfc-79d2-4f23-441e-5f4b9323d029 | jq '.Metadata.Neighbors' | head -5
[
  {
    "MAC": "ff:ff:ff:ff:ff:ff",
    "IP": "255.255.255.255",
    "State": [
```

Neighbors metadata type is:
```
[...]

PR #2385, opened 2021-08-06
Avoid losing data types while patching nodes

adrianlzt:
PATCH was not using MetadataDecoders to convert interface{} values to
custom data types.
This commits modifies types.Nodes unmarshaling, using the one from
graph.Nodes which take into account MetadataDecoders.

### DIFF
--- a/graffiti/api/types/types.go
+++ b/graffiti/api/types/types.go
@@ -88,9 +88,17 @@ func (e *Edge) Validate() error {
 }
 
 // Node object
-// easyjson:json
 // swagger:model
 type Node graph.Node
+
+// UnmarshalJSON decodes types.Node using the custom graph.Node unmarshal which
+// uses MetadataDecoders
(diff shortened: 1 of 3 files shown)


lebauce:
@adrianlzt I answered on the issue you opened https://github.com/skydive-project/skydive/issues/2375 (with this simple fix https://github.com/lebauce/skydive/commit/80f2fa55bf9e1bc27fb4d91ccbf94c037346410f)

adrianlzt:
I have removed the test on ``tests/api_test.go``. It was not testing that metadata fields defined in MetadataDecoders were processed correctly.

In fact, I don't know how to make that test, because in ``tests/api_test.go`` we only have access to the HTTP API interface, but to test that PATCH recover types correctly, we need to create a node in the graph with some fields used by MetadataEncoders, then apply the patch and check, in the graph, that types are correct.

adrianlzt:
Closed on favour of @lebauce solution on #2375

PR #2387, opened 2021-08-06
New graph methods to improve speed

adrianlzt:
``Graph.GetNodesEdges`` allows ``SubGraph`` step to be more performant.

``GetNodesFromIDs`` implemented for future steps.


lebauce:
Thank you for this ! It seems GetNodesFromIDs is not used in this PR but in an other one. Do you mind removing it from this PR and add it in the PR where you use it (the Neighbours step) ? Thank you

adrianlzt:
Closed in favour of #2391 

I have another step (it don't have a PR because I think it is not useful for the community) using ``GetNodesFromIDs``. That's why I was putting that method here.

But yes, it makes no sense to have that _PR dependency_.

PR #2389, opened 2021-08-06
Neighbors step

adrianlzt:
Like Descendants, but using edges in any direction (Descendants only
uses edges from parent to child, Neighbors uses from parent to child and
from child to parent).

The different with the Both step, is Neighbors accumulate nodes seen.
Example (pseudo-syntax):
A -> B -> C
G.V(A).Out().Out() return: C.

But:
G.V(A).Neighbors(2) return: A,B,C

The parameters allowed are the same as in Descendants.
Example:
G.V('foo').Neighbors('RelationType',Within('ownership','foobar'),2)


adrianlzt:
Should be merged after PR #2387

lebauce:
run skydive-k8s-tests
run skydive-ovn-k8s-tests

PR #2391, opened 2021-08-09
Improve subgraph performance with persistent backends

adrianlzt:
Do just one call to get all the edges for the given nodes, instead of
one call for each node.

This new call has been added as a new method to Graph and Backends.
GetNodesEdges, return the list with all edges for a list of nodes.

Batching is used to avoid hitting the max number of clauses set by ES
(is set to the default value of 512).


lebauce (approved):

lebauce:
run ovn-k8s-tests

lebauce:

lebauce:

lebauce:

lebauce:

PR #2388, opened 2021-08-06
New graph.Node method to create a copy

adrianlzt:
Node.Copy() creates a copy of the node.
It could be useful to return in a query output the modified content of
a node without actually modiying it.

PR #2392, opened 2021-08-09
Merge step

adrianlzt:
Aggregates elements from different revisions of the nodes into a new
metadata key.
Given a metadata element, that should be a map[string]interface{},
aggregate different values into another metadata key with format
map[string][]interface{}
Eg.:
  Metadata.data V1: {"a":{x}, "b":{y}}
  Metadata.data V2: {"a":{z}, "b":{y}}
  Metadata.agg:     {"a":[{x},{z}], "b":[{y}]}

It's purpose its to show data from past revisions of the same node.
Example:
G.At(1479899809,3600).V().Merge('data','agg')

It could be also called defining the time slice in the parameters
(since, from):
G.V().Merge('A','B',1500000000,1500099999)

This step return a modified copy of the last node, with all the
aggregated data, not the node stored in the graph.
This is to avoid modiying the node stored in the graph.

This PR also modifies the Reduce method of the Neighbors step.
Merge step only needs the node IDs, so Neighbors step could skip
retrieving the full content of nodes.


adrianlzt:
It needs PR #2389 (``GetNodesFromIDs`` method) and PR #2388 (``Node.Copy``)

adrianlzt:
I am not sure if this step makes sense to anyone else but me :)

The full picture is that we are adding/removing events to/from the nodes and we want to be able to see all of those events in the past 24 hours.

lebauce:
Thanks ! It does make sense to me too :-)

@safchain did something similar when working on the new historical view of the new web UI and added a Valuemap step for this so he may have an opinion on this.

Just thinking out loud, I was wondering if we could do something like :
G.Context("-24h").V().Diff() that would give the difference between every revision of the node.

lebauce:
@adrianlzt Hello Adrian. Sorry for the delay. I was thinking that a somewhat similar implementation could be to add the `Merge` step on the output of the `ValueMap` step :

Metadata.data V1: {"a":{x}, "b":{y}}
Metadata.data V2: {"a":{z}, "b":{y}}

G.At(1479899809,3600).ValueMap("a, "b").Merge() would return
=>
{"a":[x, z], "b": [y]}

Would it suit your use case ?

adrianlzt:
Hi @lebauce. More daly here!
The thing is that I want the full node with some aggregated fields.
If I have understand correctly, your solution is to only return some fields aggregated.

But maybe you have another better approach.
For me the idea is to show the nodes in Grafana and the time selector will choose the period in which you could see events (Metadata.Events) in that node.
So for example, you see the node and its relations and you want to know if one week ago it was having events (problems).

But I always need the full node to draw it (name, type, etc).

Does it makes sense?

PR #2399, opened 2021-09-01
Fix event handler filling channel and being blocked

adrianlzt:
The goroutine consuming g.eventChan channel could get blocked if a
 linker generate more than maxEvents events.
Fix this problem discarding events when the channel is full.
Return an error if this point is reached.

Increase the number of maxEvents to reduce the probability of event
discard.


lebauce:
run ovn-k8s-tests
run k8s-tests

lebauce:
run ovn-k8s-tests

lebauce:
run ovn-k8s-tests

lebauce:
run ovn-k8s-tests

PR #2448, opened 2024-06-21
Update OCP Template resource's apiVersion

masco:
Openshift moved the Template resource from api v1 to template.openshift.io/v1 on Openshift V4+. updating the OCP deploying script for the same.

more details: https://access.redhat.com/solutions/7067952


lebauce:
Thank you @masco !

PR #2452, opened 2025-07-17
WIP update golang version 1.24.5, kube, ...

nplanel:


nplanel:
compile ... but untested yet

